PR #2, opened 2026-05-29
[codex] Add terminal UI

tonyfettes:
## Summary

Adds an OpenSeek terminal UI under `cmd/tui`, with a reusable DECSTBM renderer package and an alternate-screen renderer package. The TUI supports prompt/command input, queued input display, agent progress/status updates, tool-result transcript items, mouse scroll support in the alternate-screen renderer, and a `--tui`/`OPENSEEK_TUI` renderer switch.

The DECSTBM path now rebuilds transcript scrollback from semantic transcript rows on resize, matching Codex-style behavior where short transcripts start at the top and push the composer down.

## Notes

This PR keeps the current agent API mostly stateless per submitted task. A follow-up should add shared message history support so the TUI can run true continuous conversations across turns.

## Validation

- `moon fmt`
- `moon info`
- `moon check`
- `moon test` (290 passed)
- tmux smoke for DECSTBM resize/input/exit behavior before the final rebase


tonyfettes:
Split into two stacked, single-commit PRs rebased on current main:
- #12 — `tui` terminal UI primitive packages (purely additive)
- #13 — `cmd/tui` app + agent on_event integration, stacked on #12

The `codex/tui` branch is kept as a backup of the full history.

PR #6, opened 2026-05-29
ci: enforce moon fmt, check, and test

tonyfettes:
## Summary

Adds a GitHub Actions CI workflow that enforces three gates on every push to `main` and every pull request:

- `moon fmt --check` — formatting
- `moon check --deny-warn` — no warnings
- `moon test` — all tests pass

The workflow reuses the MoonBit install steps from the existing `copilot-setup-steps.yml`.

## Commits

1. **chore: normalize formatting with moon fmt** — `moon fmt --check` was failing on `main` for 4 files in the `deepseek/` package. Normalized them in a dedicated commit so the new fmt gate is green.
2. **ci: add workflow for moon fmt, check, and test** — the workflow itself.

## Verification (local, native target)

| Check | Result |
|---|---|
| `moon fmt --check` | ✅ clean |
| `moon check --deny-warn` | ✅ passes |
| `moon test` | ✅ 245/245 |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,34 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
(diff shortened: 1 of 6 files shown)


bobzhang:
> Downloading moonbitlang/async@0.19.1

we can filter such line in stable* function

PR #41, opened 2026-06-06
[codex] add background daemon runtime and watcher-backed moon_check

bobzhang:
## Summary

Adds session-scoped background monitoring to the agent runtime and uses it in two ways:

- `check_daemon` remains a generic start/status/stop monitor for non-MoonBit long-running or polling commands such as review/status feedback.
- `moon_check` is watcher-backed: the first call for a cwd/path/options tuple starts `moon check --watch --output-json --diagnostic-limit 10`, later calls with the same tuple return the latest snapshot instead of starting a duplicate watcher, and background `[moon_check update]` messages are injected before later model turns.
- If `moon --watch` crashes with the Moon panic banner, `moon_check` compacts the crash output, carries that compact summary into the replacement watcher, and auto-restarts within a small restart budget. Ordinary non-crash non-zero exits preserve their original diagnostics and do not restart.

This gives the agent the intended “call `moon_check` once, keep editing, receive compiler updates” workflow while still allowing repeated `moon_check` calls for explicit status inspection.

## Changes

- Add `AgentRuntime`, daemon records, Moon check watcher records, and a bounded runtime event queue in `agent_tool`.
- Add `DaemonUpdate` and `MoonCheckUpdate` events, coalesced by daemon id or watcher key before model turns.
- Add the `check_daemon` tool with `start`, `status`, and `stop` actions for generic background monitors.
- Convert `moon_check` from one-shot execution to a persistent watcher that reuses existing running watchers for matching arguments.
- Translate the common single package-path case to `moon check --watch --output-json --package-path ...`; reject multiple paths with a clear tool error because Moon rejects positional path lists under `--watch`.
- Add stale watcher-id protection so old watcher output cannot overwrite a replacement watcher.
- Add crash compaction and crash-only auto-restart for `moon_check` watchers.
- Update base/flash prompts, docs, generated interfaces, and watcher/reuse/restart tests.

## Validation

- Fresh Codex CLI review of final diff: no actionable correctness issues.
- `moon fmt`
- `moon info`
- `moon test agent_tool/moon_check agent_tool agent prompt --diagnostic-limit 160` (`37/37` passed)
- `moon test --diagnostic-limit 160` (`361/361` passed)
- `moon cram test tests/cram` (`2` cram tests passed)
- `git diff --check`

## Experiment

Reran three DeepSeek V4-Pro TOML-parser trials from commit `00f2fed` using the watcher-backed `moon_check`. All three agent runs finished and all three produced workspaces that passed independent validation (`moon check`, `moon test`, file JSON CLI probe, stdin JSON CLI probe, clean duplicate-key error probe). Trial 02 exercised a real `moon --watch` crash (`exit=101`) and recovered through the compact crash-summary auto-restart path.


bobzhang:
Live dsv4-pro experiment summary:

I ran three parallel TOML-parser tasks in isolated temp workspaces using this PR branch.

- Explicit daemon task: selected `check_daemon` and used `moon check --watch --output-json`, but stopping the daemon caused the agent loop to exit with `Cancelled` before final validation. This led to the follow-up commit that spawns the monitor reader with `allow_failure=true` and adds a regression test.
- Softly worded task: eventually selected `check_daemon`, but incorrectly used one-shot `moon check --output-json --warn-list +unnecessary_annotation`; that command exited immediately and behaved like a worse `moon_check`. This led to stricter tool/prompt guidance and a quick-exit note in `check_daemon` output.
- Control task: never selected `check_daemon`; it used repeated `moon_check` calls.

None of the three generated TOML parser workspaces passed final `moon test`; the point of the run was tool behavior, and the actionable findings are now reflected in the PR.

bobzhang:
Pushed `f5325ab` with the watcher-backed `moon_check` semantics:

- first `moon_check` call starts `moon check --watch --output-json`
- repeated calls with the same cwd/path/options reuse the existing watcher and return the latest snapshot
- background `[moon_check update]` messages are now injected through the runtime event queue
- single package path uses `--package-path`; multiple watch paths return a clear tool error because Moon rejects positional path lists under `--watch`

Validation: `moon info && moon fmt`, `moon check --diagnostic-limit 160`, `moon test --diagnostic-limit 160` (`356/356`), and `git diff --check`.

bobzhang:
Watcher experiment rerun with DeepSeek V4-Pro on the TOML parser CLI task, 3 parallel trials under `.moonagent/eval_runs/toml_v4pro_watcher_20260607_083657`:

| Trial | Exit | Finished | `moon_check` calls | Watcher updates | Successful watcher states | Independent validation |
| --- | ---: | ---: | ---: | ---: | ---: | --- |
| 01 | 0 | yes | 0 | 0 | 0 | `moon check`/`moon test` pass in nested `tomljson`; file/stdin probes pass; duplicate emits useful stderr |
| 02 | 0 | yes | 1 | 43 | 17 | `moon check`/`moon test` pass at workspace root; file/stdin probes pass; duplicate emits useful stderr |
| 03 | 1 | no | 0 | 0 | 0 | fails `moon check` with 16 errors |

Observed effect: trial 02 exercised the intended semantics: one `moon_check` call started `moon-check-1`, then the agent received 43 `[moon_check update]` messages without polling `moon_check` again. The watcher reported both error cycles and later success states. Trials 01 and 03 did not choose `moon_check`, so prompt/tool selection is still inconsistent.

Follow-up suggested by the logs: make the prompt/tool description stronger about calling `moon_check` immediately after creating the MoonBit project, and consider summarizing/delta-filtering watcher diagnostics because large error tails were repeatedly injected.

PR #49, opened 2026-06-09
[codex] add explicit session support to TUI

bobzhang:
Stack: 5/5. Depends on #54 (`codex/session-cli`).

This PR is the TUI-only integration layer for durable sessions. The lower PRs define the session model, filesystem store, agent turn API, and `openseek` CLI behavior; this PR wires those capabilities into `openseek-tui` without changing the core session APIs.

What changed:

- Adds `--session` / `OPENSEEK_SESSION` to `openseek-tui` for explicitly choosing a durable session id.
- Adds `--session-root` / `OPENSEEK_SESSION_ROOT` so the TUI and spawned engine agree on where session files live.
- Parses a missing or empty session id as `None`; there is no implicit `default` session.
- Passes session settings to the spawned `openseek` engine through environment variables only when a session id is present.
- Keeps fresh TUI launches fresh: `openseek-tui` without `--session` does not persist or resume conversation state.
- Loads an existing durable session at TUI startup and renders its prior user, assistant, tool, runtime, summary, and terminal events into the transcript.
- Adds focused tests for session config parsing, env propagation to the engine, omitted env for fresh runs, session-history rendering, and prompt draining with durable sessions.
- Updates TUI cram docs so the help output includes the new session flags.

Expected behavior:

```bash
openseek-tui --session parser-fix   # create/resume durable session "parser-fix"
openseek-tui                       # fresh TUI run, no durable session
```

This mirrors the CLI behavior from #54:

```bash
openseek --session parser-fix       # create/resume durable session "parser-fix"
openseek                           # fresh one-shot run
```

Validation:

- `moon test cmd/tui cmd/openseek agent agent_session agent_session/store --warn-list +73 --diagnostic-limit 300`
- `moon info cmd/tui cmd/openseek agent agent_session agent_session/store --target native`
- `moon fmt cmd/tui cmd/openseek agent agent_session tests/cram/tui.md tests/cram/cli.md README.mbt.md`
- `moon cram test tests/cram`

### DIFF
--- a/cmd/tui/main.mbt
+++ b/cmd/tui/main.mbt
@@ -32,6 +32,19 @@ fn cli_command() -> @argparse.Command {
         default_values=["openseek"],
         about="Agent engine binary to spawn; reads its JSONL event stream from stdout.",
       ),
+      OptionArg(
+        "session",
+        long="session",
+        env="OPENSEEK_SESSION",
+        about="Create or resume this durable session id.",
+      ),
+      OptionArg(
+        "session-root",
+        long="session-root",
+        env="OPENSEEK_SESSION_ROOT",
+        default_values=[".openseek"],
+        about="Directory containing durable OpenSeek sessions.",
+      ),
     ],
     positionals=[
       PositionArg(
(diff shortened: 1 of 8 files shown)


cmd/tui/main.mbt
```diff
@@ -32,6 +32,20 @@ fn cli_command() -> @argparse.Command {
         default_values=["openseek"],
         about="Agent engine binary to spawn; reads its JSONL event stream from stdout.",
       ),
+      OptionArg(
+        "session",
+        long="session",
+        env="OPENSEEK_SESSION",
+        default_values=["default"],
+        about="Create or resume this durable session id.",
+      ),
```
tonyfettes, line 40: Just to confirm, this means that all different invocation of `openseek` binary will be create/resume-from a `default` session? Like,

```
$ openseek "My name is tonyfettes"
...
$ openseek "What's my name"
# We expect the openseek to see previous messages and response something like "tonyfettes"
```

bobzhang: Addressed in the current version of this PR: there is no implicit `default` durable session now.

The behavior is explicit:

```bash
openseek --session id      # create/resume durable session "id"
openseek                  # fresh one-shot run, no durable session
openseek-tui --session id  # create/resume durable session "id"
openseek-tui              # fresh TUI run, no durable session
```

For the TUI specifically, the `session` option no longer has a default value; `parse_config` maps a missing/empty value to `None`, and the engine only receives `OPENSEEK_SESSION` when a session id is present.

PR #50, opened 2026-06-09
[codex] add DeepSeek streaming agent support

bobzhang:
## Summary

This refreshes the streaming PR on top of current `origin/main`, after the durable session and explicit TUI session slices were merged.

It adds streaming without changing the durable session contract:

- adds DeepSeek SSE streaming through `Client::chat_stream`
- encodes `stream=true` in chat requests when streaming is used
- accumulates streamed assistant content, reasoning, tool-call deltas, and final usage into the same `ChatResponse` shape as the non-streaming client
- rejects stream EOF before `[DONE]`
- switches the agent loop from `client.chat` to `client.chat_stream` while preserving `run_turn_with_append` persistence points
- emits `assistant_delta` JSONL events as content arrives
- renders live streamed text in the TUI activity line while deltas arrive
- clears the live streaming display when the final `assistant_message` is committed, so tool execution after a message falls back to the normal working state instead of showing committed text as still streaming
- caps the transient TUI streaming preview by display width; the full assistant text is still committed from the final `assistant_message`

## Scope

This PR is intentionally limited to model-response streaming. It does not change session storage, session CLI/TUI flags, or the one-turn append API beyond making the model call stream internally.

## Review Follow-Up

- Subal flagged that committed assistant text could remain visible as a live streaming status while tool calls continued. This is addressed by splitting TUI state into `streaming_response` for live deltas and `step_response` for final-message dedupe.
- Subal flagged that live streamed previews were accumulated with repeated unbounded string copies. This is addressed by capping the TUI-only preview and leaving final content to `assistant_message`.

## Validation

- `moon check deepseek/client --target native --warn-list +73 --diagnostic-limit 300`
- `moon check agent --target native --warn-list +73 --diagnostic-limit 300`
- `moon check cmd/tui --target native --warn-list +73 --diagnostic-limit 300`
- `moon test deepseek/client --target native --warn-list +73 --diagnostic-limit 300` - 7/7 passed
- `moon test agent --target native --warn-list +73 --diagnostic-limit 300` - 2/2 passed
- `moon test cmd/tui --target native --warn-list +73 --diagnostic-limit 300` - 26/26 passed
- `moon fmt agent cmd/tui deepseek deepseek/client tests/live`
- `moon fmt cmd/tui`
- `moon info deepseek --target native`
- `moon info deepseek/client --target native`
- `moon info agent --target native`
- `moon info cmd/tui/internal/event --target native`
- `moon info cmd/tui --target native`
- `moon ide analyze deepseek/client --target native`
- `moon ide analyze deepseek --target native`
- `moon ide analyze agent --target native`
- `moon ide analyze cmd/tui/internal/event --target native`
- `moon ide analyze cmd/tui --target native`
- `moon check --target native --warn-list +73 --diagnostic-limit 300`
- `moon test --target native --diagnostic-limit 300` - 455/455 passed
- `moon cram test tests/cram` - 6/6 passed
- `git diff --check`

### DIFF
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -150,7 +150,15 @@ async fn run_with_runtime(
         messages.push(ChatMessage(User, content=update))
       }
       @xlog.info() <? { "event": "agent_step", "step": step + 1 }
-      let response = client.chat(messages, tools=tools.function_tools())
+      let response = client.chat_stream(
+        messages,
+        tools=tools.function_tools(),
+        on_content_delta=delta => {
+          if delta != "" {
+            @xlog.info() <? { "event": "assistant_delta", "content": delta }
+          }
+        },
+      )
       if response.content != "" {
         @xlog.info() <?
           { "event": "assistant_message", "content": response.content }
--- a/cmd/tui/internal/event/decode.mbt
+++ b/cmd/tui/internal/event/decode.mbt
@@ -11,6 +11,7 @@
 pub fn decode_event(object : Json) -> AgentEvent? {
   match @jsonx.string(object, "event") {
     "agent_step" => Some(AgentStep)
+    "assistant_delta" => Some(AssistantDelta(@jsonx.string(object, "content")))
     "assistant_message" =>
       Some(AssistantMessage(@jsonx.string(object, "content")))
     "runtime_update" => Some(RuntimeUpdate(@jsonx.string(object, "content")))
--- a/cmd/tui/internal/event/decode_test.mbt
+++ b/cmd/tui/internal/event/decode_test.mbt
@@ -10,6 +10,12 @@ test "decode maps each engine event kind" {
   )
   assert_true(
     @event.decode_event(
+      Json::object({ "event": "assistant_delta", "content": "tok" }),
+    )
+    is Some(AssistantDelta("tok")),
+  )
+  assert_true(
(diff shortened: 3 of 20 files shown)


bobzhang:
Subal review completed for PR #50.

Findings addressed in `d314b99`:
- P2: streamed tool-call arguments were accumulated with repeated string concatenation. Fixed by storing streamed arguments in `StringBuilder` and materializing once when synthesizing the final response.
- P2: new stream tests required local TCP bind permissions in default test runs. Fixed by skipping those socket-backed tests only when the OS denies the `127.0.0.1:0` bind.

Validation after fixes:
- `moon check deepseek/client --warn-list +73 --diagnostic-limit 300`
- `moon test deepseek/client --warn-list +73 --diagnostic-limit 300` -> 7/7
- `moon test deepseek cmd/tui/internal/event cmd/tui --warn-list +73 --diagnostic-limit 300` -> 51/51
- `moon info && moon fmt`
- `moon test --warn-list +73 --diagnostic-limit 300` -> 394/394
- `moon cram test tests/cram` -> 5/5

Note: the Subal sandbox could not run local TCP or DNS-backed tests (`Operation not permitted` for `TcpServer`, DNS blocked for the existing real-world smoke test), so I validated those locally after applying the skip guard.

PR #68, opened 2026-06-10
Stream assistant and reasoning deltas live in the TUI

bobzhang:
## Summary

Streaming previously "worked" in the client but the TUI still showed answers landing in one batch. This PR completes the end-to-end streaming experience:

- **`deepseek/client`**: new `Client::chat_stream` parses the SSE stream, fires `on_content_delta` / `on_reasoning_delta` per fragment, and returns the fully accumulated `ChatResponse`. The request pins `Accept-Encoding: identity` so no gzip-compressing intermediary can buffer and re-batch the deltas. `encode_chat_request` now sends `response_format` only when explicitly asked for `JsonObject` (previously every response was constrained to a JSON object), and gains the `stream` flag.
- **`agent` / `cmd/openseek`**: the loop streams each turn and logs `assistant_delta` / `reasoning_delta` JSONL events as fragments arrive. The engine's JSONL goes through a per-line `fflush(stdout)` handler — stdout is a pipe under the TUI, where plain `println` is block-buffered and was the root cause of the batchy output.
- **`cmd/tui` / `tui`**: the activity line shows a live **tail** preview fitted to the terminal width (`Ui::columns()`), with whitespace collapsed to keep it one row and a leading `…` marking the elided front. The old head-truncation froze the line after 240 columns. Thinking-mode runs show the dimmed reasoning tail (`Thinking …`) until content starts (`Streaming …`), so the line moves from the first token instead of sitting on `Working (Ns)` through the reasoning phase.

## Verification (live, real API)

- Engine → pipe: 76 `reasoning_delta` lines spread over 6.6s and 51 `assistant_delta` lines over 0.7s, each line flushed on arrival — incremental, not one burst.
- TUI in a pty (100×30): 207 distinct activity-line frames — `Thinking The` → `Thinking The user wants…` → `…Let me compose that.` → `Streaming A terminal multiplex…` → `…happens to your connection.` — then the full response committed to the transcript.
- Probes: Ctrl-C mid-stream commits `interrupted` and leaves a live idle composer; a 44-column terminal clamps the tail without breaking the frame.
- `moon check` clean, 431/431 unit tests, 6/6 offline cram, 4/4 live cram (the lifecycle example now filters `*_delta` events, whose count varies per run by design).

### DIFF
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -150,7 +150,20 @@ async fn run_with_runtime(
         messages.push(ChatMessage(User, content=update))
       }
       @xlog.info() <? { "event": "agent_step", "step": step + 1 }
-      let response = client.chat(messages, tools=tools.function_tools())
+      let response = client.chat_stream(
+        messages,
+        tools=tools.function_tools(),
+        on_content_delta=delta => {
+          if delta != "" {
+            @xlog.info() <? { "event": "assistant_delta", "content": delta }
+          }
+        },
+        on_reasoning_delta=delta => {
+          if delta != "" {
+            @xlog.info() <? { "event": "reasoning_delta", "content": delta }
+          }
+        },
+      )
       if response.content != "" {
         @xlog.info() <?
           { "event": "assistant_message", "content": response.content }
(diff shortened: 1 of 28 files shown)


bobzhang:
Superseded by the reviewable stack #70 → #71 → #72 → #73 (same final tree, split per layer: response-format opt-in, SSE client, engine delta events + flushed stdout without the C stub, TUI tail preview). The stack's tip is byte-identical to the live-verified state this PR held.

PR #69, opened 2026-06-10
fix(deepseek): make JSON response mode opt-in

tonyfettes:
## Problem

In the TUI, the model's final answer rendered as raw JSON instead of prose.

Root cause: `deepseek/json_encode.mbt` hardcoded `response_format={"type":"json_object"}` on **every** chat request (`encode_chat_request`), so the DeepSeek client always ran in JSON Mode. In that mode the API constrains each assistant message's `content` to be a valid JSON object. When the agent loop finished a turn (no more tool calls), `response.content` came back as a JSON string, and the TUI rendered it verbatim.

## Fix

- Add a `ResponseFormat` enum (`Text` / `JsonObject`) to the `deepseek` package.
- Thread an optional `response_format` through `encode_chat_request` and `Client`. It is **omitted by default**, so the model replies with ordinary text; pass `JsonObject` only when structured JSON output is wanted.
- The agent constructs its `Client` without `response_format`, so it now gets natural-language replies.

## Tests

- `encode_chat_request` now asserts the field is omitted by default, and added cases for `JsonObject` / `Text`.
- Updated the redaction snapshot and the realworld smoke test (which explicitly opts into `JsonObject` to keep its exact-output assertion deterministic).
- `moon check` / `moon test` (423 passed) / `moon fmt` / `moon info` all clean.

### DIFF
--- a/deepseek/client/README.mbt.md
+++ b/deepseek/client/README.mbt.md
@@ -20,8 +20,10 @@ The package depends on `moonbitlang/async/http` and is native-only.
 ## Request Behavior
 
 `Client::chat` builds a private request envelope with the client's configured
-model, thinking mode, and reasoning effort, then sends it to `api_url` as JSON.
-It always sends `stream=false` and `response_format={"type":"json_object"}`.
+model, response format, thinking mode, and reasoning effort, then sends it to
+`api_url` as JSON. It always sends `stream=false`. `response_format` defaults to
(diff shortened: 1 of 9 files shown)


tonyfettes:
Superseded by 749936a (`feat(deepseek): make JSON response format opt-in`), which already landed the same fix on main with an omit-by-default `response_format` on `Client::chat`. Closing as redundant.

PR #77, opened 2026-06-10
refactor(tui): replace per-field setters with one set_live_view redraw

bobzhang:
First item off the `improvement.md` checklist.

## Problem

`refresh_ui` runs after every message-loop turn and called `set_status`, `set_activity`, and `set_queued_inputs` — each of which queues its own full-frame repaint on the UI command queue. Three repaints per message; during a streaming delta burst that triples render work for no visual benefit.

## Change

- `Ui::set_live_view(status~, activity~, queued_inputs~)` replaces the three setters (they had no callers besides `refresh_ui`): it takes the whole derived live view for the turn and redraws **once**. The shape also makes the bug unrepresentable — you can no longer accidentally repaint per field.
- `tui/README.md` example rewritten around the derive-the-view-each-turn model.
- Ticks the checkbox in `improvement.md`.

## Verification

- `moon check` 0 warnings, fmt clean, 430/430 tests, 6/6 offline cram.
- Live pty run (real API): streaming tail preview renders identically through the single-redraw path — 49 distinct activity frames ending `…re-rendering UI elements when their underlying data or state changes.`

### DIFF
--- a/cmd/tui/status_view.mbt
+++ b/cmd/tui/status_view.mbt
@@ -138,7 +138,9 @@ fn format_activity_text(state : State, cols~ : Int) -> @doc.Text? {
 
 ///|
 async fn refresh_ui(ui : @ui.Ui, state : State) -> Unit {
-  ui.set_status(format_status_bar_text(state))
-  ui.set_activity(format_activity_text(state, cols=ui.columns()))
-  ui.set_queued_inputs(state.queued)
+  ui.set_live_view(
+    status=format_status_bar_text(state),
+    activity=format_activity_text(state, cols=ui.columns()),
+    queued_inputs=state.queued,
+  )
 }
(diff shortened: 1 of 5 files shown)


bobzhang:

**Round 1** flagged a P2 — removing `set_status`/`set_activity`/`set_queued_inputs` outright breaks downstream consumers of the published module:

> Keeping the three setters as deprecated compatibility wrappers while letting `refresh_ui` use `set_live_view` would preserve the batching benefit without breaking downstream code.

Addressed in 2e797ff: the three setters are back as `#deprecated` wrappers pointing at `set_live_view`, to be dropped in a future version bump.

**Round 2** (after the fix):

> No actionable correctness issues were found in the diff. `moon check` passes; `moon test` only failed on the network-dependent DeepSeek smoke test in this restricted environment.

PR #79, opened 2026-06-10
feat(tui): keep the turn's reasoning as a transcript thought aside

bobzhang:
Second item off the `improvement.md` checklist (plus the live-cram hardening it required).

## Problem

Thinking-mode reasoning only ever existed as the transient `Thinking …` activity tail and was discarded once content started — the *why* behind an answer never survived the moment it streamed.

## Change

- **Engine**: emits one `reasoning_message` event per turn — the full accumulated reasoning, ordered before `assistant_message` so consumers appending events in order show the thought above the answer it led to.
- **TUI**: new `Thought` transcript variant (dim `✻` marker, same titled-block layout as tool calls/notices); the reasoning lands as a compact aside (first/last lines via the shared `compact_lines` limit) and the live preview clears.
- **Live cram**: the lifecycle example now *whitelists* `{agent_step, usage, agent_finished}` instead of blacklisting payload suffixes — new event kinds (like this one) can no longer break it. Ticks that checklist item too, since this change is exactly what would have broken it.

Scope note: resume does not replay thoughts — sessions store no reasoning for no-tool turns; recorded in `improvement.md` alongside the session items.

## Verification

- `moon check` 0 warnings, fmt clean, 431/431 tests, 6/6 offline cram.
- Live pty run (real API): transcript reads
  ```
  ✻ thought
    └ Let me calculate 17 times 23.
      17 * 23 = 17 * (20 + 3) = 17*20 + 17*3 = 340 + 51 = 391.
      Let me verify: 17 * 23 = 391.
  ⏺ 391
  ```

### DIFF
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -164,6 +164,15 @@ async fn run_with_runtime(
           }
         },
       )
+      // Reasoning first, so a consumer appending events in order shows the
+      // thinking above the answer it led to.
+      if response.reasoning_content != "" {
+        @xlog.info() <?
+          {
+            "event": "reasoning_message",
+            "content": response.reasoning_content,
+          }
+      }
       if response.content != "" {
         @xlog.info() <?
           { "event": "assistant_message", "content": response.content }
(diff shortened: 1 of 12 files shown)


bobzhang:
Codex CLI review (`codex review --base main`):

> No actionable regressions were found in the reviewed diff. The new reasoning event is decoded and rendered consistently, and targeted native tests for the affected packages pass.

PR #81, opened 2026-06-10
feat(tui): add --continue to resume the most recent session

bobzhang:
Fourth item off the `improvement.md` checklist (the `--continue` half; in-TUI session switching stays listed).

## Change

- **`SessionStore::latest()`**: the most recently active session, judged by event-log mtime (the header is written once at creation; the log is touched on every append, so its mtime tracks the last actual turn). Falls back to the header mtime for created-but-never-appended sessions, and skips directories whose files can't be inspected — one broken session shouldn't make the rest unresumable.
- **`openseek-tui --continue`**: resumes that session. Rejected when combined with `--session` (two different answers to "which conversation"); an empty store keeps the generated fresh session rather than refusing to launch.
- Help-output cram snapshots updated for the new flag.

## Verification

- `moon check` 0 warnings, fmt clean, 436/436 tests (3 new: store recency ordering incl. append-makes-older-session-latest, flag conflict, resolution + empty-store fallback), 6/6 offline cram.
- Live two-launch pty run (real API): run 1 told the TUI a name in session `tui-20260610-054113-879`; run 2 with `--continue` resumed the **same id**, replayed the history into the transcript, and answered **Hongbo**.

### DIFF
--- a/agent_session/store/pkg.generated.mbti
+++ b/agent_session/store/pkg.generated.mbti
@@ -19,6 +19,7 @@ pub async fn SessionStore::append(Self, @agent_session.Session, @agent_session.S
 pub async fn SessionStore::compact(Self, @agent_session.Session, content~ : String, from_sequence~ : Int, to_sequence~ : Int) -> @agent_session.Session
 pub async fn SessionStore::create(Self, @agent_session.Session) -> Unit
 pub async fn SessionStore::exists(Self, @agent_session.SessionId) -> Bool
+pub async fn SessionStore::latest(Self) -> @agent_session.SessionId?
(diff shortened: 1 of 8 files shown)


bobzhang:

**Round 1** flagged a P2 — `latest()` could pick a torn session (stat-able event log, missing/corrupt header) and hand `--continue` an unresumable id while an older valid session exists.

Addressed in a321a3e: candidates are now probed newest-first with a real `load`; the first loadable one wins (common case still costs a single load). New test covers the torn-newest-directory scenario.

**Round 2** (after the fix):

> The --continue plumbing and latest-session selection appear consistent with the existing session store behavior, and targeted tests for the touched packages pass.

PR #82, opened 2026-06-10
feat: expose thinking controls as engine and TUI flags

bobzhang:
Fifth item off the `improvement.md` checklist.

## Change

- `--thinking enabled|disabled` (`OPENSEEK_THINKING`) and `--reasoning-effort high|max` (`OPENSEEK_REASONING_EFFORT`) on **both** `openseek` and `openseek-tui`; the TUI forwards them through the engine environment like its other settings, and the TUI parses eagerly so a typo fails at launch rather than inside the first turn.
- Threaded through `@agent.run` / `run_turn` / `run_turn_with_append` as optional params defaulting to the previously hardcoded `Enabled`/`Max` — no behavior change for existing callers.
- `ThinkingMode::parse` / `ReasoningEffort::parse` added to the pure `deepseek` package as the `Show` inverses.
- **Live-verified API constraint**: DeepSeek 400s (`thinking options type cannot be disabled when reasoning_effort is set`) if effort rides along while thinking is off — caught in live verification of the new flag, fixed by omitting effort when thinking is disabled.
- Help-output cram snapshots regenerated from the real binaries (the long new flag re-padded the whole option column).

## Verification

- `moon check` 0 warnings, fmt clean, 439/439 tests (parse round-trips + rejection, TUI config validation, engine-env forwarding), 6/6 offline cram.
- Live: `--thinking disabled` streams 27 content deltas with **zero** reasoning events and finishes cleanly; the default run still emits reasoning deltas as before.

### DIFF
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -21,12 +21,24 @@ pub async fn run(
   task : String,
   max_steps? : Int = default_max_steps,
   system_prompt_text? : String = default_system_prompt_for_model(model),
+  thinking? : @deepseek.ThinkingMode = Enabled,
+  reasoning_effort? : @deepseek.ReasoningEffort = Max,
 ) -> Unit {
   let session = @agent_session.Session(
     SessionId("one-shot"),
     system_prompt=system_prompt_text,
   )
-  ignore(run_turn(api_key, model, session, task, max_steps~))
+  ignore(
+    run_turn(
+      api_key,
+      model,
+      session,
+      task,
+      max_steps~,
+      thinking~,
+      reasoning_effort~,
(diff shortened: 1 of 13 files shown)


bobzhang:
Codex CLI review (`codex review --base main`):

> The changes consistently thread the new thinking controls through the CLI, TUI, agent, and DeepSeek parsing APIs without introducing a clear regression. `moon test` only failed on the pre-existing live DeepSeek API smoke test due network/hostname resolution.

(The smoke-test failure is the `DEEPSEEK`-keyed realworld test running without network in Codex's sandbox.)

PR #83, opened 2026-06-10
feat(openseek): list sessions with last activity and first prompt

bobzhang:
Sixth item off the `improvement.md` checklist.

## Change

- **`SessionStore::listings()`**: one row per session — id, last activity (event-log mtime, falling back to the header for never-appended sessions), and the first user prompt — ordered newest first. Unreadable sessions stay in the listing (no timestamp, empty prompt, sorted last): a listing is exactly where a user finds the broken directory they may want to clean up, in contrast to `latest()` which must skip them.
- **`--session-list`** prints tab-separated rows: id, UTC minute stamp (`-` when unknown), and the prompt collapsed to a 60-char one-line label (`-` when none). Tab separation keeps it scriptable — `| cut -f1` recovers the old bare-id output, which is what the cram test now asserts.

Live output:

```
tui-20260610-053052-148	2026-06-10 05:30	Write one short sentence about labels, then call the finish …
tui-20260610-031529-831	2026-06-10 03:16	Remember this: my name is Hongbo. Reply with just the word o…
```

## Verification

`moon check` 0 warnings, fmt clean, 441/441 tests (listing order/labels/unreadable-visibility, label collapsing/truncation, timestamp rendering), 6/6 offline cram.

### DIFF
--- a/agent_session/store/pkg.generated.mbti
+++ b/agent_session/store/pkg.generated.mbti
@@ -11,6 +11,13 @@ import {
 // Errors
 
 // Types and methods
+pub struct SessionListing {
+  // private fields
+} derive(@debug.Debug)
+pub fn SessionListing::first_prompt(Self) -> String
+pub fn SessionListing::id(Self) -> @agent_session.SessionId
+pub fn SessionListing::last_active_unix_seconds(Self) -> Int64?
+
 pub struct SessionStore {
   // private fields
 } derive(@debug.Debug)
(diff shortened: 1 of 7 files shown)


bobzhang:

**Round 1** flagged a P2 — `listings()` deserialized every event of every session just to label rows, making `--session-list` scale with total history size. Addressed in 662c699: `first_prompt_in_log` streams the log line-by-line and stops at the first user event (normally line one, capped at 50); recency comes from stat alone, and loadability stays `latest()`'s gate.

**Round 2** flagged another P2 — prompt labels were printed verbatim, so a stored prompt containing ESC/OSC sequences could drive the terminal from a mere listing. Addressed in aefb28f: C0 controls, DEL, and C1 controls are stripped from labels, with a test covering ESC/BEL/C1-CSI.

**Round 3** (after both fixes):

> No actionable correctness issues were found in the changed code. The new session listing behavior is implemented consistently with the updated tests and generated interface.

PR #84, opened 2026-06-10
docs: document the TUI and its durable-session default in the README

bobzhang:
Seventh item off the `improvement.md` checklist.

The root README never mentioned `cmd/tui`. This adds the packages-table row and a **Terminal UI** section: how the TUI drives the engine per prompt and renders its event stream (live streaming line, `✻` thought asides), why every launch converses in a generated durable session, and the resume surface (`--continue`, `--session <id>`, enriched `--session-list`). The Agent CLI section also gains a line for the new thinking flags.

Docs-only; `moon check`/`moon test` green (the README's code blocks are unchanged).


bobzhang:

**Round 1** flagged two doc-accuracy issues: the resume examples named `openseek-tui` (only the argparse display name — the installed binary is `tui`, so the documented workflow failed before reaching the app), and the packages-table row pointed at the reusable `tui` library docs instead of command docs.

Addressed in 10d5c05: examples use the `moon run cmd/tui --` form used elsewhere in the README, and `cmd/tui` now has its own README (command behavior, engine contract, session workflow) that the table row points to.

**Round 2** (after the fix):

> The changes are documentation-only and align with the current TUI/session behavior and CLI flags. I did not find any actionable correctness issues introduced by this patch.

PR #85, opened 2026-06-10
feat(agent): steerable turns in a caller-owned scope

bobzhang:
**PR 1 of 3** for the persistent-engine design (approved in review discussion; targets the "persistent engine per session" + "steering" items in `improvement.md`). PR 2 adds the engine's `--serve` command loop; PR 3 the TUI client.

## Change

- **`run_turn_in_scope`** (was private `run_with_runtime`): one turn inside a **caller-owned** `AgentRuntime` + task scope, generic over the scope's result type. A serve-mode engine holds both across turns — that's what keeps `moon_check` watchers warm between prompts, and what makes steering reachable.
- **`steer(runtime, text)`**: queues mid-turn user input on the existing runtime event bus (`extenum AgentEvent += SteerInput`). The loop folds it in at the **next step boundary** — the same place runtime updates land, ordered before them so the user's redirection reads first — appends it durably as a `User` message, and emits a `steer_applied` JSONL event. Blank input is dropped. Steering an idle runtime parks the input for the next turn.
- **`api_url?`** threaded through all run entry points, making agent turns testable against the mock SSE server pattern from `deepseek/client`.

## Verification

- New blackbox tests run a **full agent turn** against a scripted SSE server: the server asserts the steered text arrives in the request body alongside the task, and the session ends `[User(task), User(steer), Terminal(Finished)]`; a second test pins that blank steers leave no trace.
- `moon check` 0 warnings, fmt clean, 443/443 tests, 6/6 offline cram. No behavior change for existing callers (all new params optional, defaults preserved).

### DIFF
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -2,24 +2,34 @@
 let default_max_steps : Int = 1000
 
 ///|
-/// Mid-turn user input riding the shared runtime event bus. Queued by
-/// `steer` and folded into the conversation by the agent loop at its next
-/// step boundary.
-extenum @agent_runtime.AgentEvent += {
-  SteerInput(String)
-}
-
-///|
 /// Queue `text` as steering input for the turn running on `runtime`.
 ///
 /// Steering lands at the next step boundary — after the in-flight model call
 /// and its tool batch complete, where the loop already folds in runtime
 /// updates — so completed work is never thrown away. The text is appended to
 /// the session as a user message and a `steer_applied` event is emitted when
-/// it takes effect. Steering an idle runtime is harmless: the input waits on
-/// the bus for the next turn.
+/// it takes effect. Steering an idle runtime is harmless: the input waits in
+/// the queue for the next turn. The channel is lossless (unlike the runtime
+/// event bus, which discards under flood), so a steer can never be pushed
+/// out by noisy watcher updates.
 pub fn steer(runtime : @agent_runtime.AgentRuntime, text : String) -> Unit {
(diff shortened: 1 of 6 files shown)


bobzhang:
Codex CLI review trail (`codex review --base main`) — three rounds, two substantive catches:

**Round 1** flagged two P2s:
1. *Tool state not actually shared*: `tool_definitions` was rebuilt per call and the `moon_check` watcher registry lives in that closure, so the shared-scope doc's "warm watchers" promise was false — turn 2 would orphan turn 1's watchers and start duplicates. → b7bd0df: `build_tools()` is public and `run_turn_in_scope` takes `tools?`; serve-mode callers build once and pass per turn.
2. *Steering rode the lossy bus*: the runtime event queue is `DiscardOldest(32)`, so a watcher flood could push a queued steer out silently. → b7bd0df: `AgentRuntime` gains a dedicated **unbounded** steering queue (`queue_steer`/`drain_steers`); a flood test pins the guarantee.

**Round 2** flagged the terminal race: a steer arriving while the *final* model call is in flight parked until an unrelated future prompt. → 4fef7f2: **user input wins over model-initiated completion** — pending steers at natural completion or `finish` keep the turn alive (batch closed out for API validity); error terminals still end the turn, documented in `steer()`'s contract. A mock-server test races a steer against the final call and asserts the turn continues.

**Round 3**:

> I found no actionable correctness issues in the diff. The changes compile and the targeted agent/runtime tests pass in this environment.

PR #86, opened 2026-06-10
feat(openseek): add --serve, a long-lived session server on stdio

bobzhang:
**PR 2 of 3** for the persistent-engine design (PR 1: #85, merged). PR 3 wires the TUI to this.

## Change

`openseek --serve` turns the engine into a session server speaking line-delimited JSON on its standard streams:

- **stdin commands**: `{"command":"prompt","text":…}` starts a turn; `steer` injects input into the running turn via the lossless steering channel (idle steer = next prompt); `cancel` interrupts the turn — never the process. Malformed commands surface as `command_error` events instead of killing the session. stdin EOF = graceful shutdown (running turn finishes, queue drains, exit 0).
- **stdout**: the existing JSONL event stream, plus `turn_failed` for a turn that died with an error (so a controller can close the run without the process exiting; cancelled turns emit `agent_aborted`).
- One `AgentRuntime` + one `build_tools` registry span every turn — the PR-1 plumbing that keeps `moon_check` watchers warm across prompts. The session value updates on every durable append, so even a failing turn leaves the in-memory conversation consistent with the store.

## Verification

- Live, real API, **one process** for all of: turn 1 stores a name → turn 2 recalls it (`Hongbo`) → turn 3 is steered mid-run (`steer_applied` emitted, answer ends with the steered `BANANA`) → turn 4 is cancelled (`agent_aborted`, process alive) → turn 5 still serves → EOF exits 0.
- Offline cram covers the command surface without network: malformed command → `command_error`, idle cancel no-op, EOF exit 0, and the task-positional rejection.
- `moon check` 0 warnings, fmt clean, 449/449 tests (command decode), 8/8 offline cram.

### DIFF
--- a/cmd/openseek/main.mbt
+++ b/cmd/openseek/main.mbt
@@ -25,6 +25,13 @@ async fn run_cli() -> Unit {
     if handle_session_command(matches) {
       return
     }
+    if flag(matches, "serve") {
+      if values(matches, "task").length() > 0 {
+        fail("--serve reads commands from stdin; it does not take a task")
+      }
+      run_serve(matches)
+      return
+    }
     let api_key = api_key(matches)
     let model = @deepseek.Model::parse(value(matches, "model"))
     let max_steps = max_steps(matches)
@@ -173,6 +180,11 @@ fn cli_command() -> @argparse.Command {
       ),
     ],
     flags=[
+      FlagArg(
+        "serve",
+        long="serve",
+        about="Run as a session server: read JSONL commands (prompt/steer/cancel) from stdin.",
+      ),
       FlagArg(
         "session-list",
         long="session-list",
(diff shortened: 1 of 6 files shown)


bobzhang:
Codex CLI review trail (`codex review --base main`) — three rounds, two substantive catches:

**Round 1**: `turn_failed` was emitted but unknown to the event decoder — a controller would drop it and leave the run stuck while the server moved on. → 598c0b4: the wire vocabulary and its decoder land together; `turn_failed` maps to a terminal `AgentError`.

**Round 2**: the stdin reader acted on steer/cancel by inspecting `active_turn` directly, racing the serve loop — `prompt\nsteer\n` in one pipe buffer could turn the steer into a second prompt. → 9ae3a2e: all scheduling flows through one ordered queue with one consumer (wire commands + internal `TurnDone`/`Shutdown` markers); the consumer alone owns turn state. Live-verified with a single-buffer batched prompt+steer: `steer_applied`, steered word in that turn's answer, no spurious second turn.

**Round 3**:

> No actionable correctness issues were identified in the diff. The targeted package tests and cram tests pass; the only full-suite failure observed is an existing live DeepSeek network smoke test unrelated to this change.

PR #88, opened 2026-06-10
feat(tui): drive one persistent serve-mode engine per session

bobzhang:
**PR 3 of 3** for the persistent-engine design (PR 1: #85, PR 2: #86 — both merged). Completes the feature and ticks the "persistent engine per session" and "steering a running task" checklist items.

## Change

- **`EngineClient`**: spawns `openseek --serve` on first use and keeps it for the session. Commands ride stdin (`prompt`/`steer`/`cancel`); one long-lived reader decodes the event stream and stamps events with the run id of the most recently submitted input — sound because turns are strictly sequential on the wire (a new prompt is only written after the previous terminal event already precedes it in the stream). A `run_open` flag owned by the client (not inferred from the stream) ensures an engine that dies before emitting anything still fails its run, with the capped stderr prefix as diagnostics. Next prompt respawns from the durable session.
- **Steering UX**: Enter while a task runs steers it (previously a "not supported" error). The transcript echo comes from the engine's `steer_applied` event, so the input appears at the conversational position the model actually saw it. Shell commands can't steer and say so.
- **Ctrl-C ladder**: first press sends the polite `cancel` command (turn dies, engine lives); second press kills the engine process. TUI exit closes the engine's stdin — the serve protocol's graceful EOF.

## Verification (live pty, real API — one TUI session)

- Engine PID **15223 constant** across two prompts, a mid-run steer, and a cancel — one process for everything.
- Turn 2 recalled turn 1's name (`Hongbo`) through the shared engine.
- A steer typed in the composer mid-run was echoed into the transcript at its true position and honored — the answer ended with the steered `PAPAYA`.
- Ctrl-C cancelled the turn (`aborted` in transcript), TUI and engine both survived, and the next prompt answered normally.
- `moon check` 0 warnings, fmt clean, 451/451 tests (steer command dispatch, steer_applied echo, interrupt escalation), 8/8 offline cram.

### DIFF
--- a/cmd/openseek/serve.mbt
+++ b/cmd/openseek/serve.mbt
@@ -72,10 +72,13 @@ test "serve commands decode and reject with actionable errors" {
 /// The runtime and the tool registry are built once and shared by every
 /// turn, so stateful tools survive turn boundaries (a `moon_check` watcher
 /// started in turn 1 keeps reporting in turn 5). `steer` rides the
-/// runtime's lossless steering queue into the running turn; when no turn is
-/// running it simply becomes the next prompt. `cancel` interrupts the
-/// current turn but never the process. stdin EOF is the shutdown signal:
-/// the running turn finishes first, then the loop drains and exits.
+/// runtime's lossless steering queue into the running (or already
+/// submitted) turn; with no turn to land in it is rejected with a
+/// `steer_dropped` event — that only happens by racing the terminal event
+/// through the pipes, and an engine must never start a turn the controller
+/// did not ask for. `cancel` interrupts the current turn but never the
+/// process. stdin EOF is the shutdown signal: the running turn finishes
+/// first, then the loop drains and exits.
 async fn run_serve(matches : @argparse.Matches) -> Unit {
   let api_key = api_key(matches)
   let model = @deepseek.Model::parse(value(matches, "model"))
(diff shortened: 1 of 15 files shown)


bobzhang:
Codex CLI review trail (`codex review --base main`) — ten rounds, eight substantive findings, all fixed with regression tests:

1. **Stale-reader race on respawn**: a dead engine's reader could mis-attribute its death to a respawned successor's run → run bookkeeping moved onto each `EngineProcess`, identity-guarded slot cleanup.
2. **Shell-command escalation**: second Ctrl-C during a `!` command killed the innocent engine; steering during one poked the idle engine → `run_is_command` gates both.
3. **Death-window steers**: a steer between engine death and the death report auto-started a fresh engine whose serve loop treated it as a phantom prompt → mid-turn controls never spawn engines; undeliverable steers surface as a resubmit notice.
4. **Stale command handle**: a prompt starting in the same update that finished a `!` command left the dead task handle hijacking Ctrl-C → cleared at `StartAgent` dispatch.
5. **Idle-steer protocol flaw**: a steer racing the terminal through the pipes reached an idle engine and became a phantom turn — unfixable client-side → protocol change: idle engines reject steers with `steer_dropped` (offline-cram covered).
6. **Stale-filtered drop notices**: `steer_dropped` always trails the racing terminal, so run-tagged wrapping let the stale-run guard eat the notice → routed as an untagged message.
7. **Engine contract break**: serve-by-default silently broke `--engine` replay engines → `--engine-mode serve|oneshot` escape hatch (per the original design sign-off), with the one-shot path restored.
8. **Interrupting-steer contamination + oneshot watcher status**: steers during cancellation could outlive the dying turn in the persistent runtime → TUI refuses them and the serve loop sweeps leftovers at `TurnDone`; oneshot runs clear watcher status at turn end again (their engine dies per prompt).

**Final round:**

> No actionable correctness issues were found in the reviewed diff. The changed MoonBit packages type-check, and targeted TUI tests pass.

Live verification (real API, one pty session, re-run after every fix wave): engine PID constant across two prompts + a mid-run steer + a cancel; memory recalled across turns; steer echoed at its true transcript position and honored in the answer; cancel survived; EOF exits 0.

PR #90, opened 2026-06-10
feat(tui): acknowledge a sent steer on the activity line

bobzhang:
UX follow-up to #88, from user feedback: after pressing Enter to steer, the composer cleared and **nothing confirmed the input was sent** until the step-boundary echo — many seconds later during a long tool call. A steer read as an eaten keystroke.

## Change

- `State.pending_steers` tracks input sent but not yet settled. The activity line immediately grows a dim ` · steering: <tail>` segment (with `×N` when several are in flight): `Thinking …project layout · steering: "…add a unit test"`.
- The segment clears exactly when the engine settles the steer: `steer_applied` (transcript echo), `steer_dropped` (resubmit notice), or turn end (engine sweeps leftovers).
- The segment has a bounded width (44 cols) and the streaming preview cedes exactly that much — the renderer clips the row's *end* on overflow, so without the reservation the acknowledgment would be the first thing cut.

## Verification

- Live pty (real API): the acknowledgment appeared **within 1.5s of Enter** — while the steered turn was still mid-`sleep` — the steer echoed later at its true transcript position, the segment cleared on echo, and the answer honored the steered word.
- Full persistent-engine pty suite re-passed (PID stability, memory, steer, cancel).
- 455/455 tests (pending-steer lifecycle: ack on send, clear on applied/dropped/turn-end), 9/9 cram, fmt clean.

Side discovery while verifying, recorded for posterity: `cp` over the executed `~/.moon/bin/openseek` corrupts the macOS per-inode signature cache — the binary gets SIGKILLed at exec (exit 137) with no stderr, masquerading as an engine protocol bug. Install with `rm` first.

### DIFF
--- a/cmd/tui/drain_wbtest.mbt
+++ b/cmd/tui/drain_wbtest.mbt
@@ -298,6 +298,44 @@ test "a running shell command is never steered and never kills the engine" {
 }
 
 ///|
+test "a sent steer is acknowledged until the engine settles it" {
+  let state = drain_test_state()
+  let _ = update(state, UiInput(Steer(Prompt("do it"))))
+  assert_eq(state.pending_steers.length(), 0)
+
+  // Enter mid-run: acknowledged immediately, before any engine round trip —
+  // this is the only feedback between the keypress and the transcript echo.
+  let _ = update(state, UiInput(Steer(Prompt("make it rhyme"))))
+  assert_eq(state.pending_steers.length(), 1)
+  guard pending_steer_segment(state.pending_steers) is [_, _] else {
+    fail("expected an activity-line acknowledgment")
+  }
+
+  // The engine confirms: acknowledgment clears, transcript echo appends.
+  let _ = update(
+    state,
+    AgentProgress(run_id=state.run_id, SteerApplied("make it rhyme")),
+  )
+  assert_eq(state.pending_steers.length(), 0)
+  assert_eq(pending_steer_segment(state.pending_steers).length(), 0)
+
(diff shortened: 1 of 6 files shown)


bobzhang:
Codex CLI review trail (`codex review --base main`) — three rounds, three substantive findings, all fixed with regression tests:

1. **Cross-turn mis-confirmation**: an untagged `steer_dropped` trailing a finished turn could consume the *next* turn's fresh acknowledgment → settlement matched by text instead of position.
2. **Phantom acknowledgments**: whitespace-only steers are filtered engine-side with no settling event, so their entry could never clear (and would mis-settle a later real steer) → blank steers are neither sent nor tracked.
3. **Identical-text race**: even text-matching could let a stale wire drop steal a fresh identical-text entry → split the message by provenance. A wire drop's own turn-terminal always precedes it in the engine's ordered stream, so its entries are guaranteed already swept: `EngineSteerDropped` shows the notice and settles nothing, while the client-side `SteerDropped` (entry created an instant ago, cannot have been swept) settles by text. The race is now unrepresentable.

**Final round:**

> No actionable correctness issues were identified in the diff. The steering acknowledgment state is added consistently across update, engine-client routing, and status rendering paths.

Live verification (real API, pty): the ` · steering: …` acknowledgment appeared within 1.5s of Enter — mid-`sleep`, long before the step boundary — the steer echoed later at its true transcript position, the segment cleared on echo, and the answer honored the steered word. Full persistent-engine suite re-passed.

PR #91, opened 2026-06-10
feat(viz): events.jsonl session visualizer

bobzhang:
## What

A browser viewer for OpenSeek's durable session logs (`events.jsonl` + `session.json`), built as a monorepo feature: a JS frontend backed by a native MoonBit web server, sharing the existing `agent_session` types as the contract.

## Pieces

| Package | Target | Role |
|---|---|---|
| `viz/` | js | Pure parse + render. Reuses `agent_session` decoders and the real `Session::chat_messages` projection. Renders a **Raw log** (turn-grouped by terminal) and a **Model view**; tolerant of truncated/corrupt logs. |
| `cmd/viz_server/` | native | Read-only web server (`moonbitlang/async/http`) exposing a JSON/raw-file API over a `SessionStore`. Never writes. |
| `cmd/viz_app/` | js | rabbita (TEA) frontend: session browser + Raw/Model toggle. |
| `web/index.html` | — | The viewer shell (server auto-locates the built bundle). |

## UX

- Left sidebar lists sessions (most-recent first); click to load.
- Header strip shows id, event count, `session.json` status, and any parse warnings.
- **Raw log**: every event grouped into turns (delimited by `Terminal`, so mid-turn steering stays in one turn), with per-kind cards (user / assistant + reasoning + tool calls / tool result / runtime / summary / terminal badge).
- **Model view**: exactly what `chat_messages` feeds the model — summaries substitute covered events, dangling tool calls show the synthesized marker.

## API additions

- `AssistantMessage::{reasoning_content, tool_calls}` (read accessors)
- `SessionStore::{event_log_file, header_file}` (paths for the read-only server)

## Running

```
moon build cmd/viz_app --target js
moon run cmd/viz_server --target native -- --session-root .openseek
# open http://127.0.0.1:8080
```

## Testing

- Unit/snapshot tests in `viz`, `cmd/viz_app`, `cmd/viz_server` (parsing, both render modes, turn grouping, percent-decode, id validation, envelope decoding).
- Verified end-to-end against a real 355-event session via headless Chrome (sidebar → load → both view modes; rendered counts match the on-disk data).
- Hardened against early client disconnect, invalid/encoded ids, empty session root, bad port, header-only sessions, and torn-UTF-8 tails.

> Note: CI is native-only and skips the JS packages (`viz`, `cmd/viz_app`); those are covered by the local unit tests and the browser E2E above.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ _build/
 .mooncakes/
 .moonagent/
 .repos/
+web/viz_app.js
 eval/*
 !eval/README.md
 !eval/prompts/
--- a/agent_session/pkg.generated.mbti
+++ b/agent_session/pkg.generated.mbti
@@ -18,6 +18,8 @@ pub struct AssistantMessage {
 } derive(ToJson, @debug.Debug, @json.FromJson)
 pub fn AssistantMessage::AssistantMessage(String, tool_calls? : Array[@deepseek.ToolCall], reasoning_content? : String) -> Self
 pub fn AssistantMessage::content(Self) -> String
+pub fn AssistantMessage::reasoning_content(Self) -> String?
+pub fn AssistantMessage::tool_calls(Self) -> Array[@deepseek.ToolCall]
 
 pub struct RuntimeNotice {
   // private fields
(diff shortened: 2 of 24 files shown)


bobzhang:
cc @tonyfettes

PR #92, opened 2026-06-11
fix(viz): locate web assets from the executable's own path

bobzhang:
Fixes the report: `moon run ../openseek/cmd/viz_server --port 9093` from another project's directory cannot find `web/index.html` (nor `viz_app.js`).

## Cause

`moon run` keeps the invoking shell's CWD — which is the *feature* for `--session-root` (you serve the sessions of the project you're standing in) — but every asset default (`--web-dir web`, the `_build/js/...` bundle fallbacks) was CWD-relative too, so they resolved in the wrong tree.

## Fix

The binary always lives at `<root>/_build/<target>/<profile>/build/cmd/viz_server/...`, so the checkout root is recoverable from `argv[0]` (everything before the `_build` component; works for absolute, `../openseek/_build/...`, and bare `_build/...` forms). Asset lookups become ordered candidate lists: the configured CWD-relative paths first — in-checkout behavior and explicit `--web-dir`/`--bundle` flags take precedence unchanged — then the same paths under the executable-derived root. `--session-root` is untouched.

An installed binary outside any moon build tree derives nothing and behaves exactly as before; the 404 bodies now name the fix (`pass --web-dir or run from the openseek checkout`).

## Verification

- Live, replicating the report: from a foreign directory containing only a `.openseek`, `moon run <openseek>/cmd/viz_server -- --port 9193` with **no asset flags** → `GET /` 200 (8.9 KB shell), `GET /viz_app.js` 200 (659 KB bundle), `GET /api/sessions` lists the *foreign* directory's sessions.
- In-checkout run unchanged (both routes 200 via the CWD-relative candidates).
- wbtests pin the root derivation (absolute / sibling-relative / bare-relative / non-moon paths) and the candidate ordering, including `--bundle` staying highest-priority. 464/464 tests, 9/9 cram, fmt clean, 0 warnings.

### DIFF
--- a/cmd/viz_server/main.mbt
+++ b/cmd/viz_server/main.mbt
@@ -19,14 +19,20 @@ async fn main {
   }
   let web_dir = value(matches, "web-dir")
   let bundle = value(matches, "bundle")
+  let module_root = match @env.args() {
+    [exe, ..] => module_root_from_exe(exe)
+    [] => None
+  }
+  let shell = shell_candidates(web_dir, module_root)
+  let bundle_paths = bundle_candidates(bundle, web_dir, module_root)
   let store = @store.SessionStore(session_root)
   let addr = @socket.Addr::parse("127.0.0.1:\{port}")
   println("openseek viz: serving sessions under '\{session_root}'")
-  println("openseek viz: shell from '\{web_dir}/index.html'")
+  println("openseek viz: shell from '\{shell[0]}'")
   println("openseek viz: open http://127.0.0.1:\{port}")
   let server = @http.Server(addr)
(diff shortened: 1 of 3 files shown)


bobzhang:

> No actionable correctness issues were found in the changed MoonBit server asset resolution logic or tests.

Live verification: from a foreign directory containing only a `.openseek`, the server (run via `moon run` with no asset flags) served `GET /` 200 and `GET /viz_app.js` 200 through the executable-derived checkout root, while `/api/sessions` listed the foreign directory's own sessions; the in-checkout run still resolves through the CWD-relative candidates.

PR #93, opened 2026-06-11
docs: add agent findings from a real 156-step session log

bobzhang:
Checklist additions from analyzing a real agent run (`seek-toml`, 2026-06-11: "Write a toml parser in MoonBit with high quality tests" — succeeded in 156 steps / 19 minutes, 26 tests passing). Documentation-only; each item cites what the log showed. Details in the PR discussion.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #94, opened 2026-06-11
feat(session): stamp every appended event with wall-clock ts

bobzhang:
Checklist item from the [156-step log analysis](https://github.com/moonbitlang/openseek/pull/93#issuecomment-4676980288): session event logs carry no timestamps, so per-step latency can't be reconstructed (the analysis had to infer a 19-minute duration from file mtimes).

## Change

- `SessionEvent` gains a **required** `ts` stamp (unix milliseconds); `SessionStore::append`/`compact` set it from the wall clock. Lines become `{"sequence":N,"ts":<unix-ms>,"item":…}`. In-memory appends without a clock (serve's ephemeral sessions, unit tests) carry the 0 sentinel.
- **Why on the event, not the write path**: the header's `event_log_length`/`event_log_hash` durability markers cover the exact serialized log text, so the appended line and every reserialization (`write_header`, stale-snapshot comparison, compaction rewrite) must agree byte-for-byte.
- **Why required, not optional**: there are no external users, so pre-stamp logs aren't worth a compatibility codec — old local sessions are simply re-created (explicit owner decision). Required keeps the derived codec; `ts` is `Double` so it serializes as a JSON number (core encodes `Int64` as a string), exact for milliseconds until ~2255.
- **Visualizer**: each raw-log card now shows its offset into the turn (`+0.0s`, `+1.1s`, `+2m 05.5s`) and each turn summary its total duration — relative offsets answer "where did the time go" directly, without timezone mental math. Sentinel (unstamped) events render no time chips. Model view is untouched (the projection carries no timestamps).

## Verification

- Live engine run: every appended line stamped — the turn's 1.1s is measurable from the log alone — and the viz server serves the stamped envelope (`/`, `/viz_app.js`, `/api/sessions/<id>` all green against the new log).
- 465/465 native tests (JSON round trip pins the exact line shape; store stamp survives reload) and 114/114 js tests (offset and duration rendering, sentinel suppression), 9/9 cram (the compact demo writes sentinel stamps and `sed`-strips the live one), fmt clean, 0 warnings.

### DIFF
--- a/agent_session/pkg.generated.mbti
+++ b/agent_session/pkg.generated.mbti
@@ -31,10 +31,10 @@ pub struct Session {
   // private fields
 } derive(@debug.Debug)
 pub fn Session::Session(SessionId, system_prompt~ : String, events? : @vector.Vector[SessionEvent]) -> Self
-pub fn Session::append(Self, SessionItem) -> Self
-pub fn Session::append_event(Self, SessionItem) -> (Self, SessionEvent)
+pub fn Session::append(Self, SessionItem, unix_ms? : Int64) -> Self
+pub fn Session::append_event(Self, SessionItem, unix_ms? : Int64) -> (Self, SessionEvent)
 pub fn Session::chat_messages(Self) -> Array[@deepseek.ChatMessage]
-pub fn Session::compact(Self, content~ : String, from_sequence~ : Int, to_sequence~ : Int) -> Self raise
+pub fn Session::compact(Self, content~ : String, from_sequence~ : Int, to_sequence~ : Int, unix_ms? : Int64) -> Self raise
 pub fn Session::events(Self) -> @vector.Vector[SessionEvent]
 pub fn Session::id(Self) -> SessionId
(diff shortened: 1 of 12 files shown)


bobzhang:

> No actionable bugs were found in the timestamp persistence and visualization changes. Targeted package tests passed; the only full-suite failure observed was the network-dependent DeepSeek smoke test.

(The smoke-test failure is the pre-existing live-network example, unrelated to this change.)

Design note for the record: the first draft made `ts` optional with a manual codec so pre-stamp logs reserialized byte-identically; the owner ruled backward compatibility unnecessary (no external users), which simplified to a required field with the derived codec — old local sessions are simply re-created.

Live verification: a real engine run writes stamped lines (`"ts":1781151318110`, the turn's 1.1s measurable from the log alone); the viz serves the stamped envelope and renders per-card turn offsets and turn durations (js tests pin `+0.0s` / `+1.1s` / `+2m 05.5s` and sentinel suppression).

PR #95, opened 2026-06-11
fix(deepseek): never send reasoning_content back to the API

bobzhang:
Checklist item from the [156-step log analysis](https://github.com/moonbitlang/openseek/pull/93#issuecomment-4676980288): stored reasoning was echoed on every resent assistant message — 64KB of dead context (11% of the final 141K tokens) the model never reads back, and DeepSeek's docs say not to return it.

## Change

One central removal in the **wire encoder** (`deepseek/json_encode.mbt`): `ChatMessage::to_json` never emits `reasoning_content`. The encoder is the right chokepoint rather than the session→messages projection: the measured waste accrued *within* a single 156-step turn through the agent loop's incremental message pushes, not only across turn boundaries — fixing only the projection would have left the dominant path leaking.

Responses still parse `reasoning_content`, sessions still store it, and the TUI transcript (`✻` asides) and the visualizer still display it — it is response-side data end to end now.

## Verification

- Live, thinking mode, multi-step tool turn (the risky case — steps 2–3 resend prior assistant messages now stripped of reasoning mid-loop): 3 steps, `agent_finished`, reasoning still streamed to the UI, still stored in the session log.
- Encoder test updated to pin the absence: an assistant message carrying reasoning in memory encodes without it on the wire.
- 465/465 native + 114/114 js tests, 9/9 cram, fmt clean, 0 warnings.

### DIFF
--- a/deepseek/json_encode.mbt
+++ b/deepseek/json_encode.mbt
@@ -3,6 +3,12 @@
 ///
 /// Assistant messages whose `content` is empty but carry tool calls are
 /// encoded with JSON `content: null`, as the API requires.
+///
+/// `reasoning_content` is deliberately never encoded: it is response-side
+/// data the model does not read back (DeepSeek's docs say not to return it),
+/// and resending it grows every request — a real 156-step run accumulated
+/// 64KB of dead reasoning, 11% of its final 141K-token context. Sessions
+/// still store it for the TUI transcript and the visualizer.
 impl ToJson for ChatMessage with fn to_json(message : ChatMessage) -> Json {
   let content = match message.role {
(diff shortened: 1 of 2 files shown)


bobzhang:

> No actionable correctness issues were found in the diff. The focused deepseek tests pass; the full suite only failed on a network-dependent realworld smoke test unrelated to this change.

Live verification (the risky case was within-turn tool loops, where the agent resends this turn's own assistant messages): a 3-step thinking-mode tool turn completed normally with reasoning stripped from the resent history — `agent_finished`, reasoning still streamed to the UI, still stored in the session log for the transcript and visualizer.

PR #96, opened 2026-06-11
fix(moon_check): one watcher per directory, replace on option change

bobzhang:
Top item from the [156-step log analysis](https://github.com/moonbitlang/openseek/pull/93#issuecomment-4676980288) — the one failure that made the agent fight its own tooling: keying watchers by the full option set let a `deny_warn` toggle stack **three concurrent watchers on one cwd**, which fought each other (and the model's own `moon test` runs) for moon's build lock until results went stale and the model ran `pkill -9 -f "moon check"` mid-task to recover.

## Change

- The registry keys by **cwd** — one watcher per directory is the invariant.
- A call whose rendered command (`moon_check_command_line`, also used at spawn time so the fingerprint can't drift from reality) matches the running watcher **reuses** it, exactly as before.
- A call with different options **stops the old process and starts a fresh one** — `watcher=replaced`, with the superseded command recorded in the restart reason so the transcript shows what changed.
- Stopped records restart with the new options (unchanged behavior), and the tool description now tells the model about the replacement semantics.

## Verification

- New regression test drives real `moon check --watch` processes through started → replaced → reused on one fixture cwd: the replacement owns the key with the new command (`--deny-warn`), gets a fresh id, and the registry holds exactly one record throughout.
- 466/466 native + 114/114 js tests, 9/9 cram, fmt clean, 0 warnings.

### DIFF
--- a/agent_tool/moon_check/moon_check.mbt
+++ b/agent_tool/moon_check/moon_check.mbt
@@ -101,7 +101,14 @@ async fn[X] moon_check(
           is_error=moon_check_snapshot_is_error(snapshot),
         )
       } else {
+        let exit_signal = runtime.moon_check_exit_signal(key)
         ignore(runtime.stop_moon_check(key))
+        // `Process.cancel()` is asynchronous: wait until the old watcher's
+        // monitor has reaped it before spawning the replacement, or the two
+        // would briefly contend for the same build lock.
+        if exit_signal is Some(signal) {
+          signal.get()
+        }
         start_moon_check(
           runtime,
           input,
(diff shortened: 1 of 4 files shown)


bobzhang:
Codex CLI review trail (`codex review --base main`) — two rounds, one substantive catch:

**Round 1**: `Process.cancel()` is asynchronous, so the replace branch could spawn the successor while the old child was still dying — a brief overlap on the same build lock, exactly what replacement exists to prevent. → 3320901: each spawn carries a reaped signal its monitor task fires on return (every exit path has reaped or abandoned the child by then); the replace branch captures the old record's signal before stopping and awaits it before starting the successor. The replace regression test doubles as the canary — it would hang on broken wiring.

**Round 2:**

> The changes correctly replace an existing watcher when options change for the same cwd, and the targeted moon_check tests pass. No blocking regressions were identified in the diff.

PR #97, opened 2026-06-11
fix(agent): skip watcher notices whose substance is unchanged

bobzhang:
Next item from the [156-step log analysis](https://github.com/moonbitlang/openseek/pull/93#issuecomment-4676980288): 61 `[moon_check update]` notices — 11% of the final 141K-token context — were appended, mostly identical re-reports from quiet watchers. Only the `events=` batch total and per-watcher `seq=` chunk counter moved, which made every notice textually unique and defeated any equality check.

## Change

- `@moon_check.update_fingerprint` extracts a notice's substance: the update text minus the volatile counter lines. A unit test pins that counter-only churn fingerprints equal while a diagnostics change does not.
- The agent loop keeps the previous step's fingerprint and, on a match, skips the append entirely — session event, model message, and `runtime_update` wire event alike. The model already has that state; re-sending it only grows the context. Real changes (new diagnostics, status flips, restarts) append immediately, and the TUI status bar keeps its last value since nothing changed.

## Verification

- Live agent run (real API) with a watcher held active across multiple sleep steps: exactly **one** notice appended where each step previously re-appended; the appended-notices-are-all-distinct invariant checked against the session log.
- 467/467 native + 114/114 js tests, 9/9 cram, fmt clean, 0 warnings.

### DIFF
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -243,7 +243,8 @@ pub async fn[X] run_turn_in_scope(
         messages,
         append_item~,
       )
-      if runtime_update_message(runtime.drain_events()) is Some(update) {
+      let runtime_events = runtime.drain_events()
+      if runtime_update_message(runtime_events) is Some(update) {
         // A batch in which every watcher reproduces its previous notice
         // (volatile counters aside) is collapsed to a one-line marker: the
         // model still learns its edit was processed with an unchanged
@@ -251,7 +252,7 @@ pub async fn[X] run_turn_in_scope(
         // 156-step session appended 61 near-identical update notices, 11%
         // of its final context. Fingerprints are tracked per watcher so
         // alternating multi-cwd batches cannot thrash the comparison.
-        let sections = @moon_check.update_section_fingerprints(update)
+        let sections = @moon_check.update_fingerprints(runtime_events)
         let mut fresh = sections.is_empty()
         for pair in sections {
           let (key, fingerprint) = pair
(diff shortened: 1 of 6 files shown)


bobzhang:
Codex CLI review trail (`codex review --base main`) — five rounds, four substantive catches:

1. **Silent suppression swallowed the rerun signal**: the model wouldn't learn its edit was processed when results were unchanged → unchanged batches collapse to a one-line marker (`watchers re-ran; results unchanged…`) instead of vanishing, and the TUI only lifts watcher status from updates that carry one, so the status-less marker can't blank the segment. Live-verified: two comment-append edits → one full update + one compact marker.
2. **Turn-wide fingerprint thrashing**: alternating multi-cwd batches would each "differ" from the previous, defeating suppression → fingerprints track per watcher.
3. **Stale generated interface**: `.mbti` advertised the removed helper → regenerated.
4. **Blank-line phantom sections**: watcher output containing blank lines defeated text-level sectioning — and a dropped paragraph could be wrongly suppressed as unchanged → fingerprints now come from the drained snapshots themselves (`update_fingerprints`, coalesced exactly like the rendered block, keyed by the registry key), never from re-parsing text. Tests pin seq-churn stability, blank-line outputs staying one section, paragraph removal registering as change, and per-watcher keying.

**Final round:**

> No actionable correctness issues were found in the diff. `moon check --no-render` passed; `moon test` only failed the network-dependent DeepSeek smoke test due DNS/network restrictions.

PR #98, opened 2026-06-11
docs: add findings from the 5-run concurrent benchmark batch

bobzhang:
Two checklist additions from the batch validation run (5 concurrent reruns of the PR #93 benchmark; full numbers in the PR discussion): missing environment grounding in the prompt (caused a cross-run workspace hijack), and watcher notice payload size now dominating notice volume post-dedup.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #100, opened 2026-06-11
feat(openseek): ground the model in its working directory

bobzhang:
Checklist item from the [batch benchmark findings](https://github.com/moonbitlang/openseek/pull/98#issuecomment-4677851061): the prompt never states the working directory, so models guess on step 1 — one agent hallucinated `cwd="/workspace"`, then explored the parent directory and **adopted a concurrent sibling run's workspace** (two agents co-editing one project); another opened with `read /home/user`.

## Change

`configured_system_prompt` appends an `## Environment` section: the engine's realpath cwd, "treat it as the workspace root and stay inside it", "parent and sibling directories belong to other projects", and "never pass a cwd you have not verified exists". It rides every prompt source including `--system-prompt-file` (grounding describes the process, not the prompt content). Durable sessions record the prompt at creation, so the named directory is where the session started.

## Verification plan (merge gated on the experiment)

**Held unmerged pending a concurrency-5 benchmark with this engine** — judging on: zero hallucinated step-1 paths (vs 2/5 pre-fix), zero cross-workspace writes (vs one hijack in 5), artifacts in their own directories, plus the usual cost metrics. Results will be posted here.

Unit coverage: the section names the real cwd; the prompt-file replace/append composition and session creation tests pin the grounded form. 468/468 tests, 9/9 cram, fmt clean.

### DIFF
--- a/cmd/openseek/main.mbt
+++ b/cmd/openseek/main.mbt
@@ -440,16 +440,48 @@ async fn configured_system_prompt(
   } else {
     @fs.read_file(base_path).text()
   }
+  let grounded =
+    $|\{base}
+    $|
+    $|\{environment_prompt_section()}
   if addendum_path == "" {
-    base
+    grounded
   } else {
     let addendum = @fs.read_file(addendum_path).text()
     (
-      $|\{base}
+      $|\{grounded}
       $|
       $|\{addendum}
     )
   }
+}
+
+///|
+/// The runtime environment block appended to every system prompt — including
+/// a `--system-prompt-file` override, since grounding describes the process,
+/// not the prompt content. Without it the model has no idea where it is and
+/// guesses: in a five-run benchmark batch one agent passed
+/// `cwd="/workspace"` on its very first call, then explored the parent
+/// directory, found a concurrent sibling run's files, and adopted that
+/// workspace wholesale. A durable session stores the prompt at creation, so
+/// the recorded directory is where the session was started — resume it from
+/// the same project root.
+async fn environment_prompt_section() -> String {
+  let cwd = @fs.realpath(".") catch { _ => "." }
+  (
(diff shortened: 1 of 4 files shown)


bobzhang:
## Merge-gate experiment results (concurrency=5, grounded engine, same benchmark prompt)

| criterion | ungrounded baseline | grounded |
|---|---|---|
| hallucinated step-1 paths (`/workspace`, `/home/user`) | 2 / 5 runs | **0 / 5** |
| cross-workspace contamination | 1 full sibling-workspace hijack | **none** (one benign `/tmp/probe4.mbt` scratch file) |
| artifacts in their own directory | 4 / 5 | **5 / 5** |
| outcomes | 5/5 finished | 5/5 finished (steps 92–162, 16–34 min) |
| artifact tests | one co-authored artifact | 59/59, 37/37, 37/37, 45/48, 38/38 |

Qualitative shift: three of five grounded runs *opened* by reading their own absolute working directory, and a fourth passed its exact cwd explicitly — the models now demonstrably know where they are instead of guessing. The one imperfect artifact (run4, 3 failing tests) is model variance unrelated to grounding.

Codex CLI review trail — three rounds, two substantive catches, both in the eval harnesses this PR could have silently broken:
1. Trials ran with `cwd=repo_root` while staging workspaces elsewhere — grounding would have named the OpenSeek checkout as the workspace root, defeating per-trial isolation → trial engines now run inside their staged workspaces (which also exposed and fixed the file_edit harness still invoking the long-gone `cmd/main` package).
2. With the child cwd changed, launcher-relative paths (`--repo-root .`, prompt files) would resolve wrongly → canonicalized at the launch site.

> No actionable regressions were identified in the changed code. The modified packages type-check and targeted tests for the touched areas pass.

PR #102, opened 2026-06-11
perf(openseek): order the prompt for prefix caching, environment last

bobzhang:
Follow-up to #100 from review discussion: DeepSeek's context cache matches on request *prefixes* (64-token blocks), so prompt layout should put most-shared content first and per-run variables last.

## Change

The system prompt composes as `base → addendum → environment` instead of `base → environment → addendum`. The environment section's cwd is the only per-project variable in the prompt; with it last, the entire `base + addendum` text is a byte-identical, cache-hittable prefix across projects — which matters most in the eval harnesses, where N concurrent trial workspaces test one addendum and previously re-paid full price for the addendum region on every request. Bonus: the grounding instruction now sits at the prompt's end, where recency weighting helps it.

The addendum-composition test pins the new order; the no-addendum form is unchanged (`base → environment` as before).

## Verification

470/470 native tests, 9/9 cram, fmt clean, 0 warnings.

### DIFF
--- a/cmd/openseek/main.mbt
+++ b/cmd/openseek/main.mbt
@@ -460,20 +460,27 @@ async fn configured_system_prompt(
   } else {
     @fs.read_file(base_path).text()
   }
-  let grounded =
-    $|\{base}
-    $|
-    $|\{environment_prompt_section()}
-  if addendum_path == "" {
-    grounded
+  // The environment section goes LAST: DeepSeek's context cache matches on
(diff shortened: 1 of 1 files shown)


bobzhang:

> The prompt assembly change is internally consistent, covered by updated tests, and `moon test cmd/openseek` passes. I did not find any actionable regression introduced by this diff.

PR #104, opened 2026-06-11
feat(deepseek/client): retry transient failures with backoff

bobzhang:
Checklist item: both chat entry points were single-shot, making connect resets / 429 / transient 5xx the most common spurious `turn_failed` class.

## Change

- Up to `retry_attempts` total tries (default 3), exponential backoff from `retry_backoff_ms` (default 500ms, doubling, capped at 60s).
- **Built on `@async.retry`** (after upgrading moonbitlang/async 0.19.1 → 0.19.4): the runtime's `ExponentialDelay` strategy and `max_retry` budget replace the hand-rolled loop; the never-retry taxonomy lives in its `fatal_error` predicate. A thin catch keeps the public failure text identical — the private retryable marker never escapes the package.
- Explicit taxonomy: **retry** 429, 5xx, transport-shaped errors; **never retry** other 4xx, the client's own final verdicts (malformed body, truncated stream), or anything after the stream has produced an event; **always propagate** cancellation immediately.
- **Stream bail covers all output, not just text** (Codex review): the bail flag now flips on every complete SSE event — tool-call deltas and usage included — so a stream that emitted only a tool-call delta before dropping surfaces the failure instead of re-billing the completion (and possibly generating a different call) on retry.
- The retryable marker carries the already-formatted public failure message, so an exhausted budget surfaces exactly the text a non-retrying client would have produced.

## Verification

Six mock flaky-server tests (local `@http.Server`, counted hits): 503×2→200 succeeds in 3 hits; always-500 with budget 2 fails with `DeepSeek API error 500` after exactly 2 hits; 400 fails after exactly 1 hit; pre-stream 429 retries into a clean SSE stream (2 hits); mid-stream truncation after a delivered text delta fails with no second request; tool-call-only delta then drop fails with no second request. 477/477 tests, fmt clean, 0 warnings.

## Codex review

- Round 1 flagged the tool-call-only stream bail gap — fixed in 41f01ac with a regression test.
- Round 2 asked to keep a drop *between a data line and its blank-line terminator* retryable. **Declined** (rationale documented at the signal site in stream.mbt, 2e0f306): the data line itself is model output — a text payload is delivered to the callbacks in that same EOF flush, and a tool-call-only payload is exactly the replay-divergence case round 1 closed; the missing terminator changes neither. A drop with *no* data line at all remains retryable (pinned by the pre-delta-drop test).

### DIFF
--- a/deepseek/client/client.mbt
+++ b/deepseek/client/client.mbt
@@ -15,6 +15,8 @@ pub struct Client {
   api_url : String
   thinking : @deepseek.ThinkingMode?
   reasoning_effort : @deepseek.ReasoningEffort?
+  retry_attempts : Int
+  retry_backoff_ms : Int
 } derive(Debug(ignore=API_KEY))
 
 ///|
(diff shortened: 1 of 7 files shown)


bobzhang:
Codex CLI review trail (`codex review --base main`) — five rounds, four substantive catches:

1. **Decode errors after a 2xx retried** — re-running (and re-billing) the same completion for the same bad payload → parse/decode failures on a successful response are final verdicts.
2. **P1: the fix didn't compile** — `@json.ParseError` is a multi-constructor suberror, so the type-pattern catch failed `moon check` → finalized at the exact parse site inside the stream reader instead.
3. **Deterministic URL parse failures retried** — a malformed `api_url` fails identically before any network I/O → both `URIParseError` constructors propagate immediately.
4. **Pre-delta stream drops never retried** — a 2xx connection dying before any event is transport-shaped; the ended-before-[DONE] failure now raises the retryable class, with the delivered-delta bail still making it final once any output reached the consumer (same public message). New mock test pins the pre-delta drop retrying into a clean second stream.

**Final round:**

> No actionable regressions were found in the retry implementation. The changed code type-checks, and the observed full test failure was from an external real API smoke test/DNS resolution rather than the patch logic.

Six mock flaky-server tests cover the full taxonomy. 476/476 tests, 9/9 cram, 0 warnings. **Awaiting owner signoff to merge.**

PR #112, opened 2026-06-12
feat(cli): --upload collects finished sessions for experiment datasets

bobzhang:
Data-collection groundwork for upcoming experiments: a one-shot run can archive its session to a central place.

## Behavior

- `openseek --upload local "task"` → after the run ends, the session is **copied** (workspace copy stays, so resume keeps working) to `$HOME/.openseek/sessions/<id>/`.
- `--upload <dir>` → `<dir>/sessions/<id>/` — an experiment pools all its trials into one directory, which then works directly as a `--session-root` for the viz server and `--session-list`.
- **Failed runs upload too** — they are the diagnostically interesting ones. `defer` cannot call async functions, so the turn's outcome is held while the upload runs, then re-raised.
- Each copy gains a `meta.json` sidecar with provenance the session format does not store: model, thinking/effort settings, max steps, source workspace, the run's **terminal outcome** (finished/failed/missing, parsed from the event log tail), and upload time.
- **Off by default, strictly best-effort**: nothing leaves the workspace unless asked; a collection failure logs a `session_uploaded ok=false` event and never fails the run. `OPENSEEK_UPLOAD` env-backs the flag so an eval harness can switch on collection for every trial.

Scope note: a server mode (`--upload <url>` to a viz-server ingest endpoint) was prototyped and deliberately cut per owner direction — local collection only for now.

## Verification

End-to-end offline: failed runs against a closed port archived events.jsonl + session.json + meta.json to a pinned `$HOME` and to a custom directory, with `outcome:"failed"` recorded. Cram documents the workflow (new `Collecting Sessions With --upload` section, stamp-normalized). Unit test pins `terminal_outcome` on finished/failed/cut-off/empty logs. 496/496 native tests (the lone failure is the known network-dependent realworld smoke test, 401 in this environment), 14/14 cram, fmt clean.

Also ledgered (pre-existing, found during verification): a raising turn loses its stdout JSONL tail — the drain is cancelled before flushing, so the live stream goes silent exactly when things break; the durable session compensates.

### DIFF
--- a/cmd/openseek/main.mbt
+++ b/cmd/openseek/main.mbt
@@ -74,19 +74,32 @@ async fn run_cli() -> Unit {
           }
         let store = @store.SessionStore(session_root(matches))
         let session = load_or_new_session(store, id, matches, model)
-        ignore(
-          @agent.run_turn_with_append(
-            api_key,
-            model,
-            session,
-            task,
-            append_item=(session, item) => store.append(session, item),
-            max_steps~,
-            thinking~,
-            reasoning_effort~,
-            api_url?,
-          ),
-        )
+        // Collection runs on success AND failure — failed runs are the
+        // diagnostically interesting ones. `defer` cannot call async
+        // functions, so the turn's outcome is held while the upload runs,
+        // then re-raised.
+        let turn_error : Error? = try {
+          ignore(
+            @agent.run_turn_with_append(
+              api_key,
+              model,
+              session,
+              task,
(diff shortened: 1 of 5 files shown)


bobzhang:
**Codex review: sign-off.**
> No actionable correctness issues were identified in the changed code. The new upload path is best-effort, preserves the original turn error, and the targeted CLI/cram coverage passes.

PR #117, opened 2026-06-13
refactor(viz_server): pattern-match percent decoding

bobzhang:
First of a readability refactoring series: no semantic changes, one well-scoped package per PR, idioms per the MoonBit refactoring guide (pattern matching with views, functional loops, char-range matches).

- `percent_decode`: the `mut i` while-loop with a three-clause bounds-and-digits condition becomes a functional `for` matching `chars[i:]` against `['%', high, low, ..]` — the view pattern proves the remaining length, the guard validates the digits, and malformed/truncated escapes fall through to the verbatim arm. State advances via `continue i + 3` / `continue i + 1`.
- `hex_val`: if-chain → char-range `match` (`'0'..='9'` …), with literal overloading dropping `.to_int()` on the range bases.

Both functions are package-private; behavior identical (escape-at-end and bad-digit cases preserved by construction). 495/495 native tests (known realworld network test excepted), fmt clean, zero `.mbti` drift.

### DIFF
--- a/cmd/viz_server/main.mbt
+++ b/cmd/viz_server/main.mbt
@@ -348,20 +348,24 @@ fn slice_from(s : String, start : Int) -> String {
 fn percent_decode(s : String) -> String {
   let chars = s.to_array()
   let bytes : Array[Byte] = []
-  let mut i = 0
-  while i < chars.length() {
-    let c = chars[i]
-    if c == '%' &&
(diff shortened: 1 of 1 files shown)


bobzhang:
**Codex review: sign-off.**
> No actionable regressions were found in the changed percent-decoding logic. The targeted cmd/viz_server tests pass.

PR #118, opened 2026-06-13
refactor(agent_runtime): drain loops via while-is-Some

bobzhang:
Readability refactoring series, part 2 (no semantic changes, one scoped package per PR).

Owner suggestion on `runtime.mbt:88`: rather than `for ;;` + match-on-Option with a mid-loop `return`, bind the next queue item in the loop condition — `while (self.steers.try_get() catch { _ => None }) is Some(text)`. Applied to both `drain_steers` and `drain_events`: each collapses from eight lines to three, the exit condition reads at the top, and two of the `missing_invariant`/`missing_reasoning` (0038/0039) `for ;;` sites disappear as a side effect.

Behavior identical (drain until empty, oldest first; queue failure treated as empty). 495/495 native tests (known realworld network test excepted), fmt clean, zero `.mbti` drift.


bobzhang:
**Codex review: sign-off.**
> The change is a behavior-preserving refactor of the drain loops, and the relevant targeted tests pass. I did not find any actionable correctness issues introduced by this diff.

PR #119, opened 2026-06-13
refactor(agent_tool/read): comprehension + clamped slice in select_lines

bobzhang:
Readability refactoring series, part 3 (no semantic changes, one scoped package per PR).

`select_lines` in the read tool:
- `content.split("\n").map(...).collect()` → `[for line in content.split("\n") => line.to_owned()]`
- the guarded index-push loop (`if start_index < total_lines { for i in ... { selected.push(lines[i]) } }`) → a single slice `lines[start_index:end_exclusive].to_owned()` with both bounds clamped via `.min(total_lines)`; the upstream decode validates `start_line`/`max_lines` positive, so the slice is ascending and in range by construction, and a start past the last line selects nothing exactly as before
- `clamp_end` (negative branch already unreachable) is deleted

Equivalence is pinned by the existing snapshot tests (ranged read with metadata header, capped read, empty file, out-of-range start). 495/495 native tests (known realworld network test excepted), fmt clean, zero `.mbti` drift.

### DIFF
--- a/agent_tool/read/read.mbt
+++ b/agent_tool/read/read.mbt
@@ -173,23 +173,17 @@ fn select_lines(
   start_line : Int,
   max_lines : Int?,
 ) -> ReadSelection {
-  let lines : Array[String] = content
-    .split("\n")
-    .map(line => line.to_owned())
-    .collect()
+  let lines = [ for line in content.split("\n") => line.to_owned() ]
   let total_lines = lines.length()
-  let start_index = start_line - 1
-  let requested_lines = match max_lines {
-    Some(count) => count
-    None => total_lines - start_index
+  // Clamp both bounds into [0, total_lines]: start_line and max_lines are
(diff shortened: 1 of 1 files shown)


bobzhang:
Codex round 1 flagged that a huge `max_lines` (e.g. Int max) with a non-zero start overflowed `start_index + count` negative — the new slice would raise where the old `clamp_end` quietly selected nothing. Fixed by clamping the count to the remaining lines *before* the addition, using core's `Int::clamp(min~, max~)` (owner tip). Note the deliberate divergence from the old quirk: such a request now returns the remaining lines instead of an empty selection; a regression test pins it.

bobzhang:
**Codex review — round 2 (after the clamp fix): sign-off.**
> No actionable correctness issues were found in the diff. The updated line selection clamps bounds safely and the targeted read tests pass.

PR #120, opened 2026-06-13
refactor(deepseek/client): nested JSON patterns in stream deltas

bobzhang:
Readability refactoring series, part 4 (no semantic changes, one scoped package per PR).

`stream.mbt`'s delta handling drops its field-lookup helpers for direct nested patterns:

- `apply_tool_call_delta`: `Object(fields)` extraction + `json_int_field`/`json_string_field` lookups + a `guard fields.get("function") is Some(Object(...))` ladder → `json is { "function": { "name": String(name), .. }, .. }` style two-level patterns. A non-object `"function"` skips both fields exactly as the old early return did; the `index < 0` and non-object-delta early exits are preserved as guards.
- `apply_stream_delta`: map-pattern matches (`delta is { "content": String(content), .. }`) replace helper calls; the `tool_calls` match collapses to an `is`-condition.
- `json_string_field` / `json_int_field` end up unused and are deleted.

Behavior pinned by `client_stream_test` and the fake-server retry suite. 496/497 native tests (the exception is the known network-dependent realworld smoke test, which lives in this very package), fmt clean, zero `.mbti` drift.

### DIFF
--- a/deepseek/client/stream.mbt
+++ b/deepseek/client/stream.mbt
@@ -90,44 +90,25 @@ fn StreamAccumulator::response(
 }
 
 ///|
-fn json_string_field(fields : Map[String, Json], key : String) -> String? {
-  match fields.get(key) {
-    Some(String(value)) => Some(value)
-    _ => None
-  }
-}
-
-///|
-fn json_int_field(fields : Map[String, Json], key : String) -> Int? {
-  match fields.get(key) {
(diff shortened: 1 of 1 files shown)


bobzhang:
**Codex review: sign-off.**
> The change is a localized refactor from helper-based map lookups to MoonBit map/JSON patterns, preserving the previous stream delta behavior. I did not find any introduced correctness issues in the diff.

PR #122, opened 2026-06-13
refactor(agent_session): search_by and any() in projection helpers

bobzhang:
Readability refactoring series, part 6 (no semantic changes, one scoped package per PR).

Deliberately narrow: only the two leaf helpers of the `chat_messages` projection are touched — the stateful projection loop itself (pending tool-call ordering, summary skipping) is semantics-critical and left exactly as is.

- `remove_pending_tool_call`: index scan + mid-loop `remove`/`return` → `pending.search_by(call => call.id == tool_call_id)` then one `remove`, the find-and-act split made explicit.
- `is_covered_by_later_summary`: early-return loop → `events.iter().any(candidate => summary_covers_event(...))`.

Behavior pinned by the existing projection tests (compaction, summary supersession, unresolved historical tool calls, compacted tool results). Full native suite unchanged (known realworld network test excepted), fmt clean, zero `.mbti` drift.

### DIFF
--- a/agent_session/projection.mbt
+++ b/agent_session/projection.mbt
@@ -60,13 +60,13 @@ fn remove_pending_tool_call(
   pending : Array[@deepseek.ToolCall],
   tool_call_id : String,
 ) -> Bool {
-  for index in 0..<pending.length() {
-    if pending[index].id == tool_call_id {
+  match pending.search_by(call => call.id == tool_call_id) {
+    Some(index) => {
       ignore(pending.remove(index))
-      return true
+      true
     }
+    None => false
   }
-  false
 }
 
 ///|
@@ -98,12 +98,9 @@ fn is_covered_by_later_summary(
   event : SessionEvent,
   events : @vector.Vector[SessionEvent],
 ) -> Bool {
-  for candidate in events {
-    if summary_covers_event(candidate, event.sequence()) {
-      return true
-    }
-  }
-  false
+  events
+  .iter()
+  .any(candidate => summary_covers_event(candidate, event.sequence()))
 }
 
 ///|


bobzhang:
**Codex review: sign-off.**
> The changes are behavior-preserving refactors of manual loops to standard search/iterator helpers. I did not identify any actionable correctness issues in the diff.

agent_session/projection.mbt
```diff
@@ -98,12 +98,9 @@ fn is_covered_by_later_summary(
...
   events : @vector.Vector[SessionEvent],
 ) -> Bool {
-  for candidate in events {
-    if summary_covers_event(candidate, event.sequence()) {
-      return true
-    }
-  }
-  false
+  events
+  .iter()
```
bobzhang, line 102: events.any() directly?

agent_session/projection.mbt
```diff
@@ -60,13 +60,13 @@ fn remove_pending_tool_call(
   pending : Array[@deepseek.ToolCall],
   tool_call_id : String,
 ) -> Bool {
-  for index in 0..<pending.length() {
-    if pending[index].id == tool_call_id {
+  match pending.search_by(call => call.id == tool_call_id) {
```
bobzhang, line 63: check if we have `remove_if`

bobzhang:
Both inline suggestions checked against the API:

- **`events.any()` directly** (projection.mbt:102): `events` is a `@vector.Vector`, and `moon ide doc "@vector.Vector"` shows it exposes only `iter()`, `each`, `fold`, and `contains` — no direct `any`/`all`. So `events.iter().any(...)` is the minimal idiomatic form here. (Array does have `any` directly, but this collection isn't an Array.)
- **`remove_if`** (projection.mbt:63): there's no `remove_if` in core (`moon ide doc "*remove_if*"` is empty). The nearest is `Array::retain`, but it returns `Unit` (can't signal found-vs-not, which this function returns as `Bool`) and removes *all* matches rather than the first. So `search_by` + `remove` remains the right fit.

Leaving #122 as-is unless you'd prefer I convert `events` to an Array upstream (would change the field type, out of scope for a no-semantics refactor).

PR #133, opened 2026-06-13
[codex] Reject unknown openseek CLI options

### LINKED ISSUE #132: Reject unknown CLI options before treating them as task text
## Problem

`cmd/openseek` currently accepts unknown hyphen-leading tokens as part of the free-form `task` positional. This can hide stale or mistyped options.

Example seen while testing PR #130:

```bash
moon run cmd/openseek -- --dir ./harness/hi2 'how are you'
```

On a checkout that does not yet have `--dir`, this does not fail. Instead, `--dir ./harness/hi2 how are you` is parsed as the task text, and the run records under the default cwd `.openseek/sessions`.

## Expected

If a token looks like an option and is not recognized by the CLI, `openseek` should report an unknown option error, at least before the first explicit task delimiter or task capture point.

For #130 specifically, a stale checkout should have made it obvious that `--dir` was unsupported instead of silently turning it into prompt text.

## Actual

Because the `task` positional is variadic and has `allow_hyphen_values=true`, unknown options are swallowed into the prompt.

## Notes

We still need a way to pass prompts that contain flag-like text, for example asking the model to fix `--help` output. Possible approaches:

- Require `--` or another explicit delimiter before arbitrary hyphen-leading prompt text.
- Keep permissive parsing only after the first non-option task token.
- Add a pre-parse/preflight check that rejects known option-looking tokens before task capture, while preserving quoted/free-form task ergonomics where practical.

This is a follow-up cleanup; PR #130 itself parses `--dir` correctly when run from the correct branch.

bobzhang:
## Summary

Fixes #132.

- Remove `allow_hyphen_values=true` from the variadic `task` positional so unknown option-looking tokens are reported by argparse.
- Keep the standard `--` delimiter path for prompts that intentionally start with `-`.
- Avoid printing `error: error: ...` when an argparse error already includes the `error:` prefix.
- Add unit and cram coverage for unknown-option rejection and delimiter behavior.

## Why

Before this change, a stale or mistyped flag such as:

```bash
openseek --xxy he
```

was silently parsed as task text. That made a stale checkout of PR #130 confusing: `--dir ./harness/hi2 'how are you'` could run as a prompt and record under the default session root instead of failing fast.

This is the same class of ambiguity seen in other CLIs when option values or arguments start with `-`. For example, on this machine:

```bash
node -e '-3'
# node: -e requires an argument

node --eval=-3
# exits 0
```

Shell quotes do not survive into argv, so the parser needs a clear rule. OpenSeek now follows the usual CLI behavior: unknown leading options fail, and users can pass a hyphen-leading prompt with the explicit delimiter:

```bash
openseek -- '--xxy he'
```

## Validation

- `moon test --target native cmd/openseek`
- `moon cram test tests/cram`
- `DEEPSEEK= moon test --target native`
- `moon info && moon fmt`
- `git diff --check`

### DIFF
--- a/cmd/openseek/main.mbt
+++ b/cmd/openseek/main.mbt
@@ -172,6 +172,10 @@ async fn random_salt() -> String {
 
 ///|
 fn failure_message(error_text : String) -> String {
+  match error_text.strip_prefix("error: ") {
+    Some(message) => return message.to_owned()
+    None => ()
+  }
   let marker = " FAILED: "
   let pieces : Array[String] = error_text
     .split(marker)
@@ -186,6 +190,16 @@ fn failure_message(error_text : String) -> String {
   } else {
     message
   }
+}
+
+///|
+test "failure_message avoids duplicating argparse error prefix" {
+  assert_eq(
+    failure_message(
+      "error: unexpected argument '--xxy' found\n\nUsage: openseek",
+    ),
+    "unexpected argument '--xxy' found\n\nUsage: openseek",
+  )
 }
 
 ///|
@@ -316,7 +330,6 @@ fn cli_command() -> @argparse.Command {
       PositionArg(
         "task",
         num_args=ValueRange(lower=0),
-        allow_hyphen_values=true,
         about="Task description.",
       ),
     ],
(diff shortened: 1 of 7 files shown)


cmd/openseek/main.mbt
```diff
@@ -172,6 +172,9 @@ async fn random_salt() -> String {
 
 ///|
 fn failure_message(error_text : String) -> String {
+  if error_text.has_prefix("error: ") {
```
bobzhang: strip_prefix

cmd/tui/main.mbt
```diff
@@ -78,7 +78,6 @@ fn cli_command() -> @argparse.Command {
       PositionArg(
         "task",
         num_args=ValueRange(lower=0),
-        allow_hyphen_values=true,
```
bobzhang, line 81: cc @tonyfettes  allowing this would be too confusing

PR #159, opened 2026-06-15
fix(openseek): fall back to USERPROFILE for skills

tonyfettes:
This PR fixes user-level skills directory path discovery on Windows by using `USERPROFILE` on Windows.

## Relationship

This PR is independent and targets `main`. #160 (`--format=json`) and #166 (`md_to_mbt_string` basename fix) also target `main` independently. #167 is stacked only on #166.

### DIFF
--- a/cmd/openseek/main.mbt
+++ b/cmd/openseek/main.mbt
@@ -269,7 +269,7 @@ fn cli_command() -> @argparse.Command {
         long="global-skills-dir",
         env="OPENSEEK_GLOBAL_SKILLS_DIR",
         default_values=[""],
-        about="User-level skills directory advertised alongside workspace skills; empty means $HOME/.openseek/skills.",
+        about="User-level skills directory advertised alongside workspace skills; empty means $HOME/.openseek/skills, or %USERPROFILE%/.openseek/skills on Windows.",
       ),
       OptionArg(
         "session",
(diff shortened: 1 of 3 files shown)


cmd/openseek/main.mbt
```diff
@@ -705,9 +705,21 @@ async fn configured_system_prompt(
...
 
+///|
+fn default_global_skills_dir(home : String?, userprofile : String?) -> String {
+  if home is Some(home) && !home.trim().is_empty() {
+    return home.trim().to_owned() + "/.openseek/skills"
+  }
+  if userprofile is Some(home) && !home.trim().is_empty() {
+    return home.trim().to_owned() + "/.openseek/skills"
+  }
+  ""
```
tonyfettes: Honestly I don't feel it is quite right to fallback to an empty string (`""`) here, but the original `skills_prompt_section_for_cwd` function uses `""` anyway when HOME is not set, so this fix uses `""` to keep the behavior compatible.

cmd/openseek/main.mbt
```diff
@@ -705,9 +705,63 @@ async fn configured_system_prompt(
...
+    global_skills_dir_for_home(userprofile)
+  }
+  if fallback is Some(dir) {
+    return dir
+  }
+  ""
+}
+
+///|
+fn default_global_skills_dir(home? : String, userprofile? : String) -> String {
```
bobzhang: 1. return option
2. no duplication

PR #170, opened 2026-06-17
Migrate read to single-file calls

bobzhang:
## Summary
- Rebases cleanly onto current `main`.
- Keep `read` single-file-only: the schema exposes required `path`, `ReadInput` carries one `path`, and `paths` is rejected with guidance to batch separate read calls in one assistant response.
- Add an explicit directory-path error that points agents to inspect directories with `ls` or `tree` before reading specific files.
- Keep headered read output line-numbered with right-aligned `number| text` gutters, and count those gutters toward `max_output_chars`.
- Preserve Unicode-safe truncation on character boundaries and update docs/tests for the output shape.

## Validation
Latest head (`b6948db`):
- `moon info`
- `moon fmt`
- `git diff --check`
- `moon test agent_tool/read --diagnostic-limit 100` (`21/21`)
- `moon test --diagnostic-limit 100` (`16/16` JS, `612/612` native)

## Fresh Review
- `codex exec review --base upstream/main --dangerously-bypass-approvals-and-sandbox -o /tmp/codex-review-pr170-rebased.md` completed with no actionable findings.
- The review also ran the modified package tests and full `moon test`; it noted the read package is native-only when probing all targets.

## Follow-up Ideas From Kimi Code Read
- Tail reads: support reading from EOF, either with negative `start_line` or a separate `tail_lines`, while keeping absolute line numbers in output.
- Streaming read path: iterate lines and collect only the requested window or tail instead of loading the whole file, while still surfacing total line count when possible.
- Layered caps: add max-line and per-line-length caps next to `max_output_chars`, and report which limit/truncation was hit.
- Text/media guardrails: reject binary, non-UTF-8, NUL-containing, image, and video inputs with direct guidance to media or shell tools.
- Secret guardrails: refuse `.env`, credential stores, and SSH private keys, with explicit exemptions for example/public-key files.
- Status footer: include total line count and truncation reason in a machine-obvious footer so the model can page correctly.

### DIFF
--- a/agent_tool/read/README.mbt.md
+++ b/agent_tool/read/README.mbt.md
@@ -14,9 +14,11 @@ large generated files and dependency sources can flood the transcript and push
 useful context out of the model window.
 
 Ranged or capped reads include a metadata header so the model knows what it saw
-and what it did not see. Automatic character truncation is marked as a tool
-error even when the file was read successfully; that makes lossy context visible
-to the loop instead of silently pretending the returned prefix is complete.
+and what it did not see. The body of those headered blocks is line-numbered
+(`number<TAB>text`) to make follow-up edits and focused reads less error-prone.
+Automatic character truncation is marked as a tool error even when the file was
+read successfully; that makes lossy context visible to the loop instead of
+silently pretending the returned prefix is complete.
 
 Multi-file reads exist because agent traces show models reading related files
 one round trip at a time — each a full model turn. `paths` collapses those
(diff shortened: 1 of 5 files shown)


bobzhang:
Updated PR head to `eef48d3` for the aligned line-number gutter follow-up.

Changes:
- Headered read output now uses a right-aligned line-number gutter: `  9| text`, ` 99| text`, `100| text`.
- `line_format` metadata and README/docs now describe `right-aligned-number| text`.
- Added a focused test covering alignment across one-, two-, and three-digit line numbers.

Validation run locally:
- `moon test agent_tool/read --diagnostic-limit 50`
- `moon check agent_tool --diagnostic-limit 50`
- `moon test` (`11/11` JS, `582/582` native)
- `moon info && moon fmt`
- `git diff --check`

PR #174, opened 2026-06-18
[codex] Validate MoonBit manifest edits with moon_config

bobzhang:
## Summary

- Add `moonbitlang/moon_config` and use it in the edit manifest guard.
- Validate proposed `moon.mod`, `moon.pkg`, and `moon.work` writes with the parser before accepting them.
- Keep the existing clearer guardrails for legacy `moon.mod.json`, JSON-shaped manifests, empty `moon.mod`, and suspiciously tiny `moon.pkg`.
- Extend manifest guard tests and README coverage for the parser-backed behavior.

## Why

The edit tool previously used narrow string heuristics for MoonBit manifest writes. That caught some common mistakes, but malformed current-syntax manifests could still be written and only fail later during build or test commands. `moonbitlang/moon_config` provides the parser-backed validation needed to reject those writes at the edit boundary.

## Validation

- `moon check`
- `moon info`
- `moon fmt`
- `moon test` -> 585 passed, 0 failed

### DIFF
--- a/agent_tool/edit/internal/manifest_error/README.mbt.md
+++ b/agent_tool/edit/internal/manifest_error/README.mbt.md
@@ -6,6 +6,9 @@ safety checks used before the `edit` tool writes replacement content.
 This package is internal to `agent_tool/edit`. It does not edit files itself;
 it inspects the target path and proposed file content, then returns an optional
 error message for manifest rewrites that look like common agent mistakes.
+`moon.mod`, `moon.pkg`, and `moon.work` writes are also parsed with
+`moonbitlang/moon_config` so malformed current-syntax manifests are rejected at
+the edit boundary.
 
 ## API Shape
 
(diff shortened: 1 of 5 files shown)


bobzhang:
Reviewed + codex-reviewed. The change is correct in intent and rebases cleanly onto `main` (builds + tests green, 28/28), but I'm **holding it**: it surfaces a **P1** caused by an upstream bug.

`moon_config@0.3.4` flags standard `moon.mod` metadata keys (`readme`, `repository`, `license`, `keywords`, `description`, `preferred_target`, `warnings`) as `unexpected key` — including this repo's own root `moon.mod`. Since this PR turns *any* parser report into a write error, merging it would make the `edit` tool **refuse valid `moon.mod` edits**. (`moon.pkg`/`moon.work` post-processing are complete; `moon.mod` is the broken one.)

Filed upstream: moonbitlang/parser#142.

**Plan:** hold until #142 is fixed and `moon_config` is bumped, then this works as designed with no regression. Before merge, also: retarget base from `viz-fold-tool-io` (squash-merged) to `main`, and rebase (drops the 4 already-merged commits, leaving just the manifest commit).

PR #178, opened 2026-06-20
feat(viz): flag failed tool calls + error-path briefs (read/write/edit)

bobzhang:
## What

The folded tool-call line only showed a brief on **success**, so a failed `read`/`write`/`edit` rendered as a plain `→ read` with no indication it failed. This makes failures visible at a glance in the assistant's action list.

- **Viz** — appends a 🚩 to a call's brief whenever its result errored (`call_brief_display`), in both raw and model views. Even a failed call with no brief renders `→ read 🚩`.
- **read / write / edit** — now emit a brief on their **error** paths too (the target's basename), so a failure shows what it touched: `→ read · read x.txt 🚩`, `→ write · write foo.mbt 🚩`, `→ edit · edit bar.mbt 🚩`. (shell/moon_cmd already brief failures via their exit code, so they pick up the flag automatically.)

(This re-lands a commit that was stranded when #177 merged — `call_brief_display` never reached `main` — plus the new `edit` error briefs.)

## Verification

- `moon check --deny-warn` clean; `moon fmt` applied.
- Tests: viz **20/20** (failed-call flag, with/without brief, both views), edit **18/18** (added a failed-edit-briefs-the-file test), read/write pass.

### DIFF
--- a/viz/view.mbt
+++ b/viz/view.mbt
@@ -304,8 +304,7 @@ fn tool_call_briefs(session : @agent_session.Session) -> Map[String, String] {
   let briefs : Map[String, String] = Map([])
   for event in session.events() {
     if event.item() is Tool(result) {
-      let display = call_brief_display(result.brief(), result.is_error())
-      if display != "" {
+      if call_brief_display(result.brief(), result.is_error()) is Some(display) {
         briefs.set(result.tool_call_id(), display)
       }
     }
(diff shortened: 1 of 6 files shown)


viz/view.mbt
```diff
@@ -303,13 +303,34 @@ fn tool_call_view(
...
   }
   briefs
 }
 
+///|
+/// The folded-call brief text: the tool's one-line brief (if any) plus a red
+/// flag when its result errored, so a failed call (e.g. a missing-file `read`)
+/// stands out in the assistant's action list without expanding it. Returns `""`
+/// when there is nothing to show (a successful call with no brief).
+fn call_brief_display(brief : String?, is_error : Bool) -> String {
```
bobzhang: should still return `String?`

PR #179, opened 2026-06-20
feat(write): create missing parent directories

bobzhang:
## What

`write dir/moon.pkg` previously failed when `dir` didn't exist (`@fs.write_file` errors on a missing parent), so the agent had to `shell mkdir -p` first or waste a failed write — friction when scaffolding a new package/module. Now `write` creates the parent directory recursively before writing.

- Best-effort `mkdir -p` of the parent; the path is already resolved under the workspace root, so it stays in-workspace.
- A parent component that's a regular file still surfaces a normal `error writing …`.
- Doc updated; the prior "missing parent → fails" test now asserts the file-as-parent error instead.

## Verification

- `moon check --deny-warn` clean; `moon fmt` applied; **16/16** write tests (added: creates nested parents; `parent_dir` unit test).

### DIFF
--- a/agent_tool/write/write.mbt
+++ b/agent_tool/write/write.mbt
@@ -1,7 +1,7 @@
 ///|
 /// Tool definition for `write`: overwrites `arguments.path` with
 /// `arguments.content`. Existing files are truncated; missing parent
-/// directories are not created.
+/// directories are created (so `dir/moon.pkg` works without a separate mkdir).
 ///
 /// ## Arguments
 /// - `path` (string, required): filesystem path to write.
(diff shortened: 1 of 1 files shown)


agent_tool/write/write.mbt
```diff
@@ -85,6 +86,37 @@ async fn write_file(
...
 
+///|
+/// Create the parent directory of `path` (recursively) when it is missing, so a
+/// write to `dir/moon.pkg` works without a separate `mkdir`. The path is already
+/// resolved under the workspace root, so this stays inside the workspace.
+/// Best-effort: a genuine failure surfaces from the write itself.
+async fn create_parent_dirs(path : String) -> Unit {
+  let parent = parent_dir(path)
+  if parent != "" && !@fs.exists(parent) {
+    @fs.mkdir(parent, recursive=true) catch {
```
bobzhang: we should not catch it? let it propgate since swallow such error is not helpful, the follow up will fail any way

PR #180, opened 2026-06-20
feat(agent): disable the moon_check tool; run moon check through shell

bobzhang:
## What

The `moon_check` watcher tool (a persistent `moon check --watch` per project) added complexity and **blocked plain `moon check` in shell**, forcing the model through a stateful tool. This disables it.

- **Unregister** `moon_check` from the agent's tool registry, so it's no longer offered to the model. The package itself stays — the agent loop's watcher-event helpers, the eval harness, and the TUI still reference it, so a full deletion would sprawl across packages.
- **Unblock** `moon check` in the shell command policy (bare, behind `cd`, or compound). The only remaining redirect is the bare `moon_cmd` tool name typed as a command.
- **Rewrite the system prompt** (base + flash, ~10 references) to use shell `moon check` for compiler feedback instead of the moon_check tool/watcher.

## Verification

- `moon check --deny-warn` clean (native + js); `moon fmt`; generated prompt files regenerated.
- Updated the two tests that asserted moon_check registration (`tool_definition`, `steer_test`) and the command-policy tests (moon check now allowed; `moon_cmd` still redirected).


bobzhang:
A/B test of this simplification: **no regression** (5/5 ↔ 5/5 success, ~33s ↔ ~35s wall, equal tool volume) and the agent adopts the intended active shell `moon check` loop (2.4×/run vs the watcher tool's 1×). Detailed report in #183.

bobzhang:
## Complex A/B follow-up: write a TOML parser

The first A/B used a trivial task (both arms ~100% / ~30s, so it only showed *behavior*, not outcomes). This re-runs the comparison on a **complex, multi-step task**: implement a TOML parser (lexing, `key = value`, string/int/float/bool, `#` comments, `[table]` headers, inline arrays) so that a 6-case `moon test` suite passes. Same isolated worktrees (A = `main` with `moon_check`; B = this PR), `deepseek-v4-pro`, `--max-steps 90`, **N=3**. Objective success = `moon test` green.

| Config | success | wall (s, mean / median) | turns | shell `moon check` | `moon_check` tool |
|--------|:-------:|:----------------------:|:-----:|:------------------:|:-----------------:|
| **A** (with moon_check) | **3/3** | 271 / 267 | 24.3 | 0.0 | 2.0 |
| **B** (this PR)         | **3/3** | **162 / 146** | 38.3 | **4.0** | 0.0 |

<details><summary>Per-run CSV</summary>

```
config,trial,wall_s,moon_test_pass,turns,tools,moon_check_tool,shell_moon_check,shell_build_test
A,1,356,1,27,27,4,0,2
A,2,267,1,24,27,1,0,3
A,3,190,1,22,23,1,0,3
B,1,146,1,30,32,0,3,3
B,2,101,1,24,26,0,4,4
B,3,239,1,61,63,0,5,4
```
</details>

### Findings

1. **No success regression** — both arms solved the parser **3/3**.
2. **B is ~40% faster in wall-clock** (162s vs 271s mean; 146s vs 267s median), *despite using more turns* (38 vs 24). The one-shot shell `moon check` loop iterates faster than A's persistent `moon check --watch` watcher tool, which adds start/await latency per check. The harder the task, the more that per-iteration cost compounds — invisible on the trivial task (where wall-clock was a tie), clear here.
3. **Intended behavior confirmed at scale** — every B run drove the loop with shell `moon check` (4×/run avg); A leaned on the watcher tool (2×/run) and never shell-checked.

### Caveats
N=3, single task, one model; B-3 was a 61-turn outlier (239s) but B still wins on mean and median. Both 100% success, so this measures efficiency, not a success-rate delta.

### Conclusion
On a complex task the simplification is **net positive, not just neutral**: same success, **meaningfully faster**, simpler agent surface, and the agent adopts the faster check loop on its own. Reinforces merging #180.

PR #193, opened 2026-06-20
Closed: skip viz prompt backfill

bobzhang:
Closed per simplified scope.

We do not need to backfill or reconstruct project layout for older session headers. Newly created sessions already persist the project layout through `configured_system_prompt(...)` -> `environment_prompt_section(...)`, and existing tests cover that path.

The branch now contains a revert commit, so the PR has zero effective file changes.


bobzhang:
Closing this out per the simplified scope: we do not need to backfill or reconstruct layout for older session headers. New sessions already persist the project layout through `configured_system_prompt(...)` -> `environment_prompt_section(...)`, and the existing tests cover that path.

PR #205, opened 2026-06-20
dir_outline returns String?; improve environment prompt; add tests

bobzhang:
### `testkit/filesystem` — `dir_outline` returns `String?`
- Changed `dir_outline` return type from `String` to `String?`; empty/unreadable directories now return `None` instead of `""`.
- Converted the existing outline fixture from `FileSystem::from_pairs` to the JSON-style `FileSystem({})` constructor.
- Added four new tests covering `max_depth`, `max_dirs` truncation, nested skip rules (hidden/build dirs), and a full inline snapshot.

### `cmd/openseek` — environment prompt improvements
- Environment prompt now explicitly says `(empty directory)` instead of silently omitting the layout section when the workspace has no directories.
- Refactored the layout suffix into a local `fn layout(sb)` callback so the entire prompt is built with `$|` string-interpolation rather than separate string concatenation.
- Added an inline snapshot test that asserts the full rendered environment block for a deterministic temp-dir fixture (hidden/build dirs filtered, two visible dirs shown).

### `.vscode/settings.json`
- Disable MoonBit extension inlay hints at workspace level (`forBinderTypes`, `forAnonymousFunctionReturnTypes`, `forFunctionParameterNames`).

### DIFF
--- a/cmd/openseek/main.mbt
+++ b/cmd/openseek/main.mbt
@@ -908,99 +908,17 @@ async test "environment section names the cwd and appends the layout outline" {
   let section = environment_prompt_section()
   let cwd = @fs.realpath(".")
   // Normalise the machine-specific CWD so the snapshot is portable across
-  // local dev and CI, while still asserting the full rendered structure.
+  // local dev and CI, while still asserting a stable contiguous layout block.
   let portable = section.replace_all(old=cwd, new="{cwd}")
-  inspect(
-    portable,
-    content=(
-      #|## Environment
-      #|
-      #|Working directory: {cwd}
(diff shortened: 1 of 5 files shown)


cmd/openseek/main.mbt
```diff
@@ -881,29 +881,95 @@ async fn environment_prompt_section(workspace_root? : String = ".") -> String {
       _ => workspace_root
     }
   }
-  let section =
+  // FIXME(upstream)
+  // " \{sb => f(sb)}"
+  // `f` is not allowed to be async, either fix it or better error message (macro expansion should work)
```
bobzhang, line 886: we still expand `\{sb => f(sb)}` like macros

PR #220, opened 2026-06-22
[codex] Stream responses through item writer

tonyfettes:
## Summary

- Route assistant and reasoning streaming through a live transcript `ItemWriter` instead of the activity line.
- Add a step-level stream phase so reasoning and response writers cannot overlap, and so response-not-started is distinct from response-closed.
- Keep completed hard lines in scrollback while the current line redraws live, with focused tests for writer behavior and state transitions.

## Why

The previous activity-line streaming model could not preserve completed rows in scrollback or cleanly handle reasoning/response transitions. During the first pass, late `reasoning_message` events could also try to open a second live writer over an active response stream. The explicit stream phase makes those transitions representable and prevents nested writers.

## Validation

- `moon info`
- `moon fmt`
- `moon check`
- `moon test`
- tmux smoke: late reasoning after response stream does not crash or insert a thought item
- tmux smoke: reasoning deltas before response render as `✻` above the `⏺` response

### DIFF
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -332,11 +332,18 @@ pub async fn[X] run_turn_in_scope(
       let response = client.chat(
         messages,
         tools=tools.function_tools(),
-        stream=StreamHandler(on_content_delta=delta => {
-          if delta != "" {
-            @xlog.info() <? { "event": "assistant_delta", "content": delta }
-          }
-        }),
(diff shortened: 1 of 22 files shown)


tonyfettes:
streaming has gone through a series on-and-off PRs

- #68
- #214
- #220 this PR
- #498
- #820
- #825
- #978

PR #222, opened 2026-06-22
[codex] add shell parser policy analysis

bobzhang:
## Summary

- Add an internal conservative shell parser for static policy analysis of simple POSIX shell commands.
- Use that parser in shell command policy so accidental `moon_cmd` invocations are detected in flat compounds such as `cd demo && moon_cmd check`.
- Preserve the narrow existing policy: ordinary `moon ...` commands still run through shell, and complex commands are only blocked by the legacy leading `moon_cmd` fallback.

## Validation

- `moon fmt agent_tool/shell/internal/shell_parse`
- `moon check agent_tool/shell/internal/shell_parse`
- `moon test agent_tool/shell/internal/shell_parse`
- `moon info agent_tool/shell/internal/shell_parse`
- `moon fmt agent_tool/shell/internal/command_policy`
- `moon check agent_tool/shell/internal/command_policy`
- `moon test agent_tool/shell/internal/command_policy`
- `moon info agent_tool/shell/internal/command_policy`
- `moon check --target all`
- `moon test agent_tool/shell`
- `moon test`
- `git diff --check`

## Independent Reviews

- Fresh Codex CLI reviewed `05cf9f4`; findings fixed in `5b42d69`.
- Fresh Codex CLI reviewed `5b42d69`; findings fixed in `33f1105`.
- Fresh Codex CLI reviewed `33f1105`; findings fixed in `d60a02c`.
- Fresh Codex CLI reviewed `d60a02c`; no actionable issues.
- Fresh Codex CLI reviewed `2aa2f8d`; no actionable issues.

### DIFF
--- a/agent_tool/shell/README.mbt.md
+++ b/agent_tool/shell/README.mbt.md
@@ -44,6 +44,18 @@ Prefer dedicated tools when they encode useful policy. Run Moon commands such as
 `shell` with the `cwd` field. Use shell features such as pipes and heredocs for
 CLI probes.
 
+After a successful mutating Moon command, `shell` appends bounded
+`moon check --diagnostic-limit 1` feedback from the nearest MoonBit module root,
+respecting `moon -C <dir>` and explicit cwd changes such as
+`cd ./dir && moon ...`. Bare relative `cd dir` is skipped because `CDPATH` can
+change which directory the shell enters. This covers commands such as
+`moon add`, `moon remove`, `moon update`, `moon fmt`, `moon info`,
+`moon test --update`, and `moon ide rename ... --apply`. Read-only variants
+such as dry-run commands and `moon fmt --check` are left alone, as are Moon
+invocations with per-command environment overrides such as `FOO=bar moon` or
+`env FOO=bar moon`. Follow-up checks are skipped when `timeout_ms` is set and
+share the remaining `max_output_chars` budget.
+
 ## Arguments
 
 | Name | Type   | Required | Notes |
@@ -62,6 +74,8 @@ arguments, non-zero shell exit codes, and output truncation. The string body
 has one of these shapes:
 
 - `"exit=<code>\n<stdout/stderr merged>"` — normal completion.
+- For successful mutating `moon` commands, the normal completion output may be
+  followed by a `moon check:` section with bounded compiler diagnostics.
 - `"exit=<code-or-cancelled>\ntruncated=true\noutput_limit_reached=true\nshown_chars=<n>\nmax_output_chars=<n>\n<output-prefix>"` —
   output exceeded `max_output_chars` while the process pipe was being read. The
   command is cancelled if it has not already exited; no full output length is
--- a/agent_tool/shell/internal/command_policy/README.mbt.md
+++ b/agent_tool/shell/internal/command_policy/README.mbt.md
@@ -34,14 +34,17 @@ The policy allows:
 
 ## Detection Strategy
 
-The implementation is deliberately small and conservative:
+The implementation is deliberately conservative:
 
-- Split the shell command into rough words.
-- Skip simple leading environment assignments such as `DEEPSEEK_MODEL=x`.
-- Reject a leading `moon_cmd`.
+- Parse the command with the internal shell policy parser.
+- Reject `moon_cmd` when it is a statically visible simple command name,
+  including in flat compounds such as `cd demo && moon_cmd check`.
+- If parsing is too complex, keep a small fallback for the old leading
+  `moon_cmd` typo case.
 
-This is a heuristic parser, not a full POSIX shell parser. Known hardening work
-is tracked in `agent_tool/shell/TODO.md`.
+This is still not a security sandbox and not a full POSIX shell interpreter.
+Commands whose runtime argv cannot be known statically are allowed unless they
+match the narrow fallback typo check.
 
 ## Examples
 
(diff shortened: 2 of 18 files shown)


bobzhang:
## Review of #222 — shell parser policy analysis

The PR adds a conservative, statically-analyzing shell parser (`shell_parse`) and rewires `command_policy` to use it so `moon_cmd` is caught in flat compounds (`cd demo && moon_cmd check`), not just as a leading word. It also lands a large, separately-scoped feature in `moon_auto_check.mbt`: after a successful *mutating* `moon` command, it computes the effective cwd and appends a follow-up `moon check --diagnostic-limit 1`.

I traced the safety-critical paths and they hold up: the cwd state machine correctly bails on pipes, `;`-after-`cd`, env mutation, bare `cd dir` (CDPATH), and dry-run/read-only commands; `command_policy`'s `Simple`-only / fallback split is sound and the `parse_for_policy → Simple → contains_command` path strictly improves detection without new false positives; the `moonbitlang/x/path` test→main dep move is justified; no dependency cycle. I did **not** find a wrong-directory or false-block bug. The findings below are mostly **design/cleanup** — the code is correct but carries avoidable duplication and re-work.

### 1. `moon_auto_check.mbt` reimplements `shell_parse`'s wrapper grammar instead of consuming it (altitude + reuse)
`shell_parse` already resolves the effective command name through `env`/`command`/`builtin` wrappers (`SimpleCommand::command_name()`, `effective_command_name`, `command_wrapped_command_name`, `env_wrapped_command_name` in `query.mbt:46-94`) and already splits leading assignments into `command.env`. Yet `moon_auto_check.mbt` re-walks `argv[0] == "env"/"command"/"builtin"` from scratch in `command_payload_start` (`:235`), `command_wrapper_payload_start` (`:256`), `moon_command_start` (`:425`) and `env_moon_command_start` (`:443`) — `command_name()` is never called by this file.

The two copies have already **drifted**: `query.mbt`'s `env_wrapped_command_name` returns `Some("env")` on an unknown dash-option, while `env_moon_command_start` returns `None` for the same input. The root cause looks like an *altitude* gap: `SimpleCommand` exposes only `command_name() : String?`, but consumers need the command-word **index** (to keep scanning `-C`, `--dry-run`, the subcommand, and its flags). Consider exposing that index (or an env-stripped argv tail) on the struct so `moon_command_start`/`cwd_command_effect` can drop ~120 lines of parallel wrapper-walking. As written, any future fix to one wrapper grammar (e.g. `command -p --`) must be mirrored by hand in the other.

### 2. `is_env_assignment_word` / `is_valid_env_name` are defined twice, with different shapes
`parser.mbt:220` / `:225` (String-based, via `word.split("=")`) and `moon_auto_check.mbt:468` / `:476` (StringView-based, via `word.find("=")` + slice). Two implementations answer the identical "is this `NAME=value`?" question for the same command string. An env-name rule change must be made in both, in two code styles; missing one makes the policy parser and the auto-check disagree. Either export the helper from `shell_parse` or rely on the already-split `command.env`.

### 3. `parse_for_policy` runs twice (sometimes three times) per shell invocation (efficiency)
On every shell call: once in `moonbit_command_policy_error` (`command_policy.mbt:10`, reached from `shell.mbt:74`), then again in `append_moon_command_check_summary → moon_auto_check_start_cwd` (`moon_auto_check.mbt:108`). On the non-`Simple` policy branch there's a third re-lex via `rough_first_command_name → shell_words` (`command_policy.mbt:99`). The `ParseResult` is pure and could be computed once and threaded into the auto-check.

### 4. `has_cd` / `has_cd_and_command` are exported but have no non-test consumers
`query.mbt:25` / `:41` are `pub` (and in `pkg.generated.mbti`) but `git grep` finds zero callers outside `shell_parse_test.mbt` — meanwhile `moon_auto_check.mbt` rolls its own cd detection (`cwd_command_effect`). Either consume these or drop them so the public surface reflects real usage.

### 5. Lexer per-word reset is copy-pasted ~7× (simplification)
The 7-field reset (`word = StringBuilder(); word_empty = true; … word_has_assignment_tilde_expansion = false`) plus the preceding `push_word(...)` 6-arg call is repeated in the space branch and each of the `&`, `|`, `;`, `>`, `<`, newline operator branches in `lexer.mbt`. A `flush_and_reset` helper (or grouping the per-word state into one small mutable struct) removes ~50 lines and prevents a missed-reset bug where a stale flag bleeds into the next word's classification.

### 6. Minor: a couple of valid redirection forms are over-rejected
- `cmd &>>out` (append-both) lexes as `Op("&>")` + `Op(">")` and is reported as `SyntaxError("redirection &> requires a target")` (`lexer.mbt` `&` branch around `:110`).
- `cmd <&3` (fd-dup) returns `TooComplex("background operator is not supported")` (`lexer.mbt:110`) — both over-restrictive and a misleading reason (it isn't a background operator).

No security/policy impact (both resolve conservatively, so the auto-check is just skipped), but they misclassify legitimate commands. Note also the deliberately-conservative `env -u FOO moon_cmd check` (tested) means that particular `moon_cmd` typo form is *not* redirected — fine if intended, worth a comment if not.

### 7. Minor: follow-up-check budget ignores the response header
`moon_check_output_budget` (`moon_auto_check.mbt:46`) computes `remaining = max_output_chars - output.text.char_length()`, but the summary is appended to `"exit=<code>\n" + output.text` plus a `\n` joiner. When `output.text` is near `max_output_chars`, the final tool content can exceed the documented budget by ~`len("exit=<code>\n")+1` (~8 chars). The README says follow-up checks "share the remaining `max_output_chars` budget", so this is a small contract drift.

---
*Reviewed at medium effort (precision-focused). Items 1–4 are the ones I'd act on; 5–7 are minor.*

PR #228, opened 2026-06-23
read: one uniform output shape (always-numbered + status footer), budget in UTF-16 code units

bobzhang:
## Summary

Reworks the `read` tool's output so every read has **one uniform shape**, and counts the output budget in **UTF-16 code units** instead of Unicode code points.

### 1. One output shape (was two)

Before, `read` returned two different shapes:
- a plain full read that fit the budget → **raw, unnumbered** content;
- a ranged or capped read → an **11-field key=value header** (`path=`, `start_line=`, `end_line=`, `file_chars=`, `selected_chars=`, `shown_chars=`, `line_limited=`, `line_format=`, …) + `---` + numbered content.

The model had to branch on which it got, and full reads had no line numbers (couldn't cite `path:line` afterward). I checked three reference agents — **Claude Code, kimi-cli, kimi-code** — and all three use a single content-first shape: always-numbered lines + a small status line, never a fat header, and never a partial-vs-full branch.

Now openseek matches that. Every read is:
```
1\t<line one>
2\t<line two>
…
<system>start_line=N shown_lines=K total_lines=T truncated=B</system>
```
(footer only when the body is empty, e.g. a range that starts past EOF). The char-count fields are gone — `start_line`+`shown_lines`+`total_lines` already tell the model whether more lines follow, and `truncated` (the budget cut the body) is the only thing that flips `is_error`. Dropping the path from the output makes the result deterministic, so tests now assert exact output.

### 2. UTF-16 code-unit budget

`max_output_chars` is a soft budget, so it's counted in code units (`String::length`, O(1)) rather than code points (`char_length`, O(n)). The truncation cut is snapped down to a character boundary so the body never ends in a lone surrogate — and a mid-surrogate slice in this toolchain actually **panics** (`abort()`), so the guard prevents a crash, not just a garbled char.

## Validation

- `moon fmt` / `moon check` (project-wide) clean
- `moon test agent_tool/read` → 22 passed (full-output snapshots; covers full read, range read, range-past-EOF footer-only, char cap, gutter budget, surrogate boundary, empty file)
- `moon test eval/tool_harness` → 3 passed (real read dispatch)
- e2e: drove the agent to write a TOML parser in MoonBit (see PR comments)
- Independent `codex review` (see PR comments)

### DIFF
--- a/agent_tool/read/README.mbt.md
+++ b/agent_tool/read/README.mbt.md
@@ -25,7 +25,7 @@ Ranged or capped reads include a metadata header so the model knows what it saw
 and what it did not see. The body of those headered blocks is line-numbered as
 `<line-number>\t<content>`, matching the common `cat -n` style used by other
 coding agents.
-Automatic character truncation is marked as a tool error even when the file was
+Automatic output truncation is marked as a tool error even when the file was
 read successfully; that makes lossy context visible to the loop instead of
 silently pretending the returned prefix is complete.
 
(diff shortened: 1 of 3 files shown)


bobzhang:
## Independent review

Ran `codex review --base main` (gpt-5.5, reasoning xhigh) on `18533c4`.

**Verdict:** _"The changes consistently switch read output budgeting to UTF-16 code units and preserve surrogate-pair boundaries during truncation. I did not find an actionable correctness issue introduced by the patch."_

One useful confirmation from its exploration: slicing a `String` mid-surrogate **panics** (`abort()` from `String::sub.inner`), not just emits a lone surrogate — e.g. `"😀Z"[:1]` aborts while `[:2]` yields `"😀"`. So the `char_boundary_at_or_below` / `is_low_surrogate` guard isn't cosmetic; it's what keeps a mid-emoji cut from crashing the tool.

Follow-up `daf547a`: brought the docs up to date — `README.mbt.md` now describes `max_output_chars` and the `file_chars`/`selected_chars`/`shown_chars` header fields as UTF-16 code units and notes truncation snaps to a character boundary. `moon fmt` / `moon check` / `moon test agent_tool/read` (21 passed) all green.

bobzhang:
## Independent review (redesign)

Ran `codex review --base main` (gpt-5.5, xhigh) on `6838481` (the full uniform-shape + code-unit change).

**Verdict:** _"I did not find any actionable issues in the changed read tool implementation or docs. Targeted read tests and moon check passed; the full suite failure appears unrelated to this patch."_

The "full suite failure" it refers to is `deepseek/client/client_realworld_test.mbt` (the realworld DeepSeek API smoke test) failing with `ResolveHostnameError` — i.e. no network in the sandbox, unrelated to this change. The read package was 22/22 and the JS suite 18/18.

bobzhang:
## E2E regression test — write a TOML parser

Built `cmd/openseek` from this branch and drove the agent end-to-end (DeepSeek, `--max-steps 120`, fresh empty workspace):

> _Create a MoonBit package named 'toml' … bare and dotted keys, [tables], string/integer/float/boolean/array values, comments … write unit tests … iterate with 'moon test' until all pass._

**Result: agent finished cleanly in 77 steps; the generated package builds and passes its own tests.** Independently re-verified in the workspace:
- `moon check --target native` → clean
- `moon test` → **14 passed, 0 failed**
- Real parser, not a stub: `pub enum Value { … }`, `pub fn parse(input : String) -> Value raise ParseError`, ~318 lines.

**Read-tool behavior under load (the point of the test):** the agent issued 9+ `read` calls across the run; **every read result carried the new `<system>start_line=.. shown_lines=.. total_lines=.. truncated=..</system>` footer, none malformed, none truncated**, and the model consumed all of them correctly (read scaffolded files → queried `moon ide doc` → edited → `moon test`). These were whole-file reads — exactly the path that *used* to return raw, unnumbered content and now returns numbered lines + footer — so the format change went through the live loop with no confusion or regression. Sample from the run:

```
1\t// Main package file for your library code.
2\t// Put public APIs and shared helpers here.
3\t// Keep blocks small and focused for easy refactors.
4\t
<system>start_line=1 shown_lines=4 total_lines=4 truncated=false</system>
```

PR #229, opened 2026-06-24
simplify: idiomatic pattern matching and list comprehensions

bobzhang:
## What

Behavior-preserving idiomatic cleanups in the CLI / TUI / viz-server packages. No public API change.

- **List comprehensions** replace manual `push` loops:
  - `session_list_json` (cmd/openseek) and `split_output_lines` (cmd/tui).
- **Boolean-from-match → `is`/`&&`**:
  - `flag`, `already_appended`, `line_limited` collapse `match opt { Some(_) => …; None => false }` (and the equivalent `if opt is Some … else false`) into one expression.
- **Redundant guard collapse** in the `value` argument helper (tui + viz_server):
  - `Some([v]) => v` / `Some(vs) if vs.length() > 0 => vs[0]` → a single `Some([v, ..]) => v` arm.

## Validation

- `moon check` clean; `moon fmt` clean.
- Package tests pass (108 in the touched cmd packages); full suite 603/604 (the one failure is the pre-existing DeepSeek API network smoke test).
- Reviewed by **codex** (`codex review --base main`): *"behavior-preserving refactors of loops, pattern matches, and option handling… did not find any discrete regressions introduced by the patch."*

### DIFF
--- a/cmd/tui/main.mbt
+++ b/cmd/tui/main.mbt
@@ -86,10 +86,7 @@ fn cli_command() -> @argparse.Command {
 
 ///|
 fn flag(matches : @argparse.Matches, name : String) -> Bool {
-  match matches.flags.get(name) {
-    Some(value) => value
-    None => false
-  }
+  matches.flags.get(name) is Some(true)
 }
 
 ///|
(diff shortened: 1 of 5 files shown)


cmd/tui/main.mbt
```diff
@@ -86,10 +86,7 @@ fn cli_command() -> @argparse.Command {
 
 ///|
 fn flag(matches : @argparse.Matches, name : String) -> Bool {
-  match matches.flags.get(name) {
-    Some(value) => value
-    None => false
-  }
+  matches.flags.get(name) is Some(value) && value
```
bobzhang: `is Some(true)`?

bobzhang: Good call — done in cb9ada0. `is Some(true)` is cleaner than `is Some(value) && value` and equivalent for the `Bool?` here.

PR #237, opened 2026-06-24
read: suggest a same-stem sibling on file-not-found

bobzhang:
## Summary

Adopts Claude Code's `findSimilarFile` recovery hint for the `read` tool. When a read fails because the path **does not exist**, `read` looks for a sibling in the same directory with the **same stem but a different extension** (e.g. you asked for `foo.ts` but `foo.mbt` is there) and appends ` Did you mean <name>?` to the error — so the agent can retry without an extra `ls`.

Mirrors Claude's heuristic exactly: **same-stem only, no fuzzy/edit-distance matching**, so a suggestion is almost never misleading. It's also gated to fire only on genuine not-found:

- If the path **exists** (the read failed for another reason — permission, invalid UTF-8), no hint.
- If the parent dir is unreadable or nothing shares the stem, no hint.

## Examples

- `read foo.ts` with `foo.mbt` present → `error reading …/foo.ts: … Did you mean foo.mbt?`
- `read README` with `README.md` present → `… Did you mean README.md?`
- `read foo.ts` with only `unrelated.mbt` present → plain error, no suggestion.

## Validation

- `moon fmt` / `moon check` (project-wide) clean, **0 warnings** (deny-warnings gate)
- `moon test agent_tool/read` → **24 passed** (adds same-stem suggestion, extensionless-stem, and no-false-positive cases)
- `moon test eval/tool_harness` → 3 passed (real read dispatch)
- No public API change (`similar_file_hint`/`file_stem` are private), so `pkg.generated.mbti` is unchanged

### DIFF
--- a/agent_tool/read/README.mbt.md
+++ b/agent_tool/read/README.mbt.md
@@ -82,7 +82,9 @@ output truncation. The string body has one of these shapes:
   `is_error` to `true`.
 - `"error reading <path>: <error>"` — a single-file read failed. Common
   causes: the file is missing, the agent doesn't have read permissions, or
-  the bytes aren't valid UTF-8.
+  the bytes aren't valid UTF-8. When the path is missing and a sibling shares
+  its stem (e.g. `foo.ts` requested while `foo.mbt` exists), the message ends
+  with ` Did you mean <name>?` so the agent can retry without an extra `ls`.
 - `"error: read requires arguments.path"` — payload was an object but named no
   file.
 - `"error: read requires object arguments"` — payload was not a JSON object.
--- a/agent_tool/read/moon.pkg
+++ b/agent_tool/read/moon.pkg
@@ -7,6 +7,7 @@ import {
   "moonbitlang/async",
   "moonbitlang/async/fs",
   "moonbitlang/core/json",
+  "moonbitlang/x/path",
 }
 
 warnings = "+unnecessary_annotation"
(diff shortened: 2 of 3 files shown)


bobzhang:
## Independent review (codex)

Ran `codex review --base main` iteratively. It caught three real issues across rounds, each fixed and re-reviewed; the final pass is clean.

1. **Empty dirname for bare-relative paths** — `Path::dirname("README") == ""`, so `readdir("")` dropped the hint. → map empty dirname to `.` (`b2870dd`).
2. **`@fs.exists` re-raises non-ENOENT** — it only maps ENOENT to `false` and re-raises EACCES/ENOTDIR (MoonBit async propagates the raise implicitly), so the probe could turn a clean `error reading …` into an escaped exception for `<file>/child` or an unsearchable parent. → catch the probe, return no hint (`30bdf66`, + ENOTDIR regression test).
3. **Same-stem directory candidates** — for a missing `foo.ts` next to both `foo/` and `foo.mbt`, `readdir` yields `foo` first and the hint pointed at the directory (which `read` rejects). → skip directory candidates, keep scanning for a same-stem file (`d4a819d`, + regression test; a very common `pkg/` + `pkg.mbt` layout).

**Final verdict:** _"The changes add same-stem missing-file suggestions without breaking the existing read behavior… I did not find any actionable correctness issues introduced by this diff."_

`moon test agent_tool/read` → **27 passed**; `moon check` project-wide clean (0 warnings).

PR #239, opened 2026-06-24
viz: lift element children out of call parens with `<|`

bobzhang:
## What

Wholesale formatting refactor of the viz HTML layer: every rabbita element constructor call site is rewritten from

```moonbit
@html.div(class="turn", [ ... ])
```

to

```moonbit
@html.div(class="turn") <| [ ... ]
```

lifting the trailing `children` positional out of the parens via the `<|` apply operator, so the markup tree reads with one consistent shape and the outermost element drops its `])` pileup.

- `viz/view.mbt` — 55 call sites
- `cmd/viz_app/app.mbt` — 46 call sites

## How it was applied

A brace/string/comment-aware transformer lifted the **last positional argument** (always `children` in rabbita's API). Two rules fell out of the compiler:

- **Attribute-less calls are left as-is** — `@html.text(...)`, `@html.strong(name)`, `@html.div([...])` have no attrs to separate, so a `() <|` form would be pure noise.
- **Control-flow children are parenthesised** — `<|` takes a *simple expression* on the right, so the sites with `if` children became `@html.span(class="...") <| (if ... { } else { })`.

## Verification

- `moon check --target js` — clean on `viz` and `cmd/viz_app`
- `moon test --target js` — **194/194 pass** (pure formatting change, no behavior change)
- `moon fmt` — applied and idempotent

### DIFF
--- a/cmd/viz_app/app.mbt
+++ b/cmd/viz_app/app.mbt
@@ -399,13 +399,13 @@ fn events_status(parsed : @session_log.ParsedLog) -> String {
 
 ///|
 fn view(emit : @cmd.Emit[Msg], model : Model) -> @html.Html {
-  @html.div(class=app_class(model), [
+  @html.div(class=app_class(model)) <| [
     sidebar_pane(emit, model),
-    @html.section(class="main", [
+    @html.section(class="main") <| [
       main_toolbar(emit, model),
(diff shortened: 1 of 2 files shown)


bobzhang:
@codex review

PR #244, opened 2026-06-24
reuse core: UInt16::is_trailing_surrogate for char-boundary checks

bobzhang:
## Summary

The read tool (`agent_tool/read`) and `agent_tool/internal/output` each hand-rolled the same predicate to avoid splitting a surrogate pair at a truncation boundary:

```moonbit
fn is_low_surrogate(unit : UInt16) -> Bool { unit >= 0xDC00 && unit <= 0xDFFF }
```

MoonBit core already ships this as `UInt16::is_trailing_surrogate` (a "low surrogate" *is* the trailing half of a pair, U+DC00–U+DFFF). Call the core method at both sites and delete both local copies.

`moon ide doc 'UInt16::*surrogate*'` confirms core also exposes `is_leading_surrogate` / `is_surrogate` if those are needed later.

## Validation

- `moon test agent_tool/read agent_tool/internal/output` → **32 passed**
- Both predicates were private, so `pkg.generated.mbti` is unchanged

### DIFF
--- a/agent_tool/internal/output/output.mbt
+++ b/agent_tool/internal/output/output.mbt
@@ -16,18 +16,11 @@ pub fn cap_text(content : String, max_chars : Int) -> String {
 /// surrogate, step back one so the leading surrogate is dropped too. Only
 /// reached with `0 <= max_units < content.length()`.
 fn char_boundary_at_or_below(content : String, max_units : Int) -> Int {
-  if max_units > 0 && is_low_surrogate(content[max_units]) {
+  if max_units > 0 && content[max_units].is_trailing_surrogate() {
     max_units - 1
   } else {
     max_units
   }
-}
-
-///|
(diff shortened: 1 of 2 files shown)


bobzhang:
## Independent review (codex)

`codex review --base main` — **clean:** _"The change replaces local UTF-16 trailing-surrogate checks with the equivalent standard `UInt16::is_trailing_surrogate` method, and the relevant packages compile and tests pass."_ `moon test` → 32/32.

bobzhang:
Follow-up (`33b8da7`): the two char-boundary sites now use the `s[i]` index form (`content[max_units]`) instead of `.at`/`.code_unit_at`, matching how the rest of the repo accesses code units. A repo-wide grep confirmed these were the only `String::at`/`code_unit_at` calls in openseek's own code, so no separate PR was needed. `moon check` clean, 32/32 tests.

PR #245, opened 2026-06-24
dedupe: share moon --target validation across decode packages

bobzhang:
## Summary

`validate_target` was copy-pasted **identically** in three decode packages — `agent_tool/moon_cmd`, `agent_tool/moon_ide`, `agent_tool/moon_check` — each with the same backend allow-list and the same error message:

```moonbit
fn validate_target(target : String) -> Unit raise {
  let allowed = ["wasm", "wasm-gc", "js", "native", "llvm", "all"]
  if !allowed.contains(target) {
    fail("arguments.target to be one of wasm, wasm-gc, js, native, llvm, all")
  }
}
```

Adding or renaming a backend meant editing three files in lockstep — a drift hazard. Extracted to a small shared `agent_tool/internal/moon_target` package; each decode package now calls `@moon_target.validate_target`. moon_cmd's `validate_command` and `validate_target_supported` stay local (not shared).

Checked the library first per the dedupe guidance — core has no moon-target validator (the allow-list is moon-specific), so this is a plain extract-to-shared.

## Validation

- `moon test` on the four packages → **55 passed**
- The decode packages' public `.mbti` is unchanged (the shared call isn't in any public signature); the new package exposes just `pub fn validate_target(String) -> Unit raise`

### DIFF
--- a/agent_tool/internal/moon_target/moon.pkg
+++ b/agent_tool/internal/moon_target/moon.pkg
@@ -1,0 +1,1 @@
+warnings = "+unnecessary_annotation"
--- a/agent_tool/internal/moon_target/moon_target.mbt
+++ b/agent_tool/internal/moon_target/moon_target.mbt
@@ -1,0 +1,16 @@
+///|
+/// The MoonBit compilation backends accepted by `moon`'s `--target` flag. The
+/// list is shared by every tool that decodes a `target` argument so a new
+/// backend is added in exactly one place.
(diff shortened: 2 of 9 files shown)


bobzhang:
## Independent review (codex)

`codex review --base main` — **clean:** _"The refactor centralizes target validation without changing behavior, and the affected decode packages compile and their existing tests pass."_ `moon test` → 23/23.

PR #246, opened 2026-06-24
dedupe: share moon-manifest path classification (write + edit)

bobzhang:
## Summary

The four manifest-path predicates were duplicated between `agent_tool/write` and `agent_tool/edit/internal/manifest_error`:

- `is_moon_mod_path`, `is_moon_mod_json_path`, `is_moon_pkg_path`, `is_generated_mbti_path`

Same backslash normalization, same suffix checks — just written slightly differently (`write` inlined `replace_all`; `manifest_error` factored a `normalized_path` helper). Extracted to a shared `agent_tool/internal/manifest_path` package; both tools now call `@manifest_path.is_*` so they classify manifests identically from one source.

No library equivalent (these are moon-project-specific), so it's a plain extract-to-shared dedupe — same shape as #245.

### Noted, not done here
`manifest_error.write_error` and `write.manifest_write_error` are themselves nearly identical (same checks + messages). That's a larger overlap between the edit and write tools and is left for a separate, deliberate change rather than folded into this predicate dedupe.

## Validation

- `moon test` on write/edit/manifest_error/manifest_path → **46 passed**
- Consumers' public `.mbti` unchanged (calls aren't in public signatures); new package exposes the four `pub fn is_*`

### DIFF
--- a/agent_tool/edit/internal/manifest_error/manifest_error.mbt
+++ b/agent_tool/edit/internal/manifest_error/manifest_error.mbt
@@ -10,17 +10,17 @@
 /// manifest fail every later build step.
 pub async fn write_error(path : String, content : String) -> String? {
   let trimmed = content.trim()
-  if is_generated_mbti_path(path) {
+  if @manifest_path.is_generated_mbti_path(path) {
     return Some(
       "*.generated.mbti files are generated; run moon info instead of editing them",
     )
   }
-  if is_moon_mod_json_path(path) && !@fs.exists(path) {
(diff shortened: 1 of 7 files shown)


bobzhang:
## Independent review (codex)

`codex review --base main` — **clean:** _"The changes are a straightforward extraction of duplicate path-classification helpers into a shared internal package, with call sites updated consistently. The changed packages type-check and their targeted tests pass."_ `moon test` → 27/27.

PR #247, opened 2026-06-25
dedupe: share char_boundary_at_or_below in agent_tool/internal/utils

bobzhang:
## Summary

`char_boundary_at_or_below` — snap a UTF-16 code-unit offset down to a character boundary so a truncation cut never splits a surrogate pair — was **duplicated** in two places:

- `agent_tool/read/read.mbt` (used by `code_unit_prefix` when capping a numbered line)
- `agent_tool/internal/output/output.mbt` (used by `cap_text` and `truncation_header`)

Moved it to a shared `agent_tool/internal/utils` package; both call sites now use `@utils.char_boundary_at_or_below`, so the surrogate-safety rule lives in exactly one place (and already uses core `UInt16::is_trailing_surrogate` after #244).

### Scope kept minimal
The thin per-tool slices around it — read's `code_unit_prefix` and output's `cap_text` — stay where they are: they're one-line `content[:boundary(...)]` calls that differ only in their empty/over-budget guards, and read better next to their call sites. Only the substantive shared logic (the boundary computation) moves. moon_cmd / moon_ide are untouched (`@output.cap_text` API unchanged).

## Validation

- `moon test` on read / output / utils / moon_cmd / moon_ide → **68 passed**
- New package exposes just `pub fn char_boundary_at_or_below(String, Int) -> Int`; consumers' public `.mbti` unchanged

### DIFF
--- a/agent_tool/internal/output/moon.pkg
+++ b/agent_tool/internal/output/moon.pkg
@@ -1,1 +1,5 @@
+import {
+  "bobzhang/openseek/agent_tool/internal/utils",
+}
+
 warnings = "+unnecessary_annotation"
--- a/agent_tool/internal/output/output.mbt
+++ b/agent_tool/internal/output/output.mbt
@@ -6,20 +6,7 @@ pub fn cap_text(content : String, max_chars : Int) -> String {
   if content.length() <= max_chars {
     content
   } else {
-    "\{content[:char_boundary_at_or_below(content, max_chars)]}"
-  }
-}
-
-///|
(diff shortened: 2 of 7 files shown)


bobzhang:
## Independent review (codex)

`codex review --base main` — **clean:** _"The change is a behavior-preserving extraction of the UTF-16 boundary helper into a shared internal package, and the updated call sites preserve their existing preconditions. Targeted tests for the changed packages pass."_ `moon test` → 68/68.

PR #248, opened 2026-06-25
shell policy: block `sed -i` in-place editing, redirect to the edit tool

bobzhang:
## Summary

`sed -i` editing exits 0 but silently misapplies edits often enough to be untrustworthy — the agent should use the dedicated `edit`/`write` tools. This extends the shell command policy (which already redirects the `moon_cmd` tool-name typo) to reject `sed` carrying an in-place flag, with a message pointing at the `edit` tool.

## What's blocked / allowed

**Blocked** (→ "use the `edit` tool …"):
`sed -i …`, `sed -i.bak`, `sed -i '' …` (BSD), `sed --in-place[=suffix]`, `sed -Ei`/`-ni`, and the same under wrappers / compounds (`command sed -i`, `env FOO=bar sed -i`, `… | sed -i`, `cd src && sed -i`).

**Allowed** (unchanged): read-only `sed` (`sed -n '1,5p' f`, `… | sed s/a/b/`, `sed -e 's/i/x/' f`), unrelated `-i` (`grep -i sed`, `env -i sed …`), and `sed` as data (`echo sed -i`).

## Detection

Follows the existing `moon_cmd` policy's **precision** philosophy: fire only when `sed` is a statically visible simple command. The in-place flag is read off `sed`'s *own* arguments via `SimpleCommand::command_index` (which skips any `env`/`command`/`builtin` wrapper), so `env -i sed …` isn't fooled by env's `-i`. The short-cluster scan stops at `-e`/`-f` (which consume their value), so an `i` inside a script like `-e's/i/x/'` isn't mistaken for `-i`.

**Known gap (documented):** `find … -exec sed -i` and `xargs sed -i` — where `sed` is an argument of another command, not the visible command — are not caught. Catching those reliably needs runtime knowledge the static parser doesn't have, and broadening the match would risk false positives (e.g. `echo sed -i`). Can revisit if it matters.

## Validation

- `moon fmt` / `moon check` (project-wide) clean, 0 warnings
- `moon test agent_tool/shell agent_tool/shell/internal/command_policy` → **42 passed** (new unit tests for blocked/allowed shapes + a shell-level integration test)
- Helpers are private, so `pkg.generated.mbti` is unchanged

### DIFF
--- a/agent_tool/shell/internal/command_policy/README.mbt.md
+++ b/agent_tool/shell/internal/command_policy/README.mbt.md
@@ -3,14 +3,20 @@
 This internal package contains the command-routing policy used by the `shell`
 tool before it runs `sh -c`.
 
-The policy exists for one narrow reason: the bare tool name `moon_cmd`, typed as
-a shell command, is not an executable — it should be the real `moon ...`
-subcommand run through shell. Everything else, including `moon check`, runs
-normally.
+The policy redirects two command shapes that an agent should not run through
+shell:
+
+- the bare tool name `moon_cmd`, typed as a shell command — it is not an
+  executable and should be the real `moon ...` subcommand run through shell;
+- `sed -i` (in-place file editing) — it silently misapplies edits often enough
+  to be untrustworthy, so the agent is sent to the `edit`/`write` tools.
+
+Everything else, including `moon check` and read-only `sed` (`sed -n '1,5p'`,
+`... | sed s/a/b/`), runs normally.
 
 This package is not a security sandbox. It is a local automation guardrail for a
 trusted agent. The shell tool can run arbitrary commands; this package only
-rejects the `moon_cmd` tool-name typo before execution.
+redirects the `moon_cmd` typo and `sed -i` editing before execution.
 
 ## Enforced Policy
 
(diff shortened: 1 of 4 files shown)


bobzhang:

After iterating, `codex review --base main` is clean: _"No actionable correctness issues were found in the changed policy or tests. Targeted command_policy and shell tests pass; the full test suite has an unrelated network-dependent failure."_

### Note on the design evolution
Codex's earlier rounds repeatedly surfaced **false positives** in a rough, quote-unaware fallback I'd added for too-complex parses — e.g. a quoted `|sed -i` inside `grep 'a|sed -i' *.mbt`, and a here-document body that writes `sed -i`. Since a false block (rejecting a read-only command) is worse than missing a globbed edit, the final design **drops the rough fallback entirely**: `sed -i` is blocked solely on the precise, quote-aware `Simple` parse (which correctly handles env/command/builtin wrappers, separators, path-qualified `/usr/bin/sed`, and sed's option grammar — no false positives). Globbed/expansion/heredoc `sed -i` is a documented, tested false negative; the clean way to recover the common globbed case is lexer-level glob tolerance in `shell_parse`, left for a follow-up.

`moon test agent_tool/shell agent_tool/shell/internal/command_policy` → 43 passed.

PR #250, opened 2026-06-25
shell policy: drop the rough sed fallback (false positives)

bobzhang:
## Summary

**Follow-up to #248.** That PR merged at an earlier commit (`main` is at rev "match path-qualified sed") that still carried the rough, quote-unaware complex-parse fallback for `sed -i`. My later fixes (heredoc exclusion, then removing the fallback) were pushed after the merge and never landed — this PR brings the final state in.

### Why
The rough fallback **false-positives** — the worst failure mode for a command guardrail:
- `grep 'foo|sed -i' *.mbt` — a read-only grep — gets blocked, because the quote-unaware split breaks at the quoted `|` and sees a `sed -i'` segment;
- `cat > script.sh <<EOF\nsed -i 's/a/b/' f\nEOF` — writing a script via heredoc — gets blocked, because the scan reads the heredoc body as commands.

A false block (rejecting a legitimate read-only command) is worse than missing a globbed edit.

### What changes
`sed -i` is now blocked **solely on the precise, quote-aware `Simple` parse**, which already handles `env`/`command`/`builtin` wrappers, separators, pipes, path-qualified `/usr/bin/sed` (basename), and sed's own option grammar (`-e`/`-f` operands, a script file named `-i`) — with **no false positives**. Globbed / expansion / heredoc `sed -i` parses as `TooComplex` and is allowed: a documented, tested false negative. The clean way to recover the common globbed case is lexer-level glob tolerance in `shell_parse` (also helps the `moon_cmd` policy), left for a separate PR.

Removes ~90 lines of fragile heuristics (segment splitting, quote stripping, env prefixes, heredoc exclusion).

## Validation

- `moon test agent_tool/shell agent_tool/shell/internal/command_policy` → **43 passed** (incl. explicit tests that the former false-positive cases are now allowed, and that globbed `sed -i` is a documented allow)
- Independent `codex review` (see comments) — clean after this simplification

### DIFF
--- a/agent_tool/shell/internal/command_policy/README.mbt.md
+++ b/agent_tool/shell/internal/command_policy/README.mbt.md
@@ -42,36 +42,44 @@ The policy allows:
 
 ## Detection Strategy
 
-The implementation is deliberately conservative:
+Detection is **precise, not heuristic**, and prefers **no false positives**
+(never block a command that is not actually an in-place edit) over exhaustive
+recall:
 
 - Parse the command with the internal shell policy parser.
 - Reject `moon_cmd` when it is a statically visible simple command name,
   including in flat compounds such as `cd demo && moon_cmd check`.
 - Reject `sed` carrying an in-place flag when it is a statically visible simple
-  command (after any `env`/`command`/`builtin` wrapper). The in-place flag is
-  read off `sed`'s own arguments, so an unrelated `-i` such as `grep -i sed` or
-  `env -i sed ...` does not trip it.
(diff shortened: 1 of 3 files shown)


bobzhang:

`codex review --base main`: _"No actionable correctness issues were identified in the changed command policy behavior."_ (the one full-suite failure is the unrelated network-dependent DeepSeek smoke test). `moon test agent_tool/shell agent_tool/shell/internal/command_policy` → 43 passed.

PR #253, opened 2026-06-25
session analyzer: count shell failures from is_error, not output text

bobzhang:
## Summary

The eval session analyzer decided shell failures by **re-parsing the shell tool's output string** for an `exit=` line (`shell_result_failed`). That coupled the analyzer to the shell output format — it's the only reason a recent shell output-format change had to touch the analyzer at all.

`ToolResult` already records an `is_error : Bool` (the tool decided success/failure when it produced the result), and the analyzer **already** uses `result.is_error()` for `tool_errors`. This switches `shell_failures` to the same field and deletes the string parser, decoupling the analyzer from the output format for good.

Inspired by Claude Code, whose tools carry structured results (`{ exitCode, stdout, stderr }` + an `isError` flag) and whose consumers read those fields rather than re-parsing the rendered text.

## Semantics note

`is_error` is slightly broader than the old parse: a timeout, bad arguments, a launch failure, or a policy block (`sed -i`, `moon_cmd`) now counts as a shell failure — consistent with `tool_errors`, and arguably more correct ("the shell call failed"). The old parse only counted non-zero exit / cancellation.

## Deferred follow-up

Command-aware exit semantics — e.g. `moon ide doc` with no result, or `grep` with no match (exit 1), should **not** be a failure — is a separate, larger change (the shell tool would need to interpret exit codes per command, the way Claude's `commandSemantics` does). Left for later.

## Validation

- `moon test eval/session_analyzer eval/prompt_task/harness` → 24 passed (new test asserts shell_failures counts only `is_error` shell results, ignores the output text, and a non-shell error counts toward `tool_errors` but not `shell_failures`)

### DIFF
--- a/eval/session_analyzer/analyzer.mbt
+++ b/eval/session_analyzer/analyzer.mbt
@@ -116,7 +116,10 @@ fn summarize_session(session : @agent_session.Session) -> SessionMetrics {
         tool_results += 1
         if result.tool_name() == "shell" {
           shell_uses += 1
-          if shell_result_failed(result.content()) {
+          // Use the recorded `is_error` flag rather than re-parsing the shell
+          // output: the tool already decided success/failure, so this stays
+          // correct regardless of the output format.
+          if result.is_error() {
             shell_failures += 1
           }
         }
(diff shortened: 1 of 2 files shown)


bobzhang:

Fresh `codex review --base main`: _"No actionable correctness issues were found in the diff. The targeted session analyzer tests pass; the full suite only failed on an unrelated network-dependent smoke test."_ `moon test eval/session_analyzer eval/prompt_task/harness` → 24 passed.

PR #258, opened 2026-06-26
refactor(read): inline single-use helpers, share code_unit_prefix

bobzhang:
## What

The `read` tool accumulated several helpers each called exactly once, plus two structs threading the same fields through a `join("\n")`→`split("\n")` round-trip. This refactor removes that indirection without changing behavior.

### Changes
- **Inline single-use helpers** into `read_file`: `directory_read_error`, `select_lines`, `render_numbered_content`, and `read_footer` are folded into one selection/render block that numbers the `selected` lines directly (dropping the redundant join→split round-trip).
- **Collapse two structs into one**: `ReadSelection` + `RenderedSelection` → a single `RenderedSelection` carrying `content`, `start_line`, `total_lines`, `shown_lines`, `truncated`.
- **Lift `code_unit_prefix` into `agent_tool/internal/utils`** so it lives next to `char_boundary_at_or_below`, the surrogate-safe cut it already shares with the `output` package. Both now take a `StringView`. Added unit tests for `code_unit_prefix`.
- **Fold `execute_with_workspace` into `execute`** via an optional `workspace_root` parameter.

## Behavior

No behavior change. The equivalences:
- directory check: old returned `None` (proceed) when `@fs.kind` threw; new returns `false` (proceed) — same path, identical error message.
- line selection/rendering: same clamps, same before-add `max_lines` overflow guard, same `start_line + index` numbering.
- footer: identical fields, now read off the merged struct.

## Testing
- `moon check agent_tool/read agent_tool/internal/utils agent_tool/internal/output` — clean (incl. the `String`→`StringView` coercion at `output.mbt` call sites).
- `moon test` on the three packages — 34/34 pass.

### DIFF
--- a/agent_tool/internal/utils/pkg.generated.mbti
+++ b/agent_tool/internal/utils/pkg.generated.mbti
@@ -4,6 +4,8 @@ package "bobzhang/openseek/agent_tool/internal/utils"
 // Values
 pub fn char_boundary_at_or_below(String, Int) -> Int
 
+pub fn code_unit_prefix(String, Int) -> String
+
 // Errors
 
 // Types and methods
--- a/agent_tool/internal/utils/utils.mbt
+++ b/agent_tool/internal/utils/utils.mbt
@@ -26,6 +26,8 @@ pub fn code_unit_prefix(content : StringView, max_units : Int) -> StringView {
   } else if content.length() <= max_units {
     content
   } else {
-    "\{content[:char_boundary_at_or_below(content, max_units)]}"
+    // The slice is already a StringView, so return it directly rather than
+    // round-tripping through interpolation (which would allocate a String).
+    content[:char_boundary_at_or_below(content, max_units)]
   }
 }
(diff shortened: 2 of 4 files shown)


bobzhang:
Addressed the Copilot review comment in 29c1c86: `code_unit_prefix` now returns the `StringView` slice directly instead of round-tripping through string interpolation (which allocated a `String`). `moon check` clean and 34/34 tests pass.

Codex CLI review (`codex exec review --base main`, gpt-5.5) found no actionable regressions and confirmed behavior is preserved.

Note: the failing `check (nightly)` job is unrelated to this PR — it's `moon fmt --check` rewriting manifest syntax in `prompt/moon.pkg` (a file this PR doesn't touch) due to an overnight nightly-toolchain formatter change. `check (stable)` passes.

bobzhang:
Pushed two more commits continuing the refactor:

- **eab36bb** — build the read output with a single `StringBuilder` via the `<+` streaming-interpolation idiom (numbered lines + footer), dropping the `Array[String]`+`join` body, the final string concatenation, and the `RenderedSelection` struct. The empty-selection / zero-budget special cases are folded into the render loop, which already yields the same empty-body+`truncated` result.
- **2df33f9** — addressed the Copilot comment on the `max_lines` clamp comment: it referenced a previous post-add clamp no longer in the tree; now states the overflow invariant directly.

Verification: `moon check` clean, `moon fmt` applied, 34/34 read/utils/output tests pass. A second `codex exec review --base main` (gpt-5.5) found no actionable correctness issues in the updated diff.

PR #252, opened 2026-06-25
Add multi_edit tool for line-anchored batch edits

bobzhang:
## Summary

Rebuilt on current `main` and narrowed to the core `multi_edit` change.

- Add `multi_edit`, a single-file batch edit tool for multiple explicit line-anchored replacements.
- Register `multi_edit` in the agent/tool registry.
- Keep `edit` as the single-replacement tool and reject `replace_all=true` with guidance to use explicit `multi_edit` entries.
- Add focused docs and tests for `edit`/`multi_edit` behavior.

## Split Out

- Shell source-write sandboxing moved to #259.
- Eval harness, benchmark, and Markdown-generation script changes were removed from this PR.

## Validation

- `moon test agent_tool/multi_edit agent_tool/edit agent --target native --diagnostic-limit 20`
- `moon info`
- `moon fmt --check agent_tool/multi_edit agent_tool/multi_edit/internal/decode agent_tool/edit agent`

PR #259, opened 2026-06-26
Block shell source writes from agent shell

bobzhang:
## Summary
- Run agent shell commands through a source-write sandbox when `sandbox-exec` is available, while preserving normal read/build output behavior.
- Pre-block obvious shell source rewrites such as redirects, `sed -i`, and scripted source tree moves before execution.
- Add sandbox feedback and prompt/tool-description guidance that routes compiler-feedback fixes through line-anchored `edit` rather than shell-generated rewrites.

This PR is intentionally independent of #252 and does not reference `multi_edit`.

## Validation
- `moon test agent_tool/shell agent_tool/shell/internal/command_policy agent_tool/shell/internal/shell_parse agent_tool/internal/auto_check agent_tool/write prompt --target native --diagnostic-limit 20`
- `moon info`
- `moon fmt --check agent_tool/shell agent_tool/shell/internal/command_policy agent_tool/shell/internal/shell_parse agent_tool/write prompt`

### DIFF
--- a/agent_tool/internal/auto_check/auto_check.mbt
+++ b/agent_tool/internal/auto_check/auto_check.mbt
@@ -10,8 +10,8 @@ const MoonCheckCommand : String = "moon check --diagnostic-limit 1"
 ///|
 /// Append bounded raw `moon check` feedback after successful writes to MoonBit
 /// source or manifest files. The check runs from the containing MoonBit module
-/// root. Non-MoonBit paths and paths outside a MoonBit module are returned
-/// unchanged.
+/// or workspace root. Non-MoonBit paths and paths outside a MoonBit project are
+/// returned unchanged.
 pub async fn append_summary(path : String, content : String) -> String {
   let summary = auto_check_summary(path) catch {
     error if @async.is_being_cancelled() => raise error
@@ -50,7 +50,7 @@ async fn moon_check_cwd(path : String) -> String? {
   if !is_relevant_moonbit_path(path) {
     return None
   }
-  nearest_moon_module_dir(containing_dir(path))
+  nearest_moon_project_dir(containing_dir(path))
 }
 
 ///|
@@ -92,6 +92,8 @@ fn is_relevant_moonbit_path(path : String) -> Bool {
   path.has_suffix("/moon.mod") ||
   path == "moon.pkg" ||
   path.has_suffix("/moon.pkg") ||
+  path == "moon.work" ||
+  path.has_suffix("/moon.work") ||
   is_moonbit_source_path(path)
 }
 
@@ -112,9 +114,10 @@ fn containing_dir(path : String) -> String {
 }
 
 ///|
-async fn nearest_moon_module_dir(start : String) -> String? {
+async fn nearest_moon_project_dir(start : String) -> String? {
   for dir = start {
-    if @fs.exists(join_path(dir, "moon.mod")) {
+    if @fs.exists(join_path(dir, "moon.mod")) ||
+      @fs.exists(join_path(dir, "moon.work")) {
       return Some(dir)
     }
     match parent_dir(dir) {
@@ -169,6 +172,7 @@ fn join_path(dir : String, base : String) -> String {
 test "relevant path detection covers MoonBit source and manifests" {
   assert_true(is_relevant_moonbit_path("moon.mod"))
   assert_true(is_relevant_moonbit_path("lib/moon.pkg"))
+  assert_true(is_relevant_moonbit_path("moon.work"))
   assert_true(is_relevant_moonbit_path("lib/parser.mbt"))
   assert_true(is_relevant_moonbit_path("lib/README.mbt.md"))
   assert_false(is_relevant_moonbit_path("lib/parser.mbti"))
--- a/agent_tool/internal/auto_check/auto_check_test.mbt
+++ b/agent_tool/internal/auto_check/auto_check_test.mbt
@@ -145,6 +145,17 @@ async test "append_summary preserves module dependency graph failure text" {
 }
 
 ///|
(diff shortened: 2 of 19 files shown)


bobzhang:
Update after rebase/review on `b739b3a`:

- Rebased on `origin/main` / `upstream/main` at `44bcc1e`.
- Kept the sandbox in `agent_tool/shell` rather than extracting a dedicated package. The policy depends on local shell parsing, Moon command trust rules, cwd tracking, and source-tree detection, so extracting it now would expose/duplicate internal semantics without a clean package boundary.
- Addressed the latest fresh Codex review findings:
  - TooComplex dynamic preblocks are now workspace-scoped and respect static `cd` before the source-writing command, so `/tmp` source-looking writes such as heredoc or `python -c` temp writes are not blocked when outside the workspace.
  - Created-directory predictions are preserved across `||`, so masked shapes like `mkdir gen || true; cp /tmp/generated.mbt gen 2>/dev/null || true` are preblocked instead of reporting success.
- Added focused regressions for the two review cases plus shell-level coverage for the masked created-directory copy.
- E2E warning-fix experiment result remains positive: the agent used shell for discovery/checks and edit for source changes; no shell source rewrite. Fixture log: `/tmp/openseek-warning-e2e/openseek.jsonl`.

Validation:

- `moon test agent_tool/shell --target native --diagnostic-limit 20` -> 100/100 passed.
- `moon test agent_tool/shell agent_tool/shell/internal/command_policy agent_tool/shell/internal/shell_parse agent_tool/internal/auto_check agent_tool/write prompt --target native --diagnostic-limit 20` -> 181/181 passed.
- `moon check --target all --diagnostic-limit 20` -> clean.
- `moon info && moon fmt` -> clean.
- Full `moon test --target native --diagnostic-limit 20` -> 776/776 passed.
- GitHub CI: `check (stable)` passed. `check (nightly)` is still failing and was intentionally ignored per scope.

I attempted one more fresh `codex review --base upstream/main` after these fixes, but the Codex CLI hit the usage limit before completing the review, so there is no final post-fix external review verdict from that rerun.

bobzhang:
Pushed a follow-up commit with a review pass + a fresh `codex review --base origin/main`.

## Profile simplification (validated equivalent)

The SBPL profile hand-rolled a negative directory-component regex to keep `_build`/`.mooncakes` writable:

```
not_build = "(_|_b|_bu|_bui|_buil|_build[^/]+|_[^b/][^/]*|_b[^u/][^/]*|...)"
not_mooncakes = "(\\.|\\.m|...|\\.mooncakes[^/]+|...|\\.mooncake[^s/][^/]*)"
```

`sandbox-exec` SBPL is **last-match-wins** (verified empirically), so this collapses to a `deny <source>` followed by an `allow <under _build/.mooncakes>` carve-out:

```
(deny  file-write* (regex "^<root>(/|$)(.*/)?<source_file>$"))
(allow file-write* (regex "^<root>/(.*/)?(_build|\.mooncakes)(/|$)"))
```

I checked this against real `sandbox-exec` across the full path matrix the tests enumerate — `src/main.mbt`, `moon.mod`, `_build/.../main.mbt`, `pkg/_build/main.mbt`, `scripts/.mooncakes/pkg/main.mbt`, and the tricky `_buildx/main.mbt`, `.mooncakesx/main.mbt`, `pkg/_build.mbt` (a *file* named `_build.mbt`, not a dir) — all behaviorally identical to the old regex.

Also dropped the redundant `(deny file-write-create ...)` rule (a probe confirms `file-write*` already blocks *creating* a new source file) and its dead `source_write_create_regex` indirection.

Added a functional regression that exercises real `sandbox-exec` **allowing** a `_build/main.mbt` write — previously only the deny side had functional coverage, and the deny/allow split is exactly what this change touches.

## Codex review findings

Codex flagged two Windows-only PowerShell bypasses (Windows has no `sandbox-exec` fallback, so the pre-block is the only guard there):

- **[P2] `New-Item -Path . -Name main.mbt`** — the guard only inspected `-Path`, so it examined `.` and missed the leaf. **Fixed** in this commit by combining `-Path` + `-Name` (`powershell_new_item_targets`), with a detection regression test.

- **[P2] PowerShell `cp`/`mv`/`rm` aliases routed to the POSIX parser** — on Windows `cp` is an alias for `Copy-Item` and accepts `-Path`/`-Destination`, but the dispatch hardcodes `cp`/`mv`/`rm`/`tee` to POSIX parsers on all platforms, so `cp -Path x -Destination main.mbt` slips through. **Not fixed here** — the proper fix is platform-gated (`#cfg(platform="windows")`) dispatch to the existing PowerShell handlers, which I can't compile-check or functionally validate on this macOS host (and there's no Windows CI to catch a regression). Recommend handling it as a follow-up with Windows compilation, ideally covering `mv`/`rm`/`tee` aliases too, since they share the structural gap.

## Validation

- `moon test agent_tool/shell agent_tool/shell/internal/command_policy agent_tool/shell/internal/shell_parse agent_tool/internal/auto_check agent_tool/write prompt --target native` → 182/182.
- `moon test --target native` → full suite green.
- `moon check --target all` → clean (only the pre-existing `formatter` deprecation warning).
- `moon fmt` → clean.

## Note on scope

The broader observation from the review: `sandbox.mbt` is ~6.4k lines / 300+ functions for what is a *soft* guardrail (the agent isn't adversarial, and on macOS the `sandbox-exec` hard block is the real boundary). The `TooComplex` text-scanning pre-blocks re-implement a shell parser via word-scanning per language (git/sed/PowerShell/python/node) and are the main source of both false-positive risk and the long tail of bypasses like the two above. Worth considering whether the Windows/`TooComplex` heuristic surface earns its maintenance cost, or whether it should be scoped down to the high-signal cases.

PR #264, opened 2026-06-28
prompt: emphasize compiler-guided refactoring/fixing (top-placed)

bobzhang:
## Summary

Elevate **compiler-guided refactoring/fixing** to a prominent principle at the top of the system prompt: let `moon check` (fast, sound) be the source of truth for *what* to change and *where*; do not locate edit sites with regex/`sed` or by syntax/AST reasoning, because `a.method()` resolves by the *type* of `a` — only the type checker knows which uses are real. Includes the rename-via-deprecation workflow (add new name, deprecate old as an alias, fix the flagged uses, then delete the old name).

Placement matters: the same guidance buried mid-prompt was inert; at the top it changes behavior (below).

## Why / evidence

Driving the agent over a real migration (264 `length`→`len` deprecations across a tree), the agent's reflex is a regex codemod — `sed`, a python "replace .length everywhere" script, or a whole-file `write`. Those are *wrong*, not just unsafe: text/syntax matching rewrites comments, strings, and `.length()` on types that have no `.len()`, and misses spacing variants. The compiler points only at the real, type-resolved sites.

A/B (this prompt vs `origin/main`, both-tools, 5 runs each), counting the **full** shortcut behavior (sed + codemod scripts + whole-file writes):

| | shortcut attempts | steps (median) | completed |
|---|---|---|---|
| origin/main | **17** | 135 | 4/5 |
| + top-placed guidance | **6** (~3× fewer) | 69 | 5/5 |

(A first pass that counted only `sed -i` showed no difference; the effect is real once python-script and whole-file-write shortcuts are counted.) A larger n=15, diff-verified, run is in progress to firm up the steps/completion side.

Prompt-only, zero behavioral risk; codex review found no issues.

### DIFF
--- a/prompt/flash_prompt.mbt.md
+++ b/prompt/flash_prompt.mbt.md
@@ -3,6 +3,18 @@ You are OpenSeek, a MoonBit coding agent focused on implementing user tasks.<!--
 Use the native tools to inspect, create, edit, validate, and finish work. If
 work is needed, call a tool. When the task is complete, call `finish`.
 
+Refactor and fix by compiler guidance, not by text. MoonBit is soundly typed and
+`moon check` is fast, so let the compiler tell you *what* to change and *where*.
+Do not find edit sites with regex/`sed` or by syntax/AST reasoning — `a.method()`
+resolves by the *type* of `a`, so text matching hits the wrong occurrences,
+comments, strings, and other types, and misses spacing variants; only the type
+checker knows which uses are real. Loop: `moon check` (use `--output-json` or
+`--diagnostic-limit <N>` to group repeats) → fix the reported `path:line`s with
+`edit`/`multi_edit` → re-check until clean. To rename an API, add the new name,
+make the old one a deprecated alias, and fix the deprecations the compiler then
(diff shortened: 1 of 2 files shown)


bobzhang:
**Correction (n=15 follow-up).** The benefit numbers in this PR body — ~3× fewer codemod shortcuts, ~½ the steps — came from **n=5 and did not replicate at n=15.**

Fresh, diff-verified 10-vs-10 (same task, both-tools):

| | steps median | shortcuts | clean diffs |
|---|---|---|---|
| control | 114 | 20 | 5/10 |
| top-placed | 137 | 24 | 3/10 |

i.e. top-placed was **not faster, not fewer shortcuts, not cleaner**. The original "win" traced to one anomalously fast early top batch (median 69); the pooled n=15 medians look favorable only because that batch drags them down.

**Keeping the change** — the compiler-guided *principle* is sound and the prompt edit is not measurably harmful (≈neutral) — but the quantitative claims above should not be relied on. Separately, diff-verification showed **both** arms make collateral / incomplete edits that `moon check`=0 hides; that's the higher-leverage problem, which #265 (insertion) and #266 (write-guard) begin to address.

PR #276, opened 2026-07-01
feat(viz): connect each tool call to its result in the log view

bobzhang:
## Problem

In the browser viz log viewer, when one assistant turn emits **multiple tool calls**, the tool-call rows render together inside the assistant card, while each tool *result* renders as a separate top-level card whose header shows only the tool **name** (`Tool result · <name>`). With two calls to the same tool in one turn (e.g. two `read`s), the result cards were distinguishable only by order and content — there was no visible tie back to which call produced them.

## Approach

The call↔result linkage already exists in the data model (`@deepseek.ToolCall.id` / `ToolResult.tool_call_id`, matched by the projection), so **this is a rendering-only change** — no data-model changes.

- Assign each tool call a stable, 1-based **pair number** and show the same `call N` chip on both the tool-call row and its result card.
- Each chip is an anchor that **jumps to its partner** (call → result, result → call), with a `:target` outline on arrival.
- A number is assigned only to an id that forms an **unambiguous 1:1 pair** in the rendered view: non-empty, appearing exactly once among the calls *and* exactly once among the results. This deliberately skips:
  - empty ids (a streaming glitch can leave `id=""`),
  - duplicate ids,
  - orphan results (a result with no matching call),
  - dangling calls (a call with no result).

  So a chip never links to a missing anchor and no DOM id is ever emitted twice.
- **Raw view** pairs from durable events; **model view** pairs from the projected messages (so a healed synthetic tool message is a valid partner) and uses only the *structural id* — the error flag, tool name, and brief still flow solely through `model_shown_result`, preserving the model-view enrichment invariant.

## Files

- `viz/view.mbt` — pairing helpers (`assign_pair_tags`, `raw_pair_tags`, `model_pair_tags`, anchor/`pair_link` helpers); thread the id→N map through the raw and model render paths; add anchor `id`s and `call N` chips to tool calls and results.
- `web/index.html` — `.tool-link` chip styling + `:target` highlight for `.tool-call` / `.card`.
- `viz/viz_test.mbt` — update the exact-markup assertions that now carry an `id`; add tests for the two-same-tool-calls case (reciprocal anchors + shared chip, in both raw and model view) and for the unpairable edge cases (empty / duplicate / orphan ids produce no pairing markup).

## Testing

- `moon test viz --target js` — 30 passed
- `moon test cmd/viz_server` — 15 passed; `moon test cmd/viz_app --target js` — 18 passed
- `moon check --target js` + `moon check` (native) clean; `moon fmt` clean
- `moon build cmd/viz_app --target js` — JS bundle builds

## Known minor

The chip lives inside a `<summary>` (so it stays visible when the card/call is folded). Clicking a link inside a `<summary>` may also toggle that `<details>` in some browsers, in addition to navigating — a cosmetic side effect on the element you're leaving; navigation to the partner still works.

## Plan review

This was designed and reviewed against a fresh Codex CLI pass before implementation (which surfaced the empty/duplicate/orphan-id handling now enforced by the 1:1 pairing rule).

### DIFF
--- a/viz/view.mbt
+++ b/viz/view.mbt
@@ -60,11 +60,12 @@ pub fn render_notices(parsed : @session_log.ParsedLog) -> @html.Html {
 fn render_raw(session : @agent_session.Session) -> @html.Html {
   let briefs = tool_call_briefs(session)
   let failed = raw_failed_call_ids(session)
+  let tags = raw_pair_tags(session)
   let step_labels = agent_step_labels(session)
   let turns = group_turns(session.events())
   let blocks : Array[@html.Html] = []
   for index, turn in turns {
-    blocks.push(turn_block(index + 1, turn, briefs, failed, step_labels))
+    blocks.push(turn_block(index + 1, turn, briefs, failed, tags, step_labels))
   }
   if blocks.is_empty() {
     return empty_state("This session has no events yet.")
@@ -162,6 +163,7 @@ fn turn_block(
   turn : Array[@agent_session.SessionEvent],
   briefs : Map[String, String],
   failed : Map[String, Unit],
+  tags : Map[String, Int],
   step_labels : Map[Int, String],
 ) -> @html.Html {
   // Offsets are relative to the turn's first stamped event: for reading a
@@ -175,6 +177,7 @@ fn turn_block(
         time=offset_label(base, event),
         briefs,
         failed,
+        tags,
         step_label=step_labels.get(event.sequence()).unwrap_or(""),
       )
     }
@@ -292,6 +295,7 @@ fn event_card(
   time~ : String,
   briefs : Map[String, String],
   failed : Map[String, Unit],
+  tags : Map[String, Int],
   step_label~ : String,
 ) -> @html.Html {
   let seq = event.sequence()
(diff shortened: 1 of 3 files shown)


bobzhang:
## Fresh Codex CLI review — addressed

Rebased onto latest `origin/main` and ran a fresh, independent Codex CLI review of the diff. **Verdict: no blockers**, both on the initial pass and on the re-review after fixes.

Addressed from the review:
- **`<a>` inside `<summary>` toggle risk** — added a delegated click handler in `web/index.html`: clicking a `.tool-link` chip cancels the default (which would also fold the enclosing `<details>`) and drives the `:target` hash jump + centered scroll itself; modified/non-primary clicks fall through to the browser (open-in-new-tab still works).
- **Test coverage gaps** — added a duplicate-result case (ambiguous result side → not paired in raw view) and a dangling-call case (unique call id, no result → pairs only in the model view, where the projection heals it with a synthetic result).
- **Anchor-uniqueness** — the multi-call test now asserts each `tool-call-N` / `tool-result-N` id is emitted exactly once and no stray third pair exists, so every chip `href` resolves to exactly one target.
- **NIT** — dropped the redundant `!tags.contains(id)` guard (the `call_count == 1` check already guarantees a single assignment).

Remaining suggestion (non-blocking, acknowledged): the summary-click JS behavior isn't covered by a browser-level test — the MoonBit tests assert rendered HTML only.

Verification: `moon test viz --target js` 34/34, `moon check` (js + native) clean, `moon fmt` clean, `moon build cmd/viz_app --target js` builds.

PR #278, opened 2026-07-01
feat(cli): single binary — host the terminal UI as `openseek tui`

bobzhang:
## Problem

OpenSeek shipped **two** native binaries — `cmd/openseek` (the agent engine/CLI)
and `cmd/tui` (the terminal UI) — which made distribution awkward and created a
stale-binary footgun: the TUI spawned `openseek` from `PATH`, so a rebuilt engine
that wasn't reinstalled was silently mismatched.

## Approach

Consolidate into **one** `openseek` binary while keeping the good design — the UI
drives the engine over the existing stdio-JSONL `--serve` protocol as a
subprocess (crash isolation, `--engine` override for replay/fakes all preserved).

- **`openseek tui …`** launches the UI; every other invocation is the engine/CLI
  exactly as before. `cmd/tui` becomes a library exposing `pub async fn run(argv~, env~)`;
  `cmd/openseek`'s `main` peels the first `tui` token and dispatches. It's peeled
  before argparse (not modeled as an argparse subcommand) because the engine takes
  a free-form positional task that a subcommand would make ambiguous.
- **Self-relaunch (no C stub).** When `--engine`/`OPENSEEK_ENGINE` isn't set, the
  TUI defaults its engine to *this* running binary via `argv[0]`: a path-like
  `argv[0]` (`./openseek`, an absolute path, a `moon run` build path) is used
  directly, and a bare name falls back to `openseek` on `PATH` — which, for a
  single binary, is the same file. Combined with consolidation this removes the
  stale-engine footgun with pure MoonBit (no new FFI). An earlier draft used a
  `/proc/self/exe` + `_NSGetExecutablePath` C stub; dropped per review feedback to
  avoid stubs.

## Commits

1. `feat(cli)` — host the TUI as `openseek tui` (make `cmd/tui` a library, dispatch, rename the argparse command to `openseek tui`, update cram + docs).
2. `feat(tui)` — default the engine to this running binary via `argv[0]` (`with_engine_self_default` / `default_engine`, both unit-tested; handles `/` and Windows `\`).
3. `docs` — document the reserved `tui` token (`openseek -- tui …` escape) and clarify the engine paragraph.

## Behavior notes

- `--engine` / `OPENSEEK_ENGINE` still win (replay/fake engines unaffected).
- Engine-path behavior is otherwise unchanged; `parse_config`'s static default
  stays `openseek`, so `--help` output and the cram suite are stable.
- The first argument `tui` is reserved; run a task literally starting with `tui`
  via `openseek -- tui …`.
- TUI arg errors now flow through the shared dispatcher to **stderr**, matching
  the engine's own CLI (the old standalone `tui` printed to stdout); `tests/cram/tui.md`
  updated accordingly and now drives `openseek.exe tui`.

## Testing

- `moon test` — 850 passed (native)
- `moon check --deny-warn`, `moon fmt --check`, `moon info` (no `.mbti` drift) — clean
- `moon cram test tests/cram` — 18/18 (help banner, missing-key, unknown-option, fake-engine preflight, all via `openseek tui`)

## Review

Plan and each commit were reviewed by a fresh Codex CLI pass; the final holistic
review returned **no blockers** ("ready to merge"). Feedback incorporated: dropped
the C stub in favor of `argv[0]`, folded the tui→lib + dispatch + cram/docs into
one green commit, Windows separator handling, and the reserved-token doc.

### DIFF
--- a/README.mbt.md
+++ b/README.mbt.md
@@ -17,7 +17,7 @@ entry point so request encoding can be tested without network access.
 | `bobzhang/openseek/agent_tool` | Tool registry, executor, output, and control-action types. | `agent_tool/README.mbt.md` |
 | `bobzhang/openseek/agent` | Native-only OpenSeek agent loop and local tool dispatch. | `agent/README.mbt.md` |
 | `bobzhang/openseek/cmd/openseek` | Native-only command-line entry point. | `cmd/openseek/README.md` |
-| `bobzhang/openseek/cmd/tui` | Native-only terminal UI that drives the engine per prompt. | `cmd/tui/README.md` |
+| `bobzhang/openseek/cmd/tui` | Native-only terminal UI library, launched as the `openseek tui` subcommand. | `cmd/tui/README.md` |
 | `bobzhang/openseek/testkit/filesystem` | JSON-backed virtual filesystem for tests and eval fixtures. | `testkit/filesystem/README.mbt.md` |
 | `bobzhang/openseek/eval/report` | Shared Markdown/JSON report primitive for deterministic and model evals. | `eval/report/README.mbt.md` |
 | `bobzhang/openseek/eval/tool_harness` | Deterministic host-side harness that dispatches every built-in tool. | `eval/tool_harness/README.mbt.md` |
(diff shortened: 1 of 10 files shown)


bobzhang:
Superseding with a clean-slate PR off main: the future-proof subcommand CLI (TUI default + run/serve/review/sessions), which reworks this dispatch. Closing in favor of that.

PR #279, opened 2026-07-01
feat(kimi): support K2.7 Code models

tonyfettes:
This PR adds support for Kimi K2.7 Code model to OpenSeek.

The API_KEY can be read from either environment variable `KIMI` or command line argument `--kimi-api-key`. `cmd/openseek` requires `DEEPSEEK` to be set for DeepSeek V4 Flash/Pro models, or `KIMI` to set to use Kimi K2.7 Code models.

A notable difference between Kimi K2.7 Code and DeepSeek V4 Flash/Pro is how they handle the reasoning content. Kimi K2.7 Code requires all `reasoning_content` to keep as-is when echoing back conversation history back to API, while DeepSeek V4 Flash/Pro requires `reasongin_content` only in multi-turn tool call. See

- Kimi K2.7 Code: https://platform.kimi.ai/docs/guide/use-kimi-k2-thinking-model
- DeepSeek V4 Flash/Pro: https://api-docs.deepseek.com/guides/thinking_mode

Therefore we modify `ChatMessage::to_json` so that it takes an extra `model` parameter to decide if to include `reasoning_content` when serializing to JSON before requests.

We didn't introduce a provider layer in order to keep the PR small. If we are going to add more models in the future, it might be better to introduce a proper provider-based model resolution mechanism.

### DIFF
--- a/cmd/openseek/README.md
+++ b/cmd/openseek/README.md
@@ -35,10 +35,12 @@ moon run cmd/openseek -- run [--api-key sk-...] [--model deepseek-v4-pro] [--api
 ```
 
 Runs require `--api-key` or `DEEPSEEK`. `--model` can also be supplied with
-`DEEPSEEK_MODEL`; it defaults to `deepseek-v4-pro`. `--max-steps` can also be
-supplied with `OPENSEEK_MAX_STEPS`; it defaults to `1000`.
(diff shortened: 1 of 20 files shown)


tonyfettes:
Kimi sometimes returns 502 on this PR. Drafting

PR #283, opened 2026-07-02
feat(agent_tool): pre-write moonc syntax gate for write/edit

bobzhang:
## What

Adds a `revert_on_parse_errors` option (on by default) to the `write` and `edit` tools: candidate `.mbt` content is parsed standalone with `moonc compile -no-intermediate-file -stop-after-parsing -error-format json` in a scratch directory **before** anything touches the destination, and content that fails to lex/parse is rejected with the errors and numbered source excerpts.

- **write**: new `.mbt` content with parse errors is rejected before the file (or its parent dirs) is created. Since `write` refuses to overwrite existing source, the natural retry is another `write` with corrected content. At ≥3 parse errors the retry hint switches to "write a smaller skeleton first, grow it with `edit`".
- **edit**: the edited result is parsed before the write; an edit that would *increase* the parse-error count vs the original content (parsed the same way) is rejected with the file untouched. Pre-existing parse errors never block edits, so partial fixes stay possible. Symlinked targets are resolved before gating.
- **Excerpts**: the compiler's own `context` field (populated in the scratch-dir invocation shape) is preferred; a loc-based synthesis from the in-memory content is the fallback. On a reject the content never lands on disk, so these excerpts are the model's only view of the broken spot.
- `revert_on_parse_errors=false` opts out for deliberately non-parsing fixtures.

## Why moonc instead of moon check

`moon check` answers "is the project healthy"; the gate only asks "does this one file parse". The standalone parse is ~25ms, needs no project context (covers writes outside moon projects), reports only lexer/parser diagnostics (type errors stay a legitimate transient state during multi-file work), and — because it validates *before* writing — eliminates the delete/restore/re-apply rollback machinery and its cancellation windows entirely.

## Soundness details

- The gate **never certifies content it did not parse to completion**: moonc missing, timeout, non-zero exit with no diagnostics, or a truncated capture with nothing caught all fail open (write proceeds, as before the gate existed).
- moonc **exits 0 when its parser recovers** from an error (e.g. `&mut` in a parameter type) while still emitting the 3002 diagnostic — the gate trusts captured diagnostics, never exit codes (pinned by a regression test).
- Truncated captures are lower bounds: reject reports say "at least N" / `≥N`, and the edit gate requires a complete baseline before comparing counts.
- `write` commits to `CreateNew` for any path its probe called new, closing both a dangling-symlink `O_CREAT` bypass (`alias.txt -> lib/new.mbt`) and the probe→write source-creation race.

## Not covered (TODO in todo.md + parse_gate.mbt)

- `.mbt.md` is not gated: raw moonc lexes the markdown itself (headings/fences become 3001s), so gating doc files needs moon's fenced-block extraction rules.
- `multi_edit` keeps its existing moon-check error-count revert guard (`revert_when_errors_above`) and is not parse-gated; extending the moonc gate to batches is a natural follow-up.

## Validation

- 873/873 tests green, `moon fmt` + `moon info` clean. New unit coverage: gate decision table, loc/excerpt math, truncation labeling, symlink and dangling-symlink regressions, plus real-moonc round trips (safe under `moon test` — no project build lock involved).
- Each commit was reviewed with the codex CLI; findings (fail-open baseline, symlink bypasses, truncation soundness, a race my own fix re-introduced) are addressed in follow-up commits.
- End-to-end: an uncapped `openseek run` one-shot ("write a TOML parser CLI") finished on its own at 123 steps with `moon check` at 0 errors and 15/15 tests passing; the gate fired 5 times (3 write rejects, 2 edit rejects) and the model recovered from each rejection with a corrected re-issue.


bobzhang:
Pushed `671a943`: the moonc pre-write gate now also covers **multi_edit** batches — each edited `.mbt` file's rendered result is parsed before any disk write, and a batch that would increase any file's parse-error count (vs its pre-batch content parsed the same way) is rejected whole, keeping the tool's all-or-nothing contract. The post-write `revert_when_errors_above` moon-check guard stays for type-error regressions.

With `write`, `edit`, and `multi_edit` all gated, no tool path can land a syntax error in a plain `.mbt` file (escape hatches: gate fail-open, explicit `revert_on_parse_errors=false`, `.mbt.md` TODO, and non-tool writes — the shell sandbox blocks those). Codex review of the commit: no findings. 876/876 tests green.

PR #290, opened 2026-07-02
feat(agent_tool): pitfall hints in parse-gate rejects + missing prompt rules

bobzhang:
## What

Analysis of four recorded one-shot eval sessions (TOML/YAML parsers, regex engine, tiny compiler; ~110 attempted source mutations) showed that **three legacy/Rust-prior syntax classes caused every parse-gate rejection**:

1. Rust-style `mut` parameters/receivers — `fn f(mut pos : Int)`, `fn P::advance(mut self : P)` (3 of 4 runs)
2. Postfix `?` — Rust-style propagation/unwrap (`stack.pop()?`) and related legacy forms
3. Prefix `~label` arguments — `f(~label)` (14 errors in one file)

Notably, the flash prompt **already contains the `mut` rule** and the model violated it in 3 of 4 runs — prompt rules alone don't stick under generation pressure, so the correction has to land at the moment of failure.

### Changes

- **`pitfall_hints()`** in `auto_check/parse_gate.mbt`: maps gate diagnostics to one-line MoonBit idiom corrections (deduped, one per class), appended to the reject reports of `write`, `edit`, and `multi_edit` (the batch report aggregates diagnostics across rejected files). The `?`/`!` hint names all three wrong-prior causes (raise-propagation, Option unwrap, C ternary) rather than guessing which was meant.
- **Flash prompt**: adds the two genuinely missing rules — `label=value`/`label~` vs legacy `~label`, and "there is no `?` operator; `f!(...)` is deprecated" — plus a workspace-root rule (one eval nested its project via `moon new` and then paid a `cd <dir> &&` prefix on 34 of 35 shell commands).

### Verified against the installed toolchain

Wordings were probed against moonc (not docs): `f(...)?` hard-fails with a 3002 parse error, `f!(...)` parses with a deprecation warning, `~label` is a 3001 lexing error, `try?` itself is deprecated. Codex CLI review findings (over-broad `?` hint, wrong legacy-bang form) are addressed in the follow-up commit.

## Validation

877/877 tests green; new unit tests cover class mapping, dedupe, and no-hint-for-unknown-classes. Next step (separate): N=20 A/B bench of this change against baseline on fixed tasks, measuring steps/tokens/rejections.

### DIFF
--- a/agent_tool/edit/edit.mbt
+++ b/agent_tool/edit/edit.mbt
@@ -324,13 +324,15 @@ fn edit_reject_report(
   } else {
     "\nfirst \{formatted.length()} parse error(s), excerpted from the WOULD-BE content (not what is on disk):\n\{formatted.join("\n")}"
   }
+  let hints = @auto_check.pitfall_hints(gate.errors)
+  let hint_lines = if hints.is_empty() { "" } else { "\n\{hints.join("\n")}" }
   let retry = if introduced >= SmallerEditsHintThreshold {
     "\n\{introduced_label} new parse errors usually means the replacement would break the file's structure (an unbalanced brace, a misplaced insertion, or non-MoonBit text). Do not re-issue the same edit: split it into smaller pieces — multi_edit applies several small line-anchored edits in one validated call — and check after each step."
   } else {
     "\nre-issue the edit with a corrected new_string (check delimiters against the excerpt above)."
   }
   let opt_out = "\n(To intentionally produce content that does not parse, e.g. a fixture, pass revert_on_parse_errors=false.)"
-  "\{head}\{pre_existing}\{listed}\{retry}\{opt_out}"
+  "\{head}\{pre_existing}\{listed}\{hint_lines}\{retry}\{opt_out}"
 }
 
 ///|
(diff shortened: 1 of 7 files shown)


bobzhang:
**A/B bench** (N=40: {main @ a87db37, this branch} × {TOML parser, tiny compiler} × 10 seeds, deepseek-v4-pro, sessions recorded, max-steps 250):

| metric (mean, finished runs) | base | treat | Δ | MW z |
|---|---|---|---|---|
| steps | 94 | 82 | −13% | +1.18 |
| cache-miss tokens | 54.0k | 46.4k | −14% | +1.63 |
| completion tokens | 57.5k | 55.0k | −4% | +0.62 |
| finished / green | 19/20, 19 green | 20/20, 19 green | — | — |
| syntax probes | 12.7 | 11.2 | −12% | — |
| nested projects | 2 | 1 | — | — |

- Every aggregate moves in the treatment's favor; none individually crosses p<0.05 at n≈20/arm (cache-miss tokens closest at z=1.63). Confirming a ~13% effect at this variance needs roughly 40–60 seeds/arm; the cleanest cell is the compiler task (79.8 → 66.4 steps).
- Gate rejections did **not** drop (2.5 → 2.9/run): the win is **cheaper recovery, not fewer mistakes** — consistent with the observation that prompt rules alone don't stick. `mut` remains the #1 single diagnostic class (35 across arms); the hints fired on every rejection matching a known class (9/57 — the rest are one-off structural slips that the excerpts already localize, so no further hint classes are warranted).
- Side finding: one baseline run crashed at step 4 with a session-store error — `stale session snapshot for cli-…: no persisted header, in-memory last sequence 8, durable last sequence 0` — a reliability bug independent of this change, worth a separate issue.

bobzhang:
to provide good error message should be done on the tooling side

PR #291, opened 2026-07-02
refactor(agent_tool/edit): de-dup the write-and-report tail into commit_edit

bobzhang:
Obvious de-dup + idiomatic win (repo-wide "obvious wins only" simplification pass).

**De-dup:** the insertion path and the replacement path each ended with an identical ~18-line tail — the `@manifest_error.write_error` guard, the `parse_gate_rejection` guard, and the `write_file`/`append_summary`/`respond` try-catch — differing only in the success `summary` string and ok-`brief`. Extracted a shared private `async fn commit_edit(path, original_content~, new_content~, revert_on_parse_errors~, fail_brief~, summary~, ok_brief~)`; both branches now call it. Behavior and all strings preserved exactly.

**Idiom (inside the helper):** two `match f(...) { Some(x) => return …; None => () }` → `if f(...) is Some(x) { return … }`.

Left alone (not strict wins): the code-unit index loops in `line_*_offset` (converting to `for c in content` risks char-vs-code-unit semantics), and a load-bearing `.map(...)` closure.

## Validation
`moon fmt` clean, `moon check agent_tool/edit` 0 warnings/errors, `moon test agent_tool/edit` 29/29, `moon info` shows **no `pkg.generated.mbti` change** (`commit_edit` is private). Only `edit.mbt` touched.

### DIFF
--- a/agent_tool/edit/edit.mbt
+++ b/agent_tool/edit/edit.mbt
@@ -117,42 +117,15 @@ async fn edit_file(
       max=content.length(),
     )
     let new_content = "\{content.unsafe_substring(start=0, end=at)}\{input.new_string}\{content.unsafe_substring(start=at, end=content.length())}"
-    match @manifest_error.write_error(path, new_content) {
-      Some(message) =>
-        return @agent_tool.ToolAction::respond(
-          "error editing \{path}: \{message}",
-          is_error=true,
-          brief=fail_brief,
-        )
-      None => ()
-    }
-    match
-      parse_gate_rejection(
-        path,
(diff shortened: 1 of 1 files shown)


bobzhang (commented):
Reviewed the diff against `main`. The new private `commit_edit` helper preserves the previous manifest guard, parse-gate guard, write, auto-check summary, success brief, and write-error response for both insert and replace paths; only the duplicated tail moved behind parameters for the path-specific strings. `check (nightly)` is green, so this looks safe to merge.

PR #293, opened 2026-07-02
refactor(tui/internal/composer): de-dup cursor resolution + clamp/comprehension idioms

bobzhang:
Obvious de-dup + idiomatic wins (repo-wide "obvious wins only" pass), net **+26/−61**, behavior-identical.

**De-dup (main win):** six call sites (`cursor_row`, `cursor_at_first/last_visual_row`, `cursor_offset`, `reclamp_view`, `move_vertical`) each repeated the identical 6-line `resolve_cursor(self.logical_lines, rows=…, cursor=…, at_wrap_end=…)` block → extracted a private `Model::resolve_cursor_index()`; all six now call it.

**Idioms:**
- `compute_visible_count`: 3-branch if/else → `total.clamp(min=MinTextRows, max=max_text_rows)`.
- `cursor_row` / `reclamp_view` / `set_logical_lines`: `if x<0/… ` → `.clamp(...)` / `.max(0)`.
- `split_logical_lines`: push loop → `[for display in @displaytext.split_lines(text) => LogicLine::new(display)]`.
- `join_logical_lines`: `for index in 0..<len { … logical_lines[index] }` → `for index, line in logical_lines`.

Left alone (not clear wins): the `insert` builder loop (multi-branch body), `splice_logical_lines` (sub-range copies), boolean guards, and hot-path `is Some`/`.map`. The `<+ "\{…}"` interpolation stays — `<+` requires a template string.

## Validation
`moon fmt` clean, `moon check tui/internal/composer` 0 warnings/errors, `moon test tui/internal/composer` 24/24, `moon info` shows **no `pkg.generated.mbti` change** (`resolve_cursor_index` is private). Only the two `.mbt` files touched.

### DIFF
--- a/tui/internal/composer/composer.mbt
+++ b/tui/internal/composer/composer.mbt
@@ -30,13 +30,7 @@ const DefaultContentWidth : Int = 78
 /// one and `max_text_rows`.
 fn compute_visible_count(total~ : Int, max_text_rows~ : Int) -> Int {
   let max_text_rows = max_text_rows.max(MinTextRows)
-  if total < MinTextRows {
-    MinTextRows
-  } else if total > max_text_rows {
-    max_text_rows
-  } else {
-    total
-  }
+  total.clamp(min=MinTextRows, max=max_text_rows)
 }
 
 ///|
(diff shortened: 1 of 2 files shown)


bobzhang (commented):
Reviewed the diff against `main`. `resolve_cursor_index` centralizes the same `resolve_cursor(...)` call used by the cursor helpers, and the clamp/comprehension rewrites preserve the old bounds and row/text ordering. `check (nightly)` is green, so this looks safe to merge.

PR #294, opened 2026-07-02
refactor(agent_tool/multi_edit): idiomatic is/append/compare + de-dup edit-count

bobzhang:
Obvious idiomatic + de-dup wins (repo-wide "obvious wins only" pass), net **+26/−48**, behavior-identical.

**Idioms:**
- `match args.edits_file { Some(spec) => …push loop…; None => () }` → `if args.edits_file is Some(spec) { … edits.append(file_edits) }` (`is`-pattern + `Array::append` replacing a push loop).
- Three more `match f(...) { Some(x) => return …; None => () }` → `if f(...) is Some(x) { … }` (one folded into an existing `&&` guard, preserving the short-circuit so the gate isn't called when the flag is false).
- `compare_plan_ascending`: `compare_int(a.index, b.index)` → `a.index.compare(b.index)` (built-in `Int::compare`), deleting the now-unused local `compare_int`.
- Two `[ for plan in plans => plan ]` clone-comprehensions → `plans.copy()`.
- `rejections[shown:rejections.length()]` → `rejections[shown:]`.

**De-dup:** three identical `let mut total = 0; for write in writes { total += write.edit_count }` blocks (`revert_report`/`success_summary`/`success_brief`) → private `total_edit_count(writes)`.

Left alone: the `resolved` push-loop (comprehension form was awkward/ambiguous for the struct literal).

## Validation
`moon fmt` clean, `moon check agent_tool/multi_edit` 0 warnings/errors, `moon test agent_tool/multi_edit` 30/30, `moon info` shows **no `pkg.generated.mbti` change**. Only `multi_edit.mbt` touched.

### DIFF
--- a/agent_tool/multi_edit/multi_edit.mbt
+++ b/agent_tool/multi_edit/multi_edit.mbt
@@ -77,16 +77,11 @@ async fn execute_with_workspace(
   }
   // Concatenate the inline edits with the response file's edits (inline first).
   let edits = args.inline
-  match args.edits_file {
-    Some(spec) =>
-      match load_edits_file(workspace_root, spec) {
-        Ok(file_edits) =>
-          for edit in file_edits {
-            edits.push(edit)
-          }
-        Err(action) => return action
-      }
-    None => ()
+  if args.edits_file is Some(spec) {
+    match load_edits_file(workspace_root, spec) {
+      Ok(file_edits) => edits.append(file_edits)
+      Err(action) => return action
+    }
   }
   if edits.is_empty() {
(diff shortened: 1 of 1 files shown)


bobzhang (commented):
Reviewed the diff against `main`. The edits-file append keeps inline edits first, the parse-gate rewrite preserves the `revert_on_parse_errors` short-circuit, `plans.copy()` replaces clone comprehensions, and `total_edit_count` is the same accumulated edit count used by the summaries. `check (nightly)` is green, so this looks safe to merge.

PR #295, opened 2026-07-02
refactor(agent_runtime): de-dup the queue drain loops into a generic helper

bobzhang:
Obvious de-dup win (repo-wide "obvious wins only" pass).

`drain_steers` and `drain_events` were near-identical 5-line drain loops differing only in the queue field/element type. Extracted a shared private generic `fn[T] drain_queue(queue : @aqueue.Queue[T]) -> Array[T]`; both public methods now delegate. Same `try_get() catch { _ => None }`-until-empty loop, same oldest-first ordering. Helper is private → no public API change.

Left alone (not clear wins): the one-liner `emit_event`/`queue_steer` (extracting a helper is marginal); the `while … is Some(x)` drains are already idiomatic and can't be comprehensions.

## Validation
`moon fmt` clean, `moon check agent_runtime` 0 warnings/errors, `moon test agent_runtime` 8/8, `moon info` shows **no `pkg.generated.mbti` change**. Only `runtime.mbt` touched.

### DIFF
--- a/agent_runtime/runtime.mbt
+++ b/agent_runtime/runtime.mbt
@@ -168,11 +168,7 @@ pub fn AgentRuntime::queue_steer(
 /// this call. This method does not trim or filter; consumers decide how to treat
 /// blank steering input.
 pub fn AgentRuntime::drain_steers(self : AgentRuntime) -> Array[SteerInput] {
-  let inputs : Array[SteerInput] = []
-  while (self.steers.try_get() catch { _ => None }) is Some(input) {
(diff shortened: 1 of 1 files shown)


bobzhang (commented):
Reviewed the diff against `main`. The generic `drain_queue` keeps the same `try_get() catch { _ => None }` loop and returns items in the same oldest-first order for both steer and event queues. `check (nightly)` is green, so this looks safe to merge.

PR #298, opened 2026-07-02
refactor(agent_tool/shell): unwrap_or + for-index loops + de-dup mv/cp operand split

bobzhang:
Obvious idiomatic + de-dup wins (repo-wide "obvious wins only" pass), net **+7/−31**, behavior-identical, no public API change:

- **shell.mbt** (×2): `match input.timeout_ms { Some(x) => x; None => 0 }` / `match output.exit_code { … }` → `.unwrap_or(0)` (constant default, no closure).
- **sandbox.mbt** `script_generated_tree_transfer_literal_pair_target`: nested `for source_index in 0..<literals.length() { for dest_index in 0..<… }` → `for source_index, source_literal in literals { for dest_index, dest_literal in literals … }` (indices still bound for the `!=` guard).
- **sandbox.mbt** (de-dup) `parse_mv_operands`: its trailing `match target_directory { … }` block was byte-for-byte identical to the private helper `operands_with_optional_dest` → replaced with a call to it, mirroring what `parse_cp_operands` already does (removes ~13 duplicated lines).

Left alone (debatable): `is_fd_redirect_target` vs `command_redirect_target_is_fd` (byte-identical but name two distinct scanner layers), and the char-index parser loops (code-unit vs char semantics).

## Validation
`moon fmt` clean, `moon check agent_tool/shell` 0 warnings/errors, `moon test agent_tool/shell` 106/106, `moon info` shows **no `pkg.generated.mbti` change**. Only `shell.mbt`/`sandbox.mbt` touched.

### DIFF
--- a/agent_tool/shell/sandbox.mbt
+++ b/agent_tool/shell/sandbox.mbt
@@ -3956,17 +3956,11 @@ async fn script_generated_tree_transfer_literal_pair_target(
   cwd : String,
 ) -> String? {
   let literals = script_quoted_literals(text)
-  for source_index in 0..<literals.length() {
-    for dest_index in 0..<literals.length() {
+  for source_index, source_literal in literals {
+    for dest_index, dest_literal in literals {
       if source_index != dest_index {
-        let source = resolve_source_write_redirect_target(
-          cwd,
-          literals[source_index],
-        )
-        let dest = resolve_source_write_redirect_target(
-          cwd,
-          literals[dest_index],
-        )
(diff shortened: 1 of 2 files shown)


bobzhang (commented):
Reviewed the diff against `main`. The `unwrap_or(0)` changes preserve the previous defaults, the literal-pair loop still tests every distinct source/destination pair, and `parse_mv_operands` now uses the same tail helper already used by `cp` with identical operand/destination behavior. `check (nightly)` is green, so this looks safe to merge.

PR #299, opened 2026-07-02
refactor(tui/internal/viewport): de-dup repaint loops + @cmp clamps

bobzhang:
Obvious de-dup + idiomatic wins (repo-wide "obvious wins only" pass), behavior-identical, helper is private.

**De-dup:** three functions (`enter_placed_surface_rows`, `redraw_uncached_placed_surface_rows`, `replay_scrollback`) contained the identical 3-line "draw every terminal row in a range straight from a placed frame, no diffing" loop, verbatim modulo range bounds → extracted a private async `draw_frame_rows(tty, frame, top~, bottom~)`; all three call it. Names the shared full-repaint op, contrasting the diffing path. The blank-row clearing in `replay_scrollback` was left intact.

**Idioms:** `shift_down_for_insert` `if space<0 { 0 } else { space }` → `@cmp.maximum(space, 0)` and `if lines<available …` → `@cmp.minimum(lines, available)` (same `@cmp` house style already used elsewhere in the file).

Left alone (debatable): the `is Some(size)` message (already idiomatic), the distinct try/catch/finally restore patterns, and the dynamic-split index loops.

## Validation
`moon fmt` clean, `moon check tui/internal/viewport` 0 warnings/errors, `moon test tui/internal/viewport` 5/5, `moon info` shows **no `pkg.generated.mbti` change**. Only `viewport.mbt` touched.

### DIFF
--- a/tui/internal/viewport/viewport.mbt
+++ b/tui/internal/viewport/viewport.mbt
@@ -347,6 +347,21 @@ async fn draw_surface_line_at(
 }
 
 ///|
+/// Repaint every terminal row in `top..=bottom` straight from `frame`, with no
+/// diffing. Rows past the frame's own height resolve to `None` and are cleared.
+/// The caller controls auto-wrap around the batch.
+async fn draw_frame_rows(
+  tty : @tty.Tty,
+  frame : PlacedSurface,
+  top~ : Int,
+  bottom~ : Int,
+) -> Unit {
+  for row in top..<=bottom {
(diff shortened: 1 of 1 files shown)


bobzhang (commented):
Reviewed the diff against `main`. `draw_frame_rows` is the same full-row repaint loop reused at the three prior call sites, and the `@cmp.maximum`/`@cmp.minimum` changes preserve the previous clamp behavior in `shift_down_for_insert`. I also checked that an empty inclusive range does not iterate. `check (nightly)` is green, so this looks safe to merge.

PR #303, opened 2026-07-02
refactor(agent_tool/write): is-patterns + unwrap_or sentinel in parent_dir

bobzhang:
Obvious idiomatic wins (repo-wide "obvious wins only" pass), net **−8**, behavior-identical:

- `match manifest_write_error(...) { Some(message) => return …; None => () }` → `if manifest_write_error(...) is Some(message) { return … }`.
- parse gate: `match gate { Some(gate) if gate.errors.length() > 0 => {…}; _ => () }` → `if gate is Some(gate) && gate.errors.length() > 0 { … }` (the `is … &&` form is already used in `internal/decode`, `internal/auto_check`).
- `parent_dir`: the 4-arm cartesian `match (rev_find("/"), rev_find("\\"))` → `slash = rev_find("/").unwrap_or(-1)`, `backslash = …unwrap_or(-1)`, `if slash > backslash { slash } else { backslash }`. The `-1` default reproduces every arm exactly (both `rev_find`s still run, identical performance).

De-dup: none clear (wrapping a one-line interpolated brief in a helper is debatable → left).

## Validation
`moon fmt` clean, `moon check agent_tool/write` 0 warnings/errors, `moon test agent_tool/write` 26/26, `moon info` shows **no `pkg.generated.mbti` change**. Only `write.mbt` touched.

### DIFF
--- a/agent_tool/write/write.mbt
+++ b/agent_tool/write/write.mbt
@@ -78,14 +78,12 @@ async fn write_file(
   path~ : String,
 ) -> @agent_tool.ToolAction {
   try {
-    match manifest_write_error(path, input.content) {
-      Some(message) =>
-        return @agent_tool.ToolAction::respond(
-          "error writing \{path}: \{message}",
-          is_error=true,
-          brief="write \{@agent_tool.brief_basename(path)}",
-        )
-      None => ()
+    if manifest_write_error(path, input.content) is Some(message) {
+      return @agent_tool.ToolAction::respond(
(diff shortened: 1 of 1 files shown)


bobzhang (commented):
Reviewed the diff against `main`. The manifest and parse-gate guards keep the same early-return behavior, including the parse-error count in the brief, and the `parent_dir` sentinel rewrite reproduces all previous slash/backslash cases. `check (nightly)` is green, so this looks safe to merge.

PR #304, opened 2026-07-02
refactor(deepseek): de-dup Usage::from_json field lookups

bobzhang:

`Usage::from_json` had 5 near-identical inline `match json { { "<key>": Number(n, ..), .. } => n.to_int(); _ => 0 }` blocks differing only by the field key → extracted a private `json_object_int(json, key)`; the body is now 5 one-line calls. `Json::Object` + `Map::get` reproduces the record-pattern lookup exactly (missing / non-number → 0 in both).

Left alone (intentional / not clear wins): `ToolCall::from_json`'s error-path taxonomy (verbose by design), the `thinking` High/Max match, and the `ToolDefinition::to_json` conditional-field if/else. Rest of the package is already idiomatic.

## Validation
`moon fmt` clean, `moon check deepseek` 0 warnings/errors, `moon test deepseek` 37/37, `moon info` shows **no `pkg.generated.mbti` change**. Only `json_decode.mbt` touched.

### DIFF
--- a/deepseek/json_decode.mbt
+++ b/deepseek/json_decode.mbt
@@ -1,26 +1,25 @@
 ///|
+/// Reads an integer-valued field from a JSON object, returning 0 when the field
+/// is absent or is not a number.
+fn json_object_int(json : Json, key : String) -> Int {
+  match json {
+    Object(fields) =>
+      match fields.get(key) {
+        Some(Number(n, ..)) => n.to_int()
+        _ => 0
+      }
+    _ => 0
+  }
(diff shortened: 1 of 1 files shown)


bobzhang (commented):
Reviewed the diff against `main`. `json_object_int` preserves the previous `Usage::from_json` field semantics: object number fields decode via `to_int`, while missing, non-number, or non-object input defaults to 0. `check (nightly)` is green, so this looks safe to merge.

PR #309, opened 2026-07-02
refactor(agent_session/store): merge error arms + de-dup sort comparator

bobzhang:
Obvious idiomatic + de-dup wins (repo-wide "obvious wins only" pass), net **−5**, behavior-identical, all private/internal:

- `header_string_field`/`header_int_field`: two byte-identical error arms `Some(_) => …` / `None => …` → a single `_ =>` (the `Some(String(value))`/`Some(Number(...))` success arm still matches first).
- **De-dup:** `listings` and `latest` had two byte-identical `sort_by` comparator closures (destructure `(_, seconds, nanoseconds)`, newest-first) → extracted a generic private `fn[T] stamped_newest_first(a, b)`; both call `stamped.sort_by(stamped_newest_first)`. The tuples differ only in the ignored position-0 type, so `fn[T]` covers both.

Left alone (debatable): `if @fs.exists …` negation, the `@vector.Vector` counter loop, and tuple-compare collapse.

## Validation
`moon fmt` clean, `moon check agent_session/store` 0 warnings/errors, `moon test agent_session/store` 27/27, `moon info` shows **no `pkg.generated.mbti` change**. Only `store.mbt` touched.

### DIFF
--- a/agent_session/store/store.mbt
+++ b/agent_session/store/store.mbt
@@ -146,8 +146,7 @@ fn header_string_field(
 ) -> String raise @json.JsonDecodeError {
   match fields.get(key) {
     Some(String(value)) => value
-    Some(_) => json_decode_error(path.add_key(key), "expected string")
-    None => json_decode_error(path.add_key(key), "expected string")
+    _ => json_decode_error(path.add_key(key), "expected string")
   }
 }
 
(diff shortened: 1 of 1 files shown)


bobzhang (commented):
Reviewed the diff against `main`. Merging the header decode error arms preserves the same path/error text for both absent and wrong-typed fields, and the generic `stamped_newest_first` comparator is the same newest-first `(seconds, nanoseconds)` ordering used by both callers. `check (nightly)` is green, so this looks safe to merge.

PR #312, opened 2026-07-02
refactor(agent_tool/finish): de-dup finish test assertions

bobzhang:
Obvious de-dup win (repo-wide "obvious wins only" pass), test-meaning preserved:

The 6 test blocks each repeated the identical 3-line `let result = execute(...)` / `guard result is Control(Finish(answer)) else { fail("expected Finish") }` / `assert_eq(answer, …)` → extracted a private `assert_finish(arguments, expected)`; each test is now a single call. Same guard, same `fail` message, same `assert_eq` — identical assertions.

Left alone: non-test code (`definition`/`execute`/`finish`) is already idiomatic; did NOT inline the doc-referenced `finish` decode layer (debatable).

## Validation
`moon fmt` clean, `moon check agent_tool/finish` 0 warnings/errors, `moon test agent_tool/finish` 8/8, `moon info` shows **no `pkg.generated.mbti` change** (helper is private). Only `finish.mbt` touched.

### DIFF
--- a/agent_tool/finish/finish.mbt
+++ b/agent_tool/finish/finish.mbt
@@ -35,43 +35,40 @@ fn finish(input : @decode.FinishInput) -> @agent_tool.ToolAction {
 }
 
 ///|
+/// Asserts that `execute(arguments)` yields `Control(Finish(expected))`.
+fn assert_finish(arguments : Json, expected : String) -> Unit raise {
+  guard execute(arguments) is Control(Finish(answer)) else {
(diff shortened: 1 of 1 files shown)


bobzhang (commented):
Reviewed the diff against `main`. The private `assert_finish` helper keeps the same `execute(...)` call, guard pattern, failure message, and `assert_eq` assertion for each test case, so this is test de-duplication only. `check (nightly)` is green, so this looks safe to merge.

PR #313, opened 2026-07-02
refactor(tui/internal/history): flatten push guard, de-dup the push call

bobzhang:
Obvious de-dup + idiomatic win (repo-wide "obvious wins only" pass), behavior-identical:

`History::push`: nested `if entry.text.length()>0 { if last is Some(last) { if !same → push } else { push } }` (with `self.entries.push(entry)` duplicated across two branches) → a single flat guard `if entry.text.length()>0 && !(self.entries.last() is Some(last) && last.same_as(entry)) { push }`. Pushes iff text non-empty AND not a consecutive duplicate (verified all four cases); `&&` short-circuits so `last()`/`same_as` are still skipped for empty text.

Left alone (not strict wins): `match self.entries.last()` would need a no-op skip arm; the `previous`/`next` early returns are genuine guard clauses.

## Validation
`moon fmt` clean, `moon check tui/internal/history` 0 warnings/errors, `moon test tui/internal/history` 3/3, `moon info` shows **no `pkg.generated.mbti` change**. Only `history.mbt` touched.

### DIFF
--- a/tui/internal/history/history.mbt
+++ b/tui/internal/history/history.mbt
@@ -65,14 +65,9 @@ pub fn History::reset_navigation(self : History) -> Unit {
 /// Append `entry` (unless empty or a duplicate of the most recent one) and reset
 /// navigation to the fresh draft.
 pub fn History::push(self : History, entry : Entry) -> Unit {
-  if entry.text.length() > 0 {
-    if self.entries.last() is Some(last) {
-      if !last.same_as(entry) {
-        self.entries.push(entry)
-      }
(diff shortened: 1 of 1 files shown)


bobzhang (commented):
Reviewed the diff against `main`. The flattened guard in `History::push` preserves the previous behavior: empty entries do not push, consecutive duplicates do not push, non-empty first/different entries do push, and navigation is still reset afterward. `check (nightly)` is green, so this looks safe to merge.

PR #317, opened 2026-07-02
refactor(command_context): unwrap_or + drop single-use bindings

bobzhang:
Obvious idiomatic wins (repo-wide "obvious wins only" pass), behavior-identical:

- `shell_result_text`: `match result.exit_code { Some(code) => code; None => 0 }` → `result.exit_code.unwrap_or(0)` (literal default; matches house style in `cmd/tui/loop.mbt`).
- `encode` and `shell_result_text`: removed the redundant `let text = <$| multiline>; text` bindings (name used once, immediately) — return the multiline literal directly.

Left alone: `exit_header`'s `if exit_code is Some(code) … else …` (the `is`-pattern if/else is already idiomatic); no de-dup opportunities.

## Validation
`moon fmt` clean, `moon check cmd/tui/internal/command_context` 0 warnings/errors, `moon test cmd/tui/internal/command_context` 5/5, `moon info` shows **no `pkg.generated.mbti` change**. Only `command_context.mbt` touched.

### DIFF
--- a/cmd/tui/internal/command_context/command_context.mbt
+++ b/cmd/tui/internal/command_context/command_context.mbt
@@ -49,7 +49,7 @@ pub fn encode(
     Ok(result) => shell_result_text(result)
     Err(error) => "error running shell: \{error}"
   }
-  let text =
+  (
     $|<user_shell_command>
     $|<command>
     $|\{command}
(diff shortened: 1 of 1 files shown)


bobzhang (commented):
Reviewed the diff against `main`. `unwrap_or(0)` preserves the previous missing-exit-code default, and the removed single-use `text` bindings just return the same multiline literals directly. `check (nightly)` is green, so this looks safe to merge.

PR #319, opened 2026-07-02
refactor(testkit/filesystem): String::repeat + de-dup object_fields unwrap

bobzhang:
Obvious idiomatic + de-dup wins (repo-wide "obvious wins only" pass), behavior-identical:

- **outline.mbt** `outline_indent`: `StringBuilder` + `for _ in 0..<depth { sb.write_string("  ") }` → `"  ".repeat(depth)` (`repeat(0)` → `""`, matching a zero-iteration loop).
- **filesystem.mbt** (de-dup): the verbatim `let FileSystem(json) = self; guard json is Object(fields) else { fail("file system must be a JSON object") }` appeared in both `entries` and `get` → private `FileSystem::object_fields(self)`; both call `self.object_fields()`. Removes the duplicated unwrap and error-message literal.

Left alone (not identical behavior): `sanitize_name`'s manual counter (it's a code-unit offset used for slicing, not a char count), and `from_pairs`'s tuple binding.

## Validation
`moon fmt` clean, `moon check testkit/filesystem` 0 warnings/errors, `moon test testkit/filesystem` 17/17, `moon info` shows **no `pkg.generated.mbti` change** (helper is private). Only the two `.mbt` files touched.

### DIFF
--- a/testkit/filesystem/filesystem.mbt
+++ b/testkit/filesystem/filesystem.mbt
@@ -50,6 +50,17 @@ pub fn FileSystem::to_json(self : FileSystem) -> Json {
 }
 
 ///|
+/// The underlying JSON object's fields, raising when the top-level JSON is not
+/// an object as the convention requires.
+fn FileSystem::object_fields(self : FileSystem) -> Map[String, Json] raise {
+  let FileSystem(json) = self
+  guard json is Object(fields) else {
+    fail("file system must be a JSON object")
+  }
+  fields
+}
+
+///|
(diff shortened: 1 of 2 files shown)


bobzhang (commented):
Reviewed the diff against `main`. `FileSystem::object_fields` preserves the previous top-level object validation/error and removes the duplicate unwrap; `"  ".repeat(depth)` matches the old zero-or-more loop for indentation. `check (nightly)` is green, so this looks safe to merge.

PR #320, opened 2026-07-02
refactor(auto_check): guard-is unwrap + de-dup split_owned

bobzhang:
Obvious idiomatic + de-dup wins (repo-wide "obvious wins only" pass), behavior-identical, all private/internal:

- **auto_check.mbt** (×2, `auto_check_summary`/`count_errors`): `let cwd = match moon_check_cwd(path) { Some(cwd) => cwd; None => return None }` → `guard moon_check_cwd(path) is Some(cwd) else { return None }`.
- **parse_gate.mbt** (de-dup): three identical `X.split(sep).map(x => x.to_owned()).collect()` blocks (`format_parse_gate_errors`/`indent_context`/`loc_line_span`) → private `split_owned(text, sep)`.

Left alone (fails "identical performance"): a JSON-lines scaffold helper (would add a per-line closure/box on a hot path), a `matches_manifest` helper (would add interpolation where literals are used), and borderline stylistic tweaks.

## Validation
`moon fmt` clean, `moon check agent_tool/internal/auto_check` 0 warnings/errors, `moon test agent_tool/internal/auto_check` 23/23, `moon info` shows **no `pkg.generated.mbti` change**. Only the two `.mbt` files touched.

### DIFF
--- a/agent_tool/internal/auto_check/auto_check.mbt
+++ b/agent_tool/internal/auto_check/auto_check.mbt
@@ -25,10 +25,7 @@ pub async fn append_summary(path : String, content : String) -> String {
 
 ///|
 async fn auto_check_summary(path : String) -> String? {
-  let cwd = match moon_check_cwd(path) {
-    Some(cwd) => cwd
-    None => return None
-  }
+  guard moon_check_cwd(path) is Some(cwd) else { return None }
   let result = @async.with_timeout_opt(MoonCheckTimeoutMs, () => {
     @shell.collect_shell_output(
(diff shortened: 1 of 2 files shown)


bobzhang (commented):
Reviewed the diff against `main`. The `guard ... is Some(cwd)` rewrites preserve the previous early-`None` behavior, and `split_owned` centralizes the repeated `split(...).map(...to_owned()).collect()` pattern without changing line splitting or context rendering. `check (nightly)` is green, so this looks safe to merge.

PR #328, opened 2026-07-02
refactor(moon_cmd/internal/decode): de-dup parse_string_list + is-pattern

bobzhang:
Obvious de-dup + idiomatic wins (repo-wide "obvious wins only" pass), net **−11**, behavior-identical:

- De-dup: `parse_arg_list` and `parse_paths` were byte-identical modulo the field-name strings (`"arg"/"args"` vs `"path"/"paths"`) → private `parse_string_list(fields, singular, plural)` (error message derived via `"arguments.\{singular} to be a string"`). Call sites become `parse_string_list(fields, "arg", "args")` / `("path", "paths")`.
- `parse_fields` target: `match target { Some(value) => {…}; None => () }` → `if target is Some(value) { … }` (mirrors the file's own `if stdin is Some(_) && …` idiom).

Left alone (debatable / perf): `validate_test_update_guard`'s `reason` match (keeps parallelism with the adjacent 3-arm match), `uses_test_update`'s loop (`.contains` changes iteration passes), and plain-Bool conditionals.

## Validation
`moon fmt` clean, `moon check agent_tool/moon_cmd/internal/decode` 0 warnings/errors, `moon test` 11/11, `moon info` shows **no `pkg.generated.mbti` change**. Only `decode.mbt` touched.

### DIFF
--- a/agent_tool/moon_cmd/internal/decode/decode.mbt
+++ b/agent_tool/moon_cmd/internal/decode/decode.mbt
@@ -48,15 +48,12 @@ fn parse_fields(fields : Map[String, Json]) -> MoonCommandInput raise {
   validate_command(command)
   let cwd = optional_string(fields, "cwd")
   let target = optional_string(fields, "target")
-  match target {
-    Some(value) => {
-      @moon_target.validate_target(value)
-      validate_target_supported(command)
-    }
-    None => ()
+  if target is Some(value) {
+    @moon_target.validate_target(value)
+    validate_target_supported(command)
   }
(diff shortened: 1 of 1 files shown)


bobzhang (commented):
Reviewed the diff against `main`. The new private `parse_string_list` preserves the previous singular-field handling, singular-before-plural ordering, empty/null skipping, and field-specific error text; the `target is Some(value)` rewrite is equivalent to the previous `match`. `check (nightly)` is green, so this looks safe to merge.

PR #339, opened 2026-07-02
perf(agent_session): O(1) stale-snapshot fast path for append

bobzhang:
Second of four scoped store-hardening PRs (#336 merged). Rebased onto main; independent, self-contained diff.

## Problem

`ensure_session_is_current` read and JSON-parsed the **entire session file**, then re-serialized both event vectors to compare — on every `append` and `compact`. That's O(file) per event and quadratic over a session's life; with tool-result events in the tens-of-kilobytes range, a long session pays real I/O and parse cost on every single step.

## Fix

The store remembers, per session id, the exact `Session` value it last loaded from or persisted to disk, together with the file's size+mtime fingerprint at that moment. When the caller passes that very value back — the common agent loop, where each append's return feeds the next call — and the fingerprint is unchanged, the durable state cannot have diverged, so the full comparison is skipped: the stale check drops from O(file) to a stat.

Correctness is unchanged, only proven differently on the fast path:

- physical identity has no false positives — any snapshot the store did not hand out takes the full compare (the existing divergent-snapshot test still passes through that path)
- a cross-process append grows the file → size breaks the fingerprint (test: "another store instance's write defeats the fingerprint fast path")
- a same-size rewrite is caught by mtime (test added at Codex's suggestion)
- any stat failure falls back to the full compare
- torn files never enter the cache, so the repair path from #336 keeps taking the full read

## Validation

- 43/43 store tests (3 new: chained appends, cross-instance write, same-size rewrite); agent_session suite green
- Codex CLI review with adversarial prompts on mark lifecycle, fingerprint collisions, shared-lock recording, and tail-repair interplay: "Findings: None." — its one hardening suggestion (the same-size-rewrite test) is included

### DIFF
--- a/agent_session/store/README.mbt.md
+++ b/agent_session/store/README.mbt.md
@@ -309,6 +309,10 @@ maintain:
   discarding the uncommitted tail. Corruption anywhere else still fails the
   load.
 - Every append or compact checks that the caller's session still matches disk.
+  When the snapshot is physically the value this store instance last synced
+  and the file's size+mtime+ctime fingerprint is unchanged, that check is
+  O(1); any other snapshot — or any on-disk change — falls back to a full
+  read-and-compare, so the guarantee is the same either way.
 
 These rules keep persistence concerns out of `agent_session.Session` while
 giving CLI, TUI, serve mode, and visualization code one shared durable boundary.
--- a/agent_session/store/store.mbt
+++ b/agent_session/store/store.mbt
@@ -10,14 +10,42 @@ let lock_file_name : String = "session.lock"
 ///|
 /// Native filesystem-backed collection of durable OpenSeek sessions.
 ///
-/// A `SessionStore` owns only a root path string. Under that root, each session
-/// is stored in `sessions/<id>/` with two files: `openseek_session.jsonl` (a
-/// header line followed by append-only event lines) and `session.lock`. The
-/// store is not a handle to one current session; APIs that operate on a session
-/// either take a `SessionId` or a `Session` carrying its id.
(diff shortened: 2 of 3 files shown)


bobzhang:
## Codex CLI final review

_codex-cli 0.141.0 (gpt-5.5, reasoning xhigh), read-only sandbox, reviewing `origin/main...store-append-fast-path` after the rebase onto main. An earlier final-gate pass requested changes over a same-size-rewrite hole in the fingerprint; the commit was amended (size+mtime → size+mtime+ctime) and re-reviewed below. Posted verbatim._

**Verdict: approve.**
- Checked `git diff origin/main...HEAD`: one commit, scoped to `agent_session/store`; no `.mbti` API diff and `git diff --check` is clean.
- The size+mtime+ctime fingerprint closes the prior same-size rewrite hole: `FileFingerprint` includes ctime at `agent_session/store/store.mbt:42`, and the fast path only skips parsing when the same physical session and unchanged fingerprint both hold at `agent_session/store/store.mbt:295`.
- Marks are only recorded after synced clean state: create/load/append/compact at `agent_session/store/store.mbt:655`, `agent_session/store/store.mbt:692`, `agent_session/store/store.mbt:728`, `agent_session/store/store.mbt:766`; torn-tail loads intentionally avoid marking.
- Tests cover chained fast-path appends, cross-store append divergence, and same-size rewrite detection at `agent_session/store/store_test.mbt:269`, `agent_session/store/store_test.mbt:294`, `agent_session/store/store_test.mbt:319`.
- No blocking findings. I could not run `moon test agent_session/store` or `moon check agent_session/store` because the read-only sandbox cannot lock `.mooncakes`.

PR #341, opened 2026-07-02
fix(agent_session): make session creation durable across power loss

bobzhang:
Third of four scoped store-hardening PRs (#336 merged). Rebased onto main; independent, self-contained diff. Note: touches the same functions as #339, so whichever of the two merges second may need a trivial rebase.

## Problem

`write_text_atomic` (used by `create`, first-append bootstrap, and tail repair) renamed the temp file into place but never synced any directory. After the call returned, a power loss could forget the rename — the file's data blocks were durable but its directory entry was not, and the **whole session vanished** despite the store having reported success. Brand-new session directories had the same hole one level up (`sessions/`, store root). Event appends were already effectively `fdatasync`'d because `write_file` defaults to `sync=Data`.

## Fix

- The temp file is flushed (`sync=Data`) before the rename, so a rename can never install a file with unwritten data blocks.
- Every file-creating write syncs the directory chain — session dir, `sessions/`, root, innermost first — **itself**, rather than trusting whichever process happened to `mkdir` those directories moments earlier. (Codex's review caught that syncing at `mkdir` time leaves a race: of two concurrent first-writers, the one that skipped creation could return "durable" while the entries above it were still unsynced.)
- The per-append hot path is untouched: `sync=Data` is now stated explicitly (pinning today's default as policy), and no directory sync runs there.
- Guarantee is scoped honestly: POSIX-style local filesystems, best-effort directory syncs (Windows degrades to the filesystem's own guarantees; process-crash safety unaffected). Two documented scope limits: the store root's own entry belongs to the caller's environment, and appends assume the file's creation was already durably completed.

## Validation

- 43/43 store tests; agent_session suite green (behavioral surface unchanged — this PR changes when data hits the platters, not what is written)
- Directory-fd `File::sync` verified working on macOS via a live probe before adopting the approach
- Codex CLI review, three passes: pass 1 found the missing parent-directory syncs and the overclaiming README; pass 2 found the concurrent-first-create race in the mkdir-time approach (redesigned to writer-owned chain sync); pass 3 confirmed the race closed and the hot path clean, with only the two documentation caveats above, now included

### DIFF
--- a/agent_session/store/README.mbt.md
+++ b/agent_session/store/README.mbt.md
@@ -301,6 +301,20 @@ maintain:
   atomically, so a session file can never exist without its header.
 - `create` rewrites the file atomically (temp file, then rename); an
   interrupted rewrite leaves the previous file intact.
+- Durability policy: event appends and temp-file writes are flushed with
+  `fdatasync` before they return, and every file-creating write (create,
+  first append, tail repair) syncs the directory chain — session directory,
+  `sessions/`, root — itself, rather than trusting whichever process created
+  those directories. On POSIX-style local filesystems this makes a session
+  that `create` or `append` just reported durable survive a power loss, not
+  only a process crash. The directory syncs are best-effort: on platforms
+  where a directory cannot be opened or synced (notably Windows), power-loss
+  durability degrades to the filesystem's own guarantees, while
+  process-crash safety is unaffected. Two scope limits: the store root's own
+  directory entry lives in its parent, outside the store's ownership — when
+  the root is created lazily by the store, making the root itself durable is
+  the caller's setup concern; and an append to an existing file assumes that
+  file's creation was already durably completed by its creator.
 - Every load checks the header record and contiguous sequence numbers.
 - A crash can tear the final append mid-line. `load` tolerates exactly that —
   an unterminated final record is dropped (or kept when only its newline is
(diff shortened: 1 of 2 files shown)


bobzhang:
## Codex CLI final review

_codex-cli 0.141.0 (gpt-5.5, reasoning xhigh), read-only sandbox, reviewing `origin/main...store-rename-durability` after the rebase onto main (previously reviewed through three passes on its stacked base). Posted verbatim._

**Verdict: approve.**

- Rebase is self-contained: `HEAD^ == origin/main`, one commit, changed paths only `agent_session/store/store.mbt` and `agent_session/store/README.mbt.md`.
- Dropped fast-path sibling is absent: no `record_mark`, `fingerprint(s)`, or standalone `marks` references under `agent_session/store`.
- Append hot path still avoids directory sync: existing-file branch remains `@fs.write_file(... append=true, sync=Data)` at `agent_session/store/store.mbt:257`; `sync_dir_best_effort` is only reached through `write_text_atomic` at `agent_session/store/store.mbt:299`.
- File-creating/rewrite paths are covered: first append/tail repair and `create` pass `session_dir_chain` into `write_text_atomic` at `agent_session/store/store.mbt:270` and `agent_session/store/store.mbt:612`; chain order and best-effort swallowing are at `agent_session/store/store.mbt:309` and `agent_session/store/store.mbt:317`.
- README scope matches implementation: POSIX-style local FS guarantee and documented limits at `agent_session/store/README.mbt.md:304`.

No findings. Static checks: `git diff --check` clean; `pkg.generated.mbti` unchanged. Did not run `moon test` because the sandbox is read-only.

PR #342, opened 2026-07-02
fix(agent_session): name unsupported session-file versions; reject NUL ids

bobzhang:
Fourth and last of the scoped store-hardening PRs (#336 merged). Rebased onto main; independent, self-contained diff.

## Problem

A session file written by a future format version failed to load as "does not start with a header record" — misleading exactly when the message matters most, on a version skew between builds. `parse_header` collapsed every non-v1 shape into `None`, so the store couldn't tell "not a header" from "a header this build cannot read". Separately, session ids allowed NUL characters, which truncate the path at the C boundary and would silently map distinct ids onto one directory.

## Fix

- `agent_session/log` gains `classify_header : Json -> HeaderParse { Header, UnsupportedVersion(Int), NotHeader }`; `parse_header` is now a thin wrapper (API additive, no consumer breaks — verified against viz and viz_app).
- Any object with a numeric `version` other than 1 classifies as `UnsupportedVersion` — safe against false positives because event records carry `sequence`/`ts`/`item` and never `version`.
- The store's load error now reads: `uses unsupported session file version 2; this build reads version 1`.
- The tolerant viz parser is unchanged: an unsupported-version first line stays a recorded bad line rather than being adopted as a header.
- `validate_session_id` rejects NUL.

## Validation

- 51/51 log+store tests (new: classification matrix incl. malformed-v1 and event-shaped objects; unsupported-version load wording; NUL rejection); JS consumers (viz, viz_app) 78/78; full suite 903/907 (4 = known env-specific shell timing flakes); cram 23/23
- Codex CLI review across false-positive risk, API compatibility, and UTF-16 NUL semantics: "No findings."

### DIFF
--- a/agent_session/log/log.mbt
+++ b/agent_session/log/log.mbt
@@ -27,21 +27,48 @@ pub impl ToJson for SessionFileHeader with fn to_json(
 }
 
 ///|
-/// Decode a session-file header record, or `None` when `json` is not a
-/// version-1 header object. Header and event records are structurally
-/// disjoint (events carry `sequence`/`item`, never `version`), so this is
-/// also how a reader tells the two apart.
-pub fn parse_header(json : Json) -> SessionFileHeader? {
-  guard json is Object(fields) else { return None }
-  guard fields.get("version") is Some(Number(version, ..)) &&
-    version.to_int() == session_file_version else {
-    return None
+/// How a would-be header record classified. `UnsupportedVersion` is kept
+/// distinct from `NotHeader` so a reader confronted with a file written by a
+/// newer format can say so, instead of misreporting it as a file that has no
+/// header at all.
+pub(all) enum HeaderParse {
+  Header(SessionFileHeader)
+  UnsupportedVersion(Int)
+  NotHeader
+} derive(Debug)
+
+///|
+/// Classify `json` as a session-file header record.
+///
+/// Header and event records are structurally disjoint (events carry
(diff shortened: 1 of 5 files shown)


bobzhang:
## Codex CLI final review

_codex-cli 0.141.0 (gpt-5.5, reasoning xhigh), read-only sandbox, reviewing `origin/main...store-input-hardening` after the rebase onto main (previously reviewed clean on its stacked base). Posted verbatim._

**Verdict: no blockers.**

- Checked `origin/main..HEAD`: one commit, five changed `agent_session` files only; no fast-path/durability `marks`, `fingerprint`, or `sync_dir` references in the patch.
- `agent_session/log/pkg.generated.mbti:10` and `:21` show the public API change is additive: `classify_header` plus `HeaderParse`; existing `parse_header` remains.
- `agent_session/store/store.mbt:172` now uses `classify_header`, with separate unsupported-version and non-header errors at `:174` and `:178`.
- Store behavior lines up with main’s `parse_session_file` flow: malformed JSON header still fails as “header line” at `agent_session/store/store.mbt:166`, while event tail parsing remains unchanged at `:188`.
- Tests cover the new gates: classifier behavior in `agent_session/log/log_test.mbt:172`, unsupported version load error in `agent_session/store/store_test.mbt:787`, and NUL id rejection in `agent_session/store/store_test.mbt:807`.
- Findings: none. `git diff --check origin/main...HEAD` is clean.

PR #343, opened 2026-07-03
fix(deepseek/client): bound streaming reads with an idle timeout

bobzhang:
## Problem

A streaming chat request could block **forever** on `read_until` when a peer
accepted the connection then went silent — no bytes, no EOF, no error. The
retry loop only fires on a *raised* error, so a silent stall never engaged it.
Observed in a long agent run as a ~108-minute wedge at the start of a new turn
(0% CPU, no events, empty stderr).

## Fix

Add `Client.idle_timeout_ms` (default **120s**; `0` disables). A new
`with_idle_timeout` helper wraps **both** the response-header read and each SSE
`read_until` in `@async.with_timeout_opt`, converting a stall into the existing
`RetryableApiError`:

- a stall **before any event** → retried on a fresh connection;
- a stall **after output** → bails to a clean failure (never replays a
  half-consumed completion).

It is an *idle* bound measured **per read** — the header line, then each SSE
event line — so it resets on every event (not every byte); a stream that keeps
delivering events never trips it. Applies to the streaming path (the mode the
agent and TUI use).

## Tests

- `streaming chat retries an idle stall before any event` → retries, `hits==2`.
- `streaming chat fails an idle stall after a delta without retrying` → `hits==1`.
- `streaming chat does not retry an idle stall after an unterminated data line`
  → flushes the buffered delta, fails without retry (`hits==1`).

`moon test deepseek/client` passes (incl. the live DeepSeek smoke test with a
valid key); `moon info`/`fmt` clean; `.mbti` diff is one field + one optional
constructor param.

## Validation in a real run

Rebuilt the agent with this fix and ran a 301-step / ~35-min task: it finished
in a **single pass with zero hangs** (before the fix, the analogous run hung
~108 min).

## Review

Reviewed twice with `codex exec review`; both findings addressed on the branch
(flush buffered SSE before an idle retry; per-event vs per-byte doc accuracy).

### DIFF
--- a/deepseek/client/README.mbt.md
+++ b/deepseek/client/README.mbt.md
@@ -57,6 +57,7 @@ test "construct DeepSeek client configuration" {
       #|  thinking: Max,
       #|  retry_attempts: 5,
       #|  retry_backoff_ms: 200,
+      #|  idle_timeout_ms: 120000,
       #|}
     ),
   )
--- a/deepseek/client/client.mbt
+++ b/deepseek/client/client.mbt
@@ -50,9 +50,13 @@ pub struct Client {
 /// converted to the same retryable class as a mid-stream drop — so a stall
 /// before any event is retried on a fresh connection, while one after output
 /// bails to a clean failure rather than duplicating a half-consumed completion.
-/// It is an *idle* bound, reset by every byte, so a long-but-progressing
-/// completion never trips it. Pass `0` (or less) to disable. Defaults to 120s;
-/// it applies to the streaming path only (the mode the agent and TUI use).
+/// The bound is measured per read — the header line, then each SSE event line —
+/// so it resets on every event, not on every byte: a stream that keeps
+/// delivering events never trips it. (The clock therefore covers one whole line;
+/// with real DeepSeek deltas, which are small and sub-second apart, a single
+/// line never approaches the default anyway.) Pass `0` (or less) to disable.
+/// Defaults to 120s; it applies to the streaming path only (the mode the agent
+/// and TUI use).
 pub fn Client::Client(
   api_key~ : String,
   model? : @deepseek.Model = default_model,
--- a/deepseek/client/client.mbt
+++ b/deepseek/client/client.mbt
@@ -27,6 +27,7 @@ pub struct Client {
   thinking : @deepseek.ThinkingMode
   retry_attempts : Int
   retry_backoff_ms : Int
+  idle_timeout_ms : Int
 } derive(Debug(ignore=API_KEY))
 
 ///|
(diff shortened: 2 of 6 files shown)


bobzhang:
## 🤖 Codex CLI review (`codex exec review --base main`)

Ran a fresh Codex review of this branch's diff. It surfaced one legitimate edge-case bug in the new idle-timeout path:

> The new idle timeout can retry after buffered SSE output has already been received but not flushed, violating the stream retry guard and risking duplicate completions or billing.

**- [P2] Flush buffered SSE data before retrying idle timeouts — `deepseek/client/stream.mbt:247-249`**
> If the peer sends a `data:` line and then stalls before the blank-line terminator, this timeout path raises `RetryableApiError` directly. That bypasses the existing `None` branch that calls `flush_sse_event`, so `on_stream_event` is never invoked and `chat_stream_body` keeps `streamed` false, allowing the retry loop to start a second completion even though model output has already arrived. Treat a timeout with buffered `data_lines` like EOF by flushing/marking the stream before raising.

This is correct — the idle-timeout branch must mirror the EOF branch's flush-before-raise so already-received output bails the retry. Addressing it in a follow-up commit on this branch.

bobzhang:
## Addressed ✅

Fixed the [P2] finding in `f6460acd`. The SSE loop now routes **both** EOF and idle-timeout interrupts through a single flush-before-raise path (`read_stream_line` distinguishes line / EOF / timeout), so a `data:` line received without its blank-line terminator is flushed — firing `on_stream_event` and delivering the delta — before any retryable error is raised. `with_idle_timeout` now guards only the response-header read, where no buffered state can exist.

Regression test added: an unterminated `data:` line followed by a stall delivers the delta and fails **without** retrying (`hits == 1`). Full package: 25/25 green.

bobzhang:
## 🤖 Codex CLI review — pass 2 (`codex exec review`, base = branch point)

Re-reviewed the current branch after the flush fix. One [P2], addressed as a doc correction:

> **[P2] Reset SSE idle timeout on byte progress** — `deepseek/client/stream.mbt:246`
> `read_until("\n")` doesn't return until the newline, so the timeout measures the whole line's arrival, not idle time. A large/slow `data:` line could trip it while bytes are actively flowing, despite the API implying progress resets the bound. Consider timing out individual reads and accumulating lines manually.

**Assessment / resolution.** Technically correct: the bound is per **event line**, not per **byte** — and the docstring overclaimed "reset by every byte." Fixed the wording in `c9ede865` to describe the per-event semantics accurately.

Not changing the mechanism to per-byte chunking: the practical risk is nil at the 120s default. DeepSeek SSE lines are individual deltas (a token / small tool-call-args chunk, sub-second apart), so a *single* line can't approach 120s while streaming; and the production hang this PR fixes was *between* events (waiting on the first byte of a new turn), which the per-read bound already catches. Per-byte chunking would replace `read_until` with a manual buffered read loop — more surface area (and this path already produced one subtle bug last review) for a benefit that can't materialize at the default. Documented the per-event semantics so nobody relies on a per-byte promise later.

PR #349, opened 2026-07-03
Auto-compaction: bound the prompt for long runs (--compact-threshold)

bobzhang:
## Summary

Long `--goal` runs currently grow their prompt without bound: a live 487-step goal run on this branch's base reached a **414k-token peak prompt**, and **98–99.5% of its gross token spend was cache-hit re-reads of history**. Compaction existed (`@compact.compact_session`) but was reachable only through TUI `/compact`, serve's `compact` command, and `sessions compact` — never from the headless run/goal loop, which is exactly the unattended surface that needs it.

This PR wires automatic compaction into the agent loop, on top of #345 (`goal-feature`):

- **Trigger**: at the step boundary (same safe point as the goal budget gate — every tool call has a recorded result), when the previous response's provider-reported `prompt_tokens` reaches `--compact-threshold` (default **200000**, env `OPENSEEK_COMPACT_THRESHOLD`, `0` disables).
- **Mechanics**: the summary is generated through the new public `@compact.generate_summary` (`compact_session` now delegates to it); the summarizer's own usage is recorded durably first (budget folds must see it), then the `Summary` terminates the covered range — all through `append_item`, the loop's single durable path. The projection is rebuilt afterwards.
- **Goal-aware**: an Active goal's objective is enveloped into the summarizer prompt so goal state survives the checkpoint; completed/blocked goals are not misrepresented as active.
- **Turn/resume boundaries**: the trigger seeds from the log's last post-summary `Usage` event, so a continuation or resumed turn whose projection is already oversized compacts *before* its first model call; the unread opening (user task or continuation envelope) is re-delivered verbatim after the checkpoint so it stays the latest instruction.
- **Safety**: budget re-checked after the summarizer's spend (compaction cannot buy an extra model call past a goal budget); compaction failure never ends a turn; a cooldown bounds pathological re-compaction; watcher fingerprints reset so post-compaction runtime updates are emitted in full.
- **CLI**: threshold validated before the fleet branch (invalid values fail in every mode, unlike the `--goal` fleet gap found in #345's review) and threaded through fleet attempts. serve/TUI keep their controller-driven `/compact` unchanged.

## Review process

Each commit was reviewed with codex CLI; three rounds surfaced 7 findings, all fixed and mostly pinned with tests: post-compaction budget bypass, non-active objective leakage, turn-boundary trigger blindness, silent disable via empty env value, stale watcher fingerprints, summarizer paraphrasing the unread opening, and summary-after-opening precedence.

## Validation

- `moon check` clean; `moon fmt` + `moon info` applied; **977 tests pass** (7 new: step-boundary compaction shape + projection rebuild, goal-objective threading, budget-crossing summarizer stop, continuation step-0 compaction, and 5 CLI parse cases).
- Live end-to-end on DeepSeek: a `--goal` run with `--compact-threshold 6000` fired `auto_compaction` at step 1 (9,127-token prompt), appended durable `Usage`+`Summary` (handoff summary leads with the objective), rebuilt the projection, and completed its goal with the artifact verified externally.

## Motivating data (from the C89 compiler goal e2e on the base branch)

| run | budget | steps | gross tokens | genuinely new | outcome |
|---|---|---|---|---|---|
| 1 | 250k | 16 | 274k | ~30k (11%) | budget-limited mid-scaffold |
| 2 (resume) | 600k | 10 | +330k | ~30k (9%) | budget-limited |
| 3 (resume) | 2B | 487 | +140.2M | ~714k (0.5%) | completed; 414k peak prompt |

With a 200k threshold those runs would have compacted ~2x/‌~70x cheaper in prompt re-reads and stayed far from the context wall.

### DIFF
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -2,6 +2,14 @@
 let default_max_steps : Int = 1000
 
 ///|
+/// Minimum number of steps between auto-compactions in one turn. Normally
+/// the trigger self-resets — the first post-compaction prompt is a fraction
+/// of the threshold — so this only bounds the pathological case where the
+/// summary plus recent history still exceeds the threshold, and it spaces
+/// out retries after a failed summarizer call.
+let compaction_cooldown_steps : Int = 8
+
+///|
 /// Queue `text` as steering input for the turn running on `runtime`.
 ///
 /// This is a thin public wrapper over `AgentRuntime::queue_steer(Prompt(...))`.
@@ -74,6 +82,7 @@ pub async fn run(
   thinking? : @deepseek.ThinkingMode = Max,
   api_url? : String,
   workspace_root? : String = ".",
+  compact_threshold? : Int,
 ) -> Unit {
   let session = @agent_session.Session(
     SessionId("one-shot"),
@@ -89,6 +98,7 @@ pub async fn run(
       thinking~,
       api_url?,
       workspace_root~,
+      compact_threshold?,
     ),
   )
 }
@@ -121,6 +131,7 @@ pub async fn run_turn(
   thinking? : @deepseek.ThinkingMode = Max,
   api_url? : String,
   workspace_root? : String = ".",
+  compact_threshold? : Int,
 ) -> @agent_session.Session {
   run_turn_with_append(
     api_key,
@@ -132,6 +143,7 @@ pub async fn run_turn(
     thinking~,
     api_url?,
     workspace_root~,
+    compact_threshold?,
   )
 }
 
@@ -165,6 +177,7 @@ pub async fn run_turn_with_append(
   thinking? : @deepseek.ThinkingMode = Max,
   api_url? : String,
   workspace_root? : String = ".",
+  compact_threshold? : Int,
 ) -> @agent_session.Session {
   @async.with_task_group() <| group => {
     run_turn_in_scope(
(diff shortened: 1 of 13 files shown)


bobzhang:
## 🤖 Codex CLI review (`codex exec review`, post-reorganize)

Reorganized this PR from 3 commits into 2 self-contained commits (the "address codex review" fixup folded by concern into the agent and cmd commits; the reorganized history is a byte-identical tree to the prior head). Then ran a fresh Codex review, which surfaced one real [P2]:

> **Preserve usage when summaries don't cover it** — `agent/agent.mbt`
> `last_recorded_prompt_tokens` reset the auto-compaction seed to `0` on *any* `Summary`. A partial summary (e.g. via `sessions compact --from/--to`) can be appended after a high-token `Usage` it does not cover — so a resumed run would seed `0`, skip the intended step-0 auto-compaction, and send an over-threshold prompt.

**Addressed** (folded into the auto-compaction commit): the seed now tracks the last `Usage` event's sequence and resets only when a summary's `from_sequence()..to_sequence()` actually covers it — reusing the same coverage semantics as the projection. Added a whitebox regression test (`agent/compaction_wbtest.mbt`): a partial summary that leaves a later high-token usage uncovered keeps the seed; a covering summary resets it.

`moon check` clean, agent + cmd/openseek tests green (91/91).

PR #352, opened 2026-07-03
feat(agent_session/store): suffix session file names with the session id

bobzhang:
## What

Name each durable session file after its session id — `sessions/<id>/openseek_session-<id>.jsonl` instead of the fixed `sessions/<id>/openseek_session.jsonl` — so files collected out of their directories (for cross-session visualization or comparison) stay self-naming. The per-session directory remains the home for future per-session artifacts, and `session.lock` keeps its fixed name: it already lives inside the per-id directory and is never collected alongside the jsonl.

This is a **clean break**, not a migration: there are no external users yet, so there is no legacy read-fallback, no migrate-on-write, and no alias URL. Stores written before this change simply are not loadable by the new binary (delete them or regenerate).

## Run-unique eval trial ids

Eval trial sessions previously used `trial-01`, `trial-02`, … — identical across every run by construction, which defeats flat collection for exactly the run-vs-run comparison this PR enables. Trial ids now embed a wall-clock stamp via `SessionId::generated` (`trial-01-YYYYMMDD-HHMMSS-mmm`), so any two runs' session files can sit in one directory without colliding.

## Details

- The viz server serves the raw file at `GET /api/sessions/<key>/openseek_session-<id>.jsonl`; the file segment is percent-decoded like the key segment, so ids with URL-reserved characters resolve correctly (regression-tested with an id containing a space).
- The eval harness scanner matches the suffixed name; fixtures updated to match.
- No public API changes (`pkg.generated.mbti` untouched).

## Testing

Full native suite green; `agent_session/store`, `eval/prompt_task/harness`, `cmd/viz_server`, `cmd/openseek`, and the JS `cmd/viz_app` suites all pass; `moon fmt` and `moon info` clean. Two fresh Codex CLI review passes ran during development; both findings (URL decoding of the file segment; a stat race that only existed because of the migration path, now deleted) are resolved.

### DIFF
--- a/agent_session/README.mbt.md
+++ b/agent_session/README.mbt.md
@@ -233,10 +233,10 @@ The store layout is:
 
 ```text
 <root>/sessions/<session-id>/
-  openseek_session.jsonl
+  openseek_session-<session-id>.jsonl
 ```
 
-`openseek_session.jsonl` holds the whole session: a header record
+`openseek_session-<session-id>.jsonl` holds the whole session: a header record
 (`{"version":1,"id":...,"system_prompt":...}`) on the first line, then one
 append-only event per line. Loading replays the event lines into a `Session`.
 
--- a/agent_session/store/README.mbt.md
+++ b/agent_session/store/README.mbt.md
@@ -14,17 +14,21 @@ Each session lives under:
 
 ```text
 <root>/sessions/<session-id>/
-  openseek_session.jsonl
+  openseek_session-<session-id>.jsonl
   session.lock
 ```
 
-`openseek_session.jsonl` is the whole durable session: its first line is a
-header record (`{"version":1,"id":...,"system_prompt":...}`) and every
-following line is one typed `SessionEvent`. Events are append-only. Loading
-replays the event lines into an immutable `agent_session.Session`.
+`openseek_session-<session-id>.jsonl` is the whole durable session: its first
+line is a header record (`{"version":1,"id":...,"system_prompt":...}`) and
+every following line is one typed `SessionEvent`. Events are append-only.
+Loading replays the event lines into an immutable `agent_session.Session`.
+The file name carries the session id so files collected out of their
+directories — for cross-session visualization or comparison — stay
+self-naming.
 
 `session.lock` is an implementation detail used to serialize writers and keep a
-reader from seeing a half-updated session.
+reader from seeing a half-updated session. It carries no id suffix: it already
+lives inside the per-id directory and is never collected alongside the jsonl.
 
 ## API Shape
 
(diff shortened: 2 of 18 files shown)


bobzhang:
Ran a fresh Codex CLI review of the first commit; it found two real defects, both fixed in a4f3ded9:

1. **Raw-file route vs encoded ids** — the route percent-decoded the key segment but not the file segment, so the id-suffixed filename comparison 404ed for ids containing URL-reserved characters (e.g. `run 1`). The file segment is now decoded exactly like the key, with a wbtest covering an encoded id for both the suffixed name and the legacy alias.

2. **Unlocked stat racing the migration** — `latest()`/`listings()` resolved a session file path and then stat'ed it without the session lock, so a concurrent append migrating `openseek_session.jsonl` → `openseek_session-<id>.jsonl` between the two steps made the newest session vanish from resume-newest (or list as damaged). The unlocked stat now probes suffixed → legacy → suffixed again; the migration rename is atomic and happens at most once, so a third miss means the file is genuinely absent.

Full native suite: 972/972 after the fixes.

bobzhang:
Second fresh Codex pass over the full diff: it confirms the file-segment decode is safe for the legacy alias, the triple-probe closes the `latest()`/`listings()` disappearance race, and no runtime call site still assumes the fixed filename.

It flags two residual races, both accepted deliberately rather than fixed:

- The read-only viz server resolves a path and then reads it without a lock, so a request landing exactly on the migration rename can get one `{"found":false}` envelope or one raw 404. Resolution prefers the suffixed name, so the very next request succeeds. The server is deliberately lock-free against live stores, migration happens at most once per session ever, and the frontend re-fetches — a retry loop here buys microseconds of consistency for permanent complexity.
- `listings()` can render one row with a timestamp but an empty first-prompt label if migration lands between the stat and the label read. Cosmetic and self-healing on the next listing.

No persistent misbehavior, data loss, or corruption in either. If we ever see these in practice the fix is the same triple-probe applied to the read paths.

bobzhang:
Direction change per review discussion: no external users yet, so 5bffa75d drops the whole compat apparatus (read-fallback, migrate-on-write, both-layouts guard, alias URL, dual-name scanner) — the layout is now a clean break, and the residual migration races noted above no longer exist because there is no migration. `session_file` returns to a pure path build, so the PR changes no public API. Also added run-unique eval trial ids (`trial-01-YYYYMMDD-HHMMSS-mmm`) so collected runs never collide. Net −214 lines vs the compat version; full suite 968/968.

bobzhang:
Third fresh Codex pass over the clean-break state. It cleared the trial-id propagation, analyzer resolution, and the raw-file route, but caught two things the `*.mbt`-scoped sweeps missed, both fixed in 9934e001:

- `tests/cram/cli.md` still wrote the old fixed `openseek_session.jsonl` as a fixture — a genuine failure, since `sessions show`/`sessions compact` can no longer load that name (the cram suite runs separately from `moon test`, which is why 968/968 didn't catch it). `moon cram test tests/cram` is now 23/23.
- Four READMEs (`agent_session`, `agent_session/store`, `viz`, `eval/prompt_task`) still documented the old filename and raw-file URL; following them would have produced unloadable files or 404s.

PR #353, opened 2026-07-03
refactor(viz,viz_app,viz_server): inline single-use tiny helpers

bobzhang:
Part of a sweep applying the new house rule: a private function called exactly once, whose body reads as well at the call site, is indirection rather than abstraction.

- **cmd/viz_server** (9 inlined): `parse_port`, `send_reply`, `strip_query`, `file_size`, `session_key`, `is_session_store`, `should_skip_scan_dir`, `root_label`, `last_path_separator`. `path_tail` absorbs the separator search via an or-pattern over the `rev_find` pair, replacing two chained matches.
- **cmd/viz_app** (12 inlined): update-loop wrappers (`apply_theme`, `fetch_text`), view builders (`theme_toggle`, `mode_toggle`, `args_toggle`, `drop_target_attrs`, `last_active_view`), and trivial formatters/predicates (`session_root_class`, `session_view_class`, `root_collapsed`, `path_basename`, `format_span_total_seconds`).
- **viz** (4 inlined): `plural`, `is_blank_args`, `summary_card`, `terminal_projects_model_message` (the latter becomes nested `Terminal(Finished(_))` patterns in `item_projects_model_message`).

Doc comments that carried non-obvious rationale moved to the call sites. Multi-use helpers were left alone. No public API changes (`.mbti` untouched), no behavior changes.

Tests: viz_server 17/17, viz_app 24/24 (js), viz 37/37 (js), JS bundle builds, `moon fmt`/`moon info` clean.

### DIFF
--- a/cmd/viz_app/app.mbt
+++ b/cmd/viz_app/app.mbt
@@ -157,9 +157,9 @@ fn update(emit : @cmd.Emit[Msg], msg : Msg, model : Model) -> (@cmd.Cmd, Model) 
       )
     Select(key) => {
       let label = session_label(model, key)
-      let command = fetch_text(emit, session_url(key), result => {
-        SessionFetched(key, result)
-      })
+      let command = @http.get(session_url(key)).expect_text(
+        emit.map(result => SessionFetched(key, result)),
+      )
       (
         command,
         {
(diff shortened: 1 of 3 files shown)


bobzhang:
Fresh Codex CLI review of this PR found one real defect, fixed in 9fac2cc2:

- **Evaluation-order change in `session_envelope_reply`** — inlining `file_size` hoisted the stat ahead of `read_text_lossy`, while the original JSON literal evaluated `events` (read) before `events_bytes` (stat). Under concurrent append to a live session file, the response could carry more text than the byte count it reported. The read is now hoisted alongside the stat in the original order, with a comment pinning the invariant (size must describe at most what was served).

Codex also verified every removed helper had no second caller and that the pattern-match rewrites (`path_tail` or-pattern merge, nested `Terminal(...)` patterns) preserve behavior. viz_server 17/17 after the fix.

PR #354, opened 2026-07-03
refactor(eval,agentcli): inline single-use helpers; share the API-key env lookup

bobzhang:

- **eval/prompt_task/harness** (15 inlined): path resolution (`workspace_path`, `resolve_relative_path`), scan filter (`should_skip_scan_dir`), report writers (`write_eval_result_files`, `write_suite_report_files`, `prepare_suite_output_dir`), metric formatting (`display_index`, `validation_status`, `format_average`, the three `average_*` wrappers), and manifest decoding (`suite_manifest_combos`, `problem_id_from_report` — the split/collect/index collapses to `split_once`). `string_array_contains` was hand-rolled `Array::contains` — replaced with the stdlib method.
- **eval/report**: `has_detail_section` → `self.rows.any(row => row.has_details())`.
- **internal/agentcli**: `api_key_arg_for_model` / `api_key_provider_for_model` inlined into `api_key`, the only caller.
- **De-dup**: `api_key_env_for_model` existed byte-for-byte in both eval harnesses while `internal/agentcli` already exports it — the exact drift hazard agentcli's package header describes. Both harnesses now import the shared copy.

Deliberately kept: `run_finish_case` in eval/tool_harness — one of eight symmetric `run_*_case` builders; breaking the family for the one short member costs more than the indirection.

No public API changes, no behavior changes. Tests: prompt_task/harness 18/18, file_edit/harness 7/7, report 4/4, cmd/tui 80/80 (agentcli consumer); `moon fmt`/`moon info` clean.

### DIFF
--- a/eval/file_edit/harness/harness.mbt
+++ b/eval/file_edit/harness/harness.mbt
@@ -177,16 +177,8 @@ fn trial_extra_env(
     // own ~/.openseek/skills must never reach a trial prompt.
     "OPENSEEK_GLOBAL_SKILLS_DIR": "\{real_workspace}/.no-global-skills",
   }
-  env[api_key_env_for_model(config.model)] = config.api_key
+  env[@agentcli.api_key_env_for_model(config.model)] = config.api_key
   env
-}
-
-///|
-fn api_key_env_for_model(model : @deepseek.Model) -> String {
-  match model {
-    KimiK27Code | KimiK27CodeHighSpeed => "KIMI"
-    V4Flash | V4Pro => "DEEPSEEK"
-  }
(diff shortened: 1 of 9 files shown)


bobzhang:
Fresh Codex CLI review: **no findings**. It verified the inline sites (path resolution, suite-manifest combo parsing, average rounding half-up, API-key env/provider lookup), the `Array::contains`/`Array::any` stdlib replacements, both new `@agentcli` imports, and confirmed every removed helper had exactly its definition plus the one expected caller — no stragglers in tests or wbtests.

PR #355, opened 2026-07-03
refactor(agent_session,store,agent,agent_skill): inline single-use helpers

bobzhang:

- **agent_session**: `terminal_kind` + `terminal_message` collapse into one match extracting `(kind, message)` in `TurnTerminal::to_json`; `summary_model_content` and `summary_covers_event` inline into the projection (`is_covered_by_later_summary` gains a `guard … is Summary` + range check in place of the predicate call); single-site named constructors `SessionEvent::SessionEvent` and `SessionToolCall::from_deepseek` become struct literals.
- **agent_session/store**: `session_file_name`, `header_line`, `sync_dir_best_effort`, and the `ensure_dir_exists` attempts-defaulting wrapper fold into their only callers, doc comments moving with them.
- **agent**: `plan_reminder_text` inlines into the reminder-injection site with its rationale comment.
- **agent_skill**: the `Skill::Skill` constructor's one construction site uses a struct literal.

Deliberately out of scope: **deepseek/client** (4 candidates) — the open `fix/deepseek-idle-timeout` branch rewrites its stream internals, and sweeping it now would manufacture conflicts. It can follow once that branch lands.

No public API changes, no behavior changes. Tests: agent_session 17/17, store 45/45, agent 14/14, agent_skill 5/5; `moon fmt`/`moon info` clean.

### DIFF
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -11,21 +11,6 @@ let default_max_steps : Int = 1000
 let plan_reminder_after_steps : Int = 10
 
 ///|
-/// The staleness reminder injected when the `plan` tool has gone unused for
-/// `plan_reminder_after_steps` model requests. With a known current plan the
-/// reminder re-surfaces it — the last plan call may sit dozens of steps back
-/// in context — and otherwise it nudges once toward recording a plan, while
-/// staying explicit that trivial work should ignore it.
-fn plan_reminder_text(checklist : String?) -> String {
-  match checklist {
-    Some(lines) =>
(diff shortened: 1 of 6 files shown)


bobzhang:
Fresh Codex CLI review: **no findings**. It verified the merged TurnTerminal match maps all four variants to the original kind strings, the projection inlines keep the coverage comparisons and summary template byte-equivalent, `@fs.File::close` is non-raising (so the sync-loop inline adds no failure path), `ts: unix_ms.to_double()` survived the constructor inline, the plan-reminder strings are byte-identical, and no removed helper has a remaining caller anywhere including wbtests and .mbt.md.

PR #356, opened 2026-07-03
refactor(agent_tool): inline single-use helpers outside shell

bobzhang:

- **internal/auto_check**: `format_moon_check_output` (arg-splat wrapper), `containing_dir`, `is_moonbit_source_path`, `is_windows_drive_root` fold into their conditions/callers; `indent_context` and `pad_line_number` inline into the parse-gate excerpt rendering with the rationale comment preserved.
- **moon_check**: `ensure_trailing_newline` becomes the if-expression at its one `initial_output=` site.
- **moon_cmd/internal/decode**: `required_string`, `validate_command`, `validate_target_supported`, `cap_max_output_chars` fold into `parse_fields`, where the field order already narrates the validation. `optional_string` stays — five callers.
- **read/internal/decode**: its single-use `optional_string` and `cap_max_output_chars` copies inline away (removing two of the three cross-package duplicates of the latter).
- **multi_edit / plan / edit / write**: message formatters (`format_failure`, `line_range_label`, `range_detail`), the `is_verification_title` heuristic, and `create_parent_dirs` inline with their doc comments as call-site comments.

The shell package (48 candidates) is deliberately a separate PR — its helper names document a dense policy engine and need case-by-case judgment.

No public API changes, no behavior changes. Tests: auto_check 23, moon_check 19, moon_cmd 21, multi_edit 30, plan 12, read 23, edit 29, write 26 — all green; `moon fmt`/`moon info` clean.

### DIFF
--- a/agent_tool/edit/edit.mbt
+++ b/agent_tool/edit/edit.mbt
@@ -133,8 +133,12 @@ async fn edit_file(
     .map(part => part.to_owned())
     .collect()
   if parts.length() <= 1 {
+    let range_detail = match input.end_line {
+      Some(end) => " in line range \{input.start_line}-\{end}"
+      None => " from line \{input.start_line}"
+    }
     return @agent_tool.ToolAction::respond(
-      "error editing \{path}: old_string not found\{range_detail(input)}",
+      "error editing \{path}: old_string not found\{range_detail}",
       is_error=true,
       brief=fail_brief,
     )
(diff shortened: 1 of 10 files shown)


bobzhang:
Fresh Codex CLI review: **no findings**. It checked the flagged high-risk spots — the `||`-chain block for the .mbt suffix check, the `context_lines` rename against the outer `lines` binding, the nested-interpolation padding, decode validation order and fail strings, and the read-path error-shape merge — and confirmed behavior and error strings are preserved with no orphaned callers. It also ran the eight affected package test suites itself; all green.

PR #357, opened 2026-07-03
refactor(cmd/openseek): inline single-use helpers

bobzhang:

Inlined: `within_root` (guard in `copy_symlink`), the `FlushedStdout` constructor, `explicitly_provided` (argv/env source check), `format_session_listing` (the tab-separated row moves into the `sessions list` loop with its scriptability comment), and `is_windows_drive_root` (absorbed into `is_path_root`, its only caller, as a commented clause).

Deliberately kept, and why:
- `run_task_command` / `run_serve_command` / `run_review_command` / `run_sessions_command` and `sessions_show` — symmetric dispatcher families; the one short member reads better matching its siblings, and the docstrings carry per-command logging policy.
- `default_global_skills_dir` — a `#cfg(platform)` pair; the function *is* the conditional-compilation dispatch.

No public API changes, no behavior changes. Tests: cmd/openseek 59/59, cram suite 23/23 (CLI behavior unchanged); `moon fmt`/`moon info` clean.

### DIFF
--- a/cmd/openseek/concurrent.mbt
+++ b/cmd/openseek/concurrent.mbt
@@ -278,18 +278,13 @@ async fn copy_regular(s : String, d : String) -> Unit {
 /// skipped to avoid loops or escaping `--dir`.
 async fn copy_symlink(s : String, d : String, root : String) -> Unit {
   let resolved = @fs.realpath(s) catch { _ => return }
-  guard within_root(resolved, root) else { return }
+  // Only targets that are the workspace root or nested under it.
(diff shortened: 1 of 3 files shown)


bobzhang:
Fresh Codex CLI review: **no findings**. It verified the `FlushedStdout` literal matches the removed constructor's defaults, the `sessions list` row stays byte-identical (tab separation, `-` fallback), the absorbed drive-root clause is unchanged, the `within_root` guard keeps root-or-nested semantics, and no removed helper has other callers including tests.

PR #359, opened 2026-07-03
refactor(scripts,testkit): inline single-use helpers

bobzhang:

- **scripts/internal/markdown_comments**: `comment_ranges`, `inline_raw_html_is_comment`, and `push_loc_range` fold into their single call sites; the comment test and the loc-presence guard merge into one condition. `is_html_comment` stays — it has two callers (the initial sweep scan miscounted it; verified by grep before touching).
- **scripts/md_to_mbt_string**: `trim_windows_drive_prefix` inlines into the windows-only `basename` arm — the caller is already `#cfg(platform="windows")`, so this leaves one cfg block instead of two.
- **testkit/filesystem**: the outline dir filter and `outline_indent` inline with their comments.

Note: four `md_to_mbt_string` expect tests fail identically on clean `origin/main` (pre-existing snapshot drift, verified by stashing the diff and re-running on the baseline) — unrelated to this change. The markdown_comments and testkit suites pass.

No public API changes, no behavior changes. `moon fmt` (both modules) and `moon info` clean.

### DIFF
--- a/scripts/internal/markdown_comments/markdown_comments.mbt
+++ b/scripts/internal/markdown_comments/markdown_comments.mbt
@@ -6,18 +6,12 @@ priv struct SourceRange {
 
 ///|
 pub fn without_comments(markdown : String) -> String {
-  let ranges = comment_ranges(markdown)
-  guard !ranges.is_empty() else { markdown }
-  ranges.sort_by(fn(left, right) { left.first.compare(right.first) })
-  remove_source_ranges(markdown, ranges)
-}
-
-///|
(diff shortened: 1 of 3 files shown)


bobzhang:
Fresh Codex CLI review: **no findings**. It verified the merged comment-and-loc condition is equivalent to the two-step original (including evaluation order), the RawHtml arm braces, the cfg'd drive-prefix pattern, and unchanged testkit outline output; one nuance it noted: `is_html_comment` keeps one direct caller (the other call was the one folded in), and `loc_is_absent` is the helper retaining two sites. Reviewer ran the affected package checks/tests itself; all green.

PR #360, opened 2026-07-03
refactor(agent_tool/shell): inline the mechanical single-use helpers

bobzhang:
Final part of the single-use-helper sweep, and the one where the rule needed the most judgment.

**Inlined (18)** — bodies that read as well at the call site:
- String utilities: `shell_quote`, `chars_to_string`, `strip_trailing_cr`, `text_prefix_by_chars`, `exit_header`, and the `test_shell_quote` duplicate in the test file.
- `internal/decode`: `required_string`, `optional_string`, `optional_positive_int_or_default`, `cap_max_output_chars` fold into `parse_fields`.
- `internal/command_policy`: `sed_in_place_error_message` (a constant), `command_basename` (guard with comment), `clean_shell_word`.
- `internal/git_policy`: `short_cluster_has`, `clean_arg_is_exclude` (commented condition clauses).
- `internal/shell_parse`: `is_shell_space` and `is_separator` become is-patterns (`ch is (' ' | '\t')`, `op is ("&&" | "||" | ";" | "|")`); the three `env_option_*` predicates merge into their two scan conditions.

**Deliberately kept (~35)** — the policy predicates in `sandbox.mbt` and `moon_auto_check.mbt` (`heredoc_*`, `find_*`, `awk_*`, `install_*`, `powershell_*`, `sandbox_denial_*`, `*_option_consumes_next`, …). Their names *are* the documentation of the sandbox policy engine; the bodies are token-index arithmetic that reads far worse inlined into the scanning loops. Also kept: `shell_description` (a `#cfg` platform pair — the function is the dispatch) and `first_command_word` (a search loop that cannot sit inside its `else if` condition; its callee `looks_like_env_assignment` has two callers).

No public API changes, no behavior changes. Tests: shell 106/106, command_policy 9/9, git_policy 6/6, shell_parse 22/22, decode 6/6; `moon fmt`/`moon info` clean.

### DIFF
--- a/agent_tool/shell/internal/command_policy/command_policy.mbt
+++ b/agent_tool/shell/internal/command_policy/command_policy.mbt
@@ -27,7 +27,9 @@ pub fn moonbit_command_policy_error_for_parse(
     Some(moon_cmd_error_message())
   } else if parsed is Simple(commands) {
     if commands.any(sed_edits_files) {
-      Some(sed_in_place_error_message())
+      Some(
+        "error: `sed -i` edits files in place and is unreliable for code changes; use line-anchored `edit` (or `multi_edit` for several fixes in one file) for source changes",
+      )
     } else {
       None
     }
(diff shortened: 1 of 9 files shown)


bobzhang:
Fresh Codex CLI review: **no findings**. It checked the sandbox quote/probe string escaping, heredoc CR-strip/tab-strip ordering, `chars_to_string` loop bounds, the decode failure strings for every input shape (required vs optional differ on null/missing), the sed basename guard's `None => name` arm, the git/shell-parse condition merges and is-pattern sets, and confirmed the kept policy predicates are untouched with no orphaned callers. Test suites were run before push: shell 106/106 plus all four internals green.

PR #363, opened 2026-07-04
refactor(deepseek,agent,tui): usage log event embeds ToJson-encoded Usage

bobzhang:
## Summary

Two commits, ordered for review:

1. **`test(tui): drop redundant Json::object wrappers in event fixtures`** — pure mechanical cleanup. `append_items` and `decode_event` take `Json`, so map literals coerce via Json literal overloading; 42 wrappers removed across `jsonl_wbtest.mbt` and `decode_test.mbt`. The one survivor is `Json::string("not an object")`, where non-object-ness is the point of the test. No behavior change; validated green in isolation.

2. **`refactor(deepseek,agent,tui): usage log event embeds ToJson-encoded Usage`** — the actual schema change:
   - `@deepseek.Usage` derives `ToJson`; `print_usage` becomes `@xlog.info() <? { "event": "usage", "usage": usage }`. The log line now nests the struct verbatim — raw field names (`prompt_cache_hit_tokens`, not the old renamed `cache_hit_tokens`) and `total_tokens` included.
   - The package's existing lenient `FromJson` impl for `Usage` (absent/malformed fields → 0, never raises) becomes `pub`, and the TUI decoder round-trips the nested object through it instead of hand-reading fields — the wire schema is owned by `deepseek` in both directions.
   - TUI decode stays crash-free by construction; fixtures updated to the nested shape.

## Design note: `ToJson` derived, `FromJson` manual — intentional asymmetry

`derive(FromJson)` was considered and rejected:

- A derived decoder is strict — it raises `JsonDecodeError` on a missing field. The manual impl decodes absent/malformed fields as 0 and never raises.
- That leniency is load-bearing beyond the TUI: `ChatResponse`'s decoder routes API usage objects through this impl, including the streaming client's accumulated envelope (`stream.mbt`). `client_stream_test.mbt` fixtures a usage object with only `prompt_tokens`/`completion_tokens`/`total_tokens` — the shape OpenAI-compatible endpoints send when they lack DeepSeek's cache-token extension. A derived impl would raise on those streams.
- Derive's per-field `default=` argument is deprecated (the compiler's advice: "Use a manual trait implementation instead") and still raises on wrong-typed fields, which `json_object_int` tolerates.

So: `ToJson` derived (we own the writer; the struct is the schema), `FromJson` manual (we don't own every producer; tolerance required). The doc comment on the now-`pub` impl states this contract so it doesn't get "simplified" into a derive later.

## Public API impact

`deepseek/pkg.generated.mbti`: `Usage` gains `derive(ToJson)` and `pub impl @json.FromJson for Usage`. Both additive.

**Log schema note:** the `usage` JSONL event changes shape (nested + original field names). Engine and TUI are updated together here; viz (`viz/`, `cmd/viz_server`, `cmd/viz_app`) does not consume this event. An older TUI binary reading a newer engine would show 0 tokens (the decoder is deliberately tolerant), not crash.

## Validation

- `moon check` clean; `moon fmt` stable; `moon info` shows only the additive API above
- `moon test deepseek agent cmd/tui`: 134/134
- Full `moon test`: 964/968 — the 4 failures are the known pre-existing `agent_tool/shell` output-limit tests, unrelated

### DIFF
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -592,12 +592,5 @@ async fn @agent_session.Session::apply_steer_inputs(
 
 ///|
 fn print_usage(usage : @deepseek.Usage) -> Unit {
-  @xlog.info() <?
-    {
-      "event": "usage",
-      "prompt_tokens": usage.prompt_tokens,
-      "completion_tokens": usage.completion_tokens,
-      "cache_hit_tokens": usage.prompt_cache_hit_tokens,
-      "cache_miss_tokens": usage.prompt_cache_miss_tokens,
-    }
+  @xlog.info() <? { "event": "usage", "usage": usage }
 }
--- a/cmd/tui/internal/event/decode.mbt
+++ b/cmd/tui/internal/event/decode.mbt
@@ -28,9 +28,15 @@ pub fn decode_event(object : Json) -> AgentEvent? {
     "runtime_update" => Some(RuntimeUpdate(@jsonx.string(object, "content")))
     "session_error" => Some(SessionError(@jsonx.string(object, "error")))
     "usage" => {
-      let prompt = @jsonx.int(object, "prompt_tokens")
-      let completion = @jsonx.int(object, "completion_tokens")
-      Some(Usage({ total_tokens: prompt + completion }))
+      let usage : @deepseek.Usage = @json.from_json(
+        @jsonx.field(object, "usage").unwrap_or(Json::null()),
+      ) catch {
+        // Unreachable with the lenient Usage impl; drop rather than crash.
+        _ => return None
+      }
+      Some(
+        Usage({ total_tokens: usage.prompt_tokens + usage.completion_tokens }),
+      )
     }
     "tool_result" =>
       Some(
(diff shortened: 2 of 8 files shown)


bobzhang:
**Why `FromJson` for `Usage` stays hand-written while `ToJson` is derived** (considered during review of this change):

- A derived `FromJson` is strict: it raises `JsonDecodeError` on a missing field. The manual impl decodes absent/malformed fields as 0 and never raises.
- That leniency is load-bearing beyond the TUI: `ChatResponse`'s decoder routes API usage objects through this impl, including the streaming client's accumulated envelope (`stream.mbt`). `client_stream_test.mbt` fixtures a usage object with only `prompt_tokens`/`completion_tokens`/`total_tokens` — the shape OpenAI-compatible endpoints send when they lack DeepSeek's cache-token extension. A derived impl would raise on those streams.
- Derive's per-field `default=` argument would only partly help and is deprecated — the compiler's own advice is "Use a manual trait implementation instead". It also still raises on wrong-typed fields, which `json_object_int` tolerates.

So the asymmetry is intentional: `ToJson` derived (we own the writer; the struct is the schema), `FromJson` manual (we don't own every producer; tolerance required). The doc comment on the now-`pub` impl states this contract so it doesn't get "simplified" into a derive later.

PR #375, opened 2026-07-06
feat(shell): explicit background jobs (run_in_background + shell_output + shell_stop)

bobzhang:
## Why

Follows Claude Code's practice for long-running shell commands: run in the background, poll output, stop on demand — instead of blocking a turn or being killed at a timeout. This is **PR A** (explicit background jobs); a follow-up **PR B** adds detach-on-timeout + push-completion.

## What

- **`agent_tool/bgjobs`** — a session-scoped `BgJobRuntime` that spawns children on the agent's task group (the same structured-concurrency pattern as the former `moon_check` watcher), so jobs are reaped on session end and never orphan. Each job keeps a tail-capped merged-output buffer and a status (`Running | Exited(code) | Stopped`); a per-job monitor drains the pipe and records the terminal state. `start`/`snapshot`/`list`/`stop`.
- **`shell` gains `run_in_background`** — detaches the command as a job and returns its id instead of blocking. It runs **after the same cwd + source-tree write guards and through the same sandbox** (`sandbox-exec` write-blocking profile) as a foreground command, so it cannot bypass any guardrail. `timeout_ms` + `run_in_background` is rejected (contradictory).
- **`shell_output(job_id)`** — read a job's captured output + status; a non-zero exit is a tool error.
- **`shell_stop(job_id)`** — cancel a job's process.
- `build_tools` creates one shared runtime per session and wires all three tools (9 standard tools now).

## Design notes

- The runtime is intentionally **not generic** over the task group's return type: `new` captures the group into spawn closures, so `BgJobRuntime` can be an optional tool argument without leaking an unresolved type variable to callers that never use background jobs.
- Cancellation reaches only the launched process (like the foreground `shell` tool — `@process.hard_cancel`); the process library exposes no process-group kill. Documented, not overclaimed.

## Testing

39 tests across `agent_tool/bgjobs` (16), `shell` background paths, `shell_output`, `shell_stop`, and the agent registry. bgjobs coverage includes: run-to-exit + full output capture, EOF-vs-exit decoupling (a child that exits while a descendant holds the pipe), stop/wait races, concurrent stops, streaming RFC-3629 UTF-8 across read-chunk boundaries, output-budget enforcement, and terminal-status flush ordering.

Reviewed commit-by-commit with `codex review`; ~15 real correctness/safety findings across the series were fixed with regression tests, and the final pass is clean.

### DIFF
--- a/agent/README.mbt.md
+++ b/agent/README.mbt.md
@@ -114,7 +114,9 @@ calls `@agent.run`, but that decision lives outside the `agent` package.
 `build_tools(runtime, scope)` returns the standard local tool registry:
 
 - `shell`: run a command under the workspace root or an explicit cwd (including
-  `moon check` for compiler feedback);
+  `moon check` for compiler feedback), optionally detached via `run_in_background`;
+- `shell_output`: read the captured output and status of a background `shell` job;
+- `shell_stop`: stop a background `shell` job;
 - `read`: read a text file;
 - `edit`: replace exact text in a file;
 - `multi_edit`: apply several explicit line-anchored replacements to one file;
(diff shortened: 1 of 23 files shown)


bobzhang:

PR #378, opened 2026-07-06
feat: push-completion notices for background jobs (stacked on #375)

bobzhang:
## Why

**PR B**, stacked on #375 (explicit background jobs). Adds Claude Code's *push-notification* behavior: when a background job finishes, the model is told — it doesn't have to poll `shell_output`.

Base this PR on `feat/background-shell` (PR #375); rebase onto `main` once that merges.

## What

- **`bgjobs` gains an `on_job_exit` callback** — fired once when a job exits on its own (not when stopped, not on session teardown).
- **`build_tools` wires it to `runtime.queue_steer(Notice(...))`** and an `on_background_wake` hook. A `Notice` variant is added to `SteerInput` (delivered as user-role model content, like `Prompt`/`Command`) and persisted as a `Runtime(RuntimeNotice)` item so it renders as a notice — not user input — on resume.
- **Two delivery paths, both correct:**
  - Job finishes **mid-turn** → the loop injects it live at the next step boundary (run-id-tagged, shown).
  - Job finishes while the engine is **fully idle** (serve mode) → a new `BackgroundNotice` serve step wakes the loop, persists the notice, and emits it as an **untagged** `background_notice` event that survives the TUI's stale-run guard → renders as a `⚙` transcript line.
- **`shell` tool docs** describe `run_in_background` and the `shell_output`/`shell_stop` companions.

## Not included

**Detach-on-timeout** (moving a timed-out command to the background instead of killing it) is intentionally **not** here. Routing foreground commands through `bgjobs` hit irreducible foreground/background semantic mismatches (output retention is prefix-vs-tail, cancel-on-limit vs keep-running, notify semantics). The correct implementation is a deeper refactor — run in the foreground but spawned on the session group, hand the process off to `bgjobs` on timeout — which preserves foreground semantics natively. That's a focused follow-up (a spike is on `spike/detach-on-timeout`).

## Testing

The `Notice` path was hardened over several review passes across the queue filter, serve loop (idle drain + wake), TUI decode/render, and session persistence. Full touched-package suites pass (agent, bgjobs, shell, serve, tui, tui/core, event); the only failures are the pre-existing flaky shell output-limit/UTF-8 timing tests that fail on `main` too.

Reviewed commit-by-commit with `codex review`.

### DIFF
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -296,7 +296,7 @@ fn @agent_runtime.AgentRuntime::pending_steers(
 ) -> Array[@agent_runtime.SteerInput] {
   [
     for input in runtime.drain_steers() if input
-    is (Prompt(text) | Command(text)) &&
+    is (Prompt(text) | Command(text) | Notice(text)) &&
     !text.trim().is_empty() => input
   ]
 }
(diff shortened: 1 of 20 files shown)


bobzhang:

PR #380, opened 2026-07-06
feat: unified shell execution model + detach-on-timeout (stacked on #378)

bobzhang:
## Why

**PR C**, stacked on #378. Implements `docs/plans/background-shell.md` — converging foreground `shell` and background jobs onto Claude Code's **single-execution model**, which finally makes **detach-on-timeout** clean.

Detach-on-timeout was dropped earlier (the B2 spike) because routing foreground commands through `bgjobs` hit irreducible foreground/background mismatches (prefix-vs-tail output, cancel-vs-run, notify semantics). The fix wasn't to patch those — it was to **unify the output model first**, so there's nothing to reconcile.

Base on `feat/background-shell-parity` (#378); rebase onto `main` as the stack merges.

## What (7 commits, each reviewable)

- **C1** `ShellOutputSink` — one prefix-retaining, UTF-8-decoding output owner (StringBuilder-backed, pure/sync).
- **C2** `ShellExecution` — one process + one sink + a status flag (`Running | Backgrounded | Exited | Killed`); generalizes bgjobs' monitor. "Background" is a status, so detach is a flip.
- **C2b** bgjobs delegates to `ShellExecution` — deletes its duplicate monitor/UTF-8/buffer; a job is now metadata + a shared execution. **This is what unifies the output model** (background is now prefix-retained, like foreground).
- **C3** foreground `shell` routes through `ShellExecution` (on the session group, with a foreground output-limit kill), when the agent provides a spawner. Scope-less callers (the TUI `!`, unit tests) keep the per-call path.
- **C4** **detach-on-timeout** = `exec.background()` + `BgJobRuntime::adopt`. A timed-out command becomes a background job the model watches with `shell_output`/`shell_stop`, instead of being killed.
- **C6** tool/prompt reconcile: description says timeouts detach; `shell_output`/`shell_stop` needed no change (they already read through bgjobs → the shared sink).
- **fix** stop a foreground execution on **turn cancellation** (Ctrl-C/teardown) — a session-group process would otherwise orphan. Uses a synchronous `request_stop()` (found by `codex review`).

**C5 (size watchdog): intentional no-op.** C1 shipped as a bounded *pure-memory* sink (not buffer-then-spill), so there's no file to fill — the plan itself notes the watchdog is only load-bearing with file-backed output. The large-output **file pointer** is likewise deferred (large output still returns the output-limit error); documented in the plan.

## Testing

Every commit built + tested; reviewed commit-by-commit and as a whole with `codex review` (which caught the cancellation orphan). Full repo: `moon check` clean, **1018/1018** native tests pass.


bobzhang:

PR #384, opened 2026-07-07
feat: shell execution model — file-backed output foundation (1/3)

bobzhang:
## The big picture

This is **PR 1 of 3** re-delivering OpenSeek's shell background-job feature on a single, future-proof **execution model**, foundation-first (off `main`). It replaces the earlier stack (#375/#378/#380), which built the feature in historical order (explicit jobs → push → unify-and-refactor) with a bounded pure-memory sink.

**The one idea:** a shell command is *one process object with one output owner and a status flag*. "Foreground" vs "background" is only whether a tool call is currently awaiting it — so detaching is a **status flip**, not a re-route. (The earlier `spike/detach-on-timeout` failed precisely by routing foreground *through* a background runtime with opposite output policies; the lesson is one runtime, not two reconciled.)

Full architecture: `docs/plans/shell-execution-model.md`.

## This PR — the model (layers L0–L2), pure infrastructure, no tool wiring

- **`ShellOutputSink` (L0)** — one output owner, **file-backed**. Memory-first: output is kept in memory up to an inline cap, then the *full* output spills to a file and only a bounded head is kept for inline rendering. **The file is the source of truth**, so a reader picks the window (`read_all` now; prefix/tail/range later) rather than the buffer baking in a retention policy. This is what makes it future-proof: large output and tail-follow are read variants, not redesigns. No spill path → memory-only (bounded).
- **`ShellExecution` (L1)** — process (session group) + sink + `status: Running|Backgrounded|Exited|Killed` + a monitor (concurrent reader + authoritative `wait` + drain + terminal flush). `background()` detaches; `kill_when_full` is the foreground output-limit kill; `request_stop()` is a **synchronous** cancel safe to call while a turn unwinds (so a session-group process can't orphan).
- **`BgJobRuntime` (L2)** — registry `id → ShellExecution`: `start` spawns + registers; `adopt` registers an already-running execution (how detach will work); `snapshot` is the inline preview, `read_output` returns the full spill file; a watcher fires `on_job_exit` once on natural exit.

## Testing

**24/24** native tests (real processes): spill + seed-from-head + full read-back, memory-only fallback, UTF-8 across chunk *and* spill boundaries, hard-cap truncation, exit codes, stop/`request_stop`/`background`, `kill_when_full`, `adopt`, `on_job_exit` once, id↔file consistency. Reviewed with `codex` (fixed a job-id/spill-file mismatch).

## Next

- **PR 2 — tools (L3):** `shell` (foreground through the model, `run_in_background`, detach-on-timeout, cancellation cleanup), `shell_output`, `shell_stop`.
- **PR 3 — integration:** push-completion, serve-loop wake, TUI.

### DIFF
--- a/agent_tool/bgjobs/bgjobs.mbt
+++ b/agent_tool/bgjobs/bgjobs.mbt
@@ -1,0 +1,242 @@
+///|
+/// Default retained-output budget for a background job's inline preview, in
+/// characters. Matches the `shell` tool's foreground default so a command reads
+/// the same whether it runs in the foreground or the background. Full output
+/// beyond this spills to the job's file (see `ShellOutputSink`).
+let default_max_output_chars : Int = 12000
+
+///|
+/// Terminal or live state of a background job.
+///
+/// - `Running`: still executing (a detached-but-running execution also reads as
+///   `Running` — a poller has no separate "detached" state).
+/// - `Exited(code)`: exited on its own with `code`.
+/// - `Stopped`: cancelled (`stop`, a reader error, or teardown) before exit.
+pub(all) enum BgJobStatus {
+  Running
+  Exited(Int)
+  Stopped
+} derive(Eq)
+
+///|
+/// Session-scoped registry of background ("detached") jobs, each backed by one
+/// shared `@shell_exec.ShellExecution` (a process on `scope.group()`, a
+/// file-backed output sink, and a status flag). Every child is spawned on the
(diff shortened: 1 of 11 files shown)


bobzhang:
Superseded by the reorganized series: umbrella #389 (complete diff + architecture) and merge slices #399 → #400 → #401 → #402. This PR's content is contained in slice #399 (plus later hardening); its review history remains valid — the foundation was additionally re-reviewed clean against current main.

PR #388, opened 2026-07-07
feat: background shell tools on the execution model (2/3)

bobzhang:
Stacked on #384 (foundation). This is **2/3** of the shell execution-model series.

## Big picture

The foundation (#384) gave us one object — `ShellExecution` (a process + a single file-backed `ShellOutputSink` + a status flag) — where **"background" is a status, not a re-route**. This PR builds the tool surface on it:

```
L3   shell (run_in_background, foreground-through-execution, detach-on-timeout)
     shell_output (poll)   shell_stop (cancel)
L2   BgJobRuntime  ── session registry of executions, spill files, watchers
L1   ShellExecution ── one process, one output owner, one status   (#384)
L0   ShellOutputSink ── memory-first, file-backed full output        (#384)
```

The key idea carried through: **foreground and background are the same execution.** A foreground `shell` call now runs *through* a `ShellExecution` on the session group (not a separate per-call collector), which is exactly what lets a command that outlives its `timeout_ms` be **detached** — flipped to a background job the model keeps watching — instead of killed. Detach is a status flip + `adopt`, not a second code path.

## What's here

- **`shell` `run_in_background`**: start a command as a job, return its id immediately.
- **`shell_output` / `shell_stop`**: poll a job's output (bounded tail window + truncation metadata) and cancel it by id.
- **Foreground-through-execution**: the wired foreground path uses the shared model; a timed-out command detaches rather than dying.
- **`BgJobRuntime`**: one per session, owns the job registry, per-job spill files (removed at scope teardown), and exit watchers. It stays generic-free by capturing the session group in spawn closures.

## Full foreground/background parity (the hardening)

Background jobs read later via `shell_output` now match every foreground error semantic — each of these was found and fixed under adversarial review:

- **Output limits**: a timed runaway is killed at the inline cap *promptly* (immediate `output_limit_reached`), not run to the timeout; a job detached while still under the limit keeps growing (file-backed) afterward.
- **Binary output**: non-UTF-8 output is a `Malformed` tool error in both paths, surfaced *mid-stream* (not only at exit).
- **Sandbox denials**: a source-write denial in a background/detached job is a tool error with the same guidance the foreground path appends; detection scans the full retained output and the launch's sandbox metadata rides through detach.
- **Truncation reporting**: `shell_output` marks truncated whenever what it shows is less than what the job produced (window, memory-only drop, or hard-cap), never presenting a prefix as complete; metadata is sampled *after* the awaited read to avoid a stale-snapshot TOCTOU.
- **Lifecycle**: retained foreground spill files that finish un-detached are discarded; the session spill dir is removed when the scope ends; `run_in_background` is only advertised when a runtime is wired.

## Tests

954/954 native tests pass; `moon fmt` clean. New tests cover run_in_background, poll/stop, detach-on-timeout, detached-job retention, the prompt output-limit kill, binary output (both paths), truncation/windowing, and the memory-only drop case. Each background test uses an isolated `@fs.tmpdir`.

### DIFF
--- a/agent/README.mbt.md
+++ b/agent/README.mbt.md
@@ -127,7 +127,17 @@ async test "standard tools are registered in dispatch order" {
         for tool in tools.function_tools() => tool.name
       ],
       content=(
-        #|["shell", "read", "edit", "multi_edit", "write", "plan", "finish"]
+        #|[
+        #|  "shell",
+        #|  "shell_output",
+        #|  "shell_stop",
+        #|  "read",
+        #|  "edit",
+        #|  "multi_edit",
+        #|  "write",
+        #|  "plan",
+        #|  "finish",
+        #|]
       ),
     )
   }
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -282,7 +282,12 @@ pub async fn[X] run_turn_in_scope(
 ) -> @agent_session.Session {
   let client = @client.Client(api_key~, model~, thinking~, api_url?)
   let session = session |> append_item(User(UserMessage(task)))
-  let tools = tools.unwrap_or_else(() => build_tools(runtime, scope)) catch {
+  let tools = try {
+    match tools {
+      Some(tools) => tools
+      None => build_tools(runtime, scope)
+    }
+  } catch {
     error => {
       @xlog.error() <? { "event": "agent_setup_failed", "error": "\{error}" }
       return session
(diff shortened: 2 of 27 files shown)


bobzhang:
Superseded by the reorganized series: umbrella #389 (complete diff + architecture) and merge slices #399 → #400 → #401 → #402. This PR's 12-round review history drove the hardening now folded into slices #400/#401.

PR #411, opened 2026-07-08
simplify(serve): share the idle-time durable-append path

bobzhang:
`append_background_notice` (added with the L3.5 completion notices) was a line-for-line copy of the older `append_command_context`: same store-or-memory append, same cancellation-aware `catch`, differing only in the session item, the error noun, and the success log line. Extracted a local `append_idle_item(item, what) -> Bool` holding the shared append/catch; each wrapper keeps its own success log and skips it when the durable append fails — behavior unchanged, net −13 lines.

## Testing
`moon check`/`fmt` clean; `cmd/openseek` 59/59.

### DIFF
--- a/cmd/openseek/serve.mbt
+++ b/cmd/openseek/serve.mbt
@@ -274,25 +274,38 @@ async fn run_serve(
       }
     }
 
-    async fn append_command_context(text : String, kind : String) -> Unit {
+    // Append one idle-time item to the session, durably when a store is wired.
+    // Returns false (after logging a session_error naming `what`) when the
+    // durable append fails, so the caller skips its success log; cancellation
+    // still propagates.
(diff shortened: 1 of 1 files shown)


bobzhang:
**Codex CLI review — sign-off** (`codex review --base 97cf98a4`, full PR diff):

> No actionable correctness issues were found in the refactor. The change preserves the previous append-and-log behavior while centralizing shared failure handling.

Per-commit review of `simplify(serve): share the idle-time durable-append path` reached the same verdict: "The refactor preserves the previous append/error-handling behavior while sharing the duplicate logic, and targeted cmd/openseek tests pass."

🤖 Posted by [Claude Code](https://claude.com/claude-code) on behalf of Codex CLI

PR #412, opened 2026-07-08
feat(agent): auto-compaction as a context-ceiling guard

bobzhang:
Implements the auto-compaction item from `improvement.md` as exactly the reframe it prescribes: a **context-ceiling guard** that fires only when the projected prompt approaches the model's context window — where the alternative is request failure — never as a cost optimization (rewriting the prefix busts DeepSeek prefix-cache pricing).

## How it works

- **`Model::context_window()`** (new, `deepseek`): 1M for the DeepSeek V4 variants (per api-docs.deepseek.com), 256K for the kimi-k2.7-code family.
- **Trigger**: after every response, the loop compares the reported `usage.total_tokens` (last prompt + completion — the closest proxy for the next request's prompt) against 80% of the window. Checking post-response means a final-answer step still compacts, so chat-only sessions can't creep past the ceiling turn after turn.
- **Two stages, bounded per turn**: stage one folds only events *before* the current turn's task — the active task, steers, and tool results stay verbatim in context, and the checkpoint request summarizes a rebuilt prefix session so the summary describes exactly the range it covers. If pressure persists (or there's no prior history), one full-range checkpoint (the manual `/compact` semantic) runs, then the guard disarms until the next turn.
- **Persistence**: the summary is validated via the new `Session::summary_item` and appended through the loop's `append_item` seam — the same store-backed path as every other event in serve mode (covered by a store reload test).
- **Fail-open**: a failed summary logs `auto_compaction_failed`, disarms for the turn, and the turn continues — a guardrail must not kill a turn that might still fit; one that truly no longer fits fails visibly through the existing `Terminal(Failed)` path. Cancellation re-raises.
- **Event names** `auto_compaction_started/finished/failed` are deliberately distinct from the manual `compaction_*` family: the TUI decodes `compaction_finished`/`compaction_failed` as **run-terminal**, which a mid-turn auto-compaction is not. The new names drop harmlessly in the TUI decoder and stay visible in engine logs/viz JSONL.
- **Projection ordering fix** (found by codex review of the trigger commit): a stage-one summary is appended *after* the active task but covers only prior turns, so `chat_messages` now emits a surviving summary **at the start of its covered range** instead of at its append position — otherwise the prior-context recap would read as newer input than the task, durably. Full-range summaries project exactly as before; the viz model view's two label walks (index-parallel to `chat_messages`) move in lockstep.

## Test approach

The feature sounds hard to test ("build a 1M-token conversation in CI?") but isn't: the trigger input is usage **reported by the API response**, and the agent tests already run against an in-process mock HTTP server — so the mock fabricates token counts to drive the trigger deterministically, with no live API and no giant contexts. The mock branches SSE (main steps) vs plain JSON (the non-streaming checkpoint call) and the tests assert request `messages` structurally:

- stage one folds prior turns, keeps the active task verbatim (summary projected before it), re-seeds the in-flight request buffer (tool batch stays API-valid around the summary)
- escalation to one full checkpoint, then disarm — exactly two checkpoint calls, ranges asserted
- exact 80% boundary triggers on a final answer (creep guard), summary appended before the terminal
- below threshold: complete silence
- fail-open: empty summary → no Summary event, full history retained, no retry within the turn
- store: a `Summary` appended via generic `SessionStore::append` (the serve `append_item` path) is durable, timestamped, and honored by the projection after reload

`moon test` 970/970, `moon cram test` clean, `moon info && moon fmt` applied. Plan and every commit reviewed with codex CLI (plan review reshaped the design: task preservation, per-turn bound, distinct event names, structural test assertions).

## Out of scope (noted for follow-up)

- TUI/viz surfacing of auto-compaction progress (new decode arm + status line).
- Tool-output capping (todo.md): a single tool result larger than the window can still overflow between checks — compaction can't fix that class.
- Threshold/window config knobs.


bobzhang:
todo: all auto-compat raito to be set dynamically 50% etc

PR #416, opened 2026-07-08
feat: /goal — a durable long-horizon objective

bobzhang:
Adds `/goal` — a durable long-horizon objective the agent stays oriented toward across turns, restarts, and compactions. It completes the existing trio: the **plan tool** orients within a turn, **`/loop`** re-fires prompts across idle gaps, and **`/goal`** is the standing objective both serve. For fully autonomous long tasks the two compose: `/goal set <objective>` + `/loop 60 continue the goal`.

## Usage

- `/goal set <text>` — set or replace the standing goal (explicit `set` keyword, so a goal starting with "clear" stays expressible)
- `/goal` — show it
- `/goal clear` — clear it

Serve engines only (oneshot rejects, like `/compact`).

## Design

- **No session-schema change.** The goal rides `Runtime(RuntimeNotice)` markers — `[goal]\n<text>` / the `[goal cleared]` tombstone — because the session decoder is strict about unknown kinds: old binaries keep loading new logs. `Session::current_goal` scans the raw append-only log backward (last marker wins), so a goal whose set marker was covered by a compaction summary is still recovered verbatim.
- **Model visibility**: `GoalCadence` (parallel to `PlanCadence`, seeded from the session per turn) injects a `[goal reminder]` at turn start — suppressed while no model response follows the set marker, which still rides the next request anyway — and every 30 requests within a long turn.
- **The long-horizon teeth**: a finish attempt (final answer or `finish` tool) with an unconfirmed standing goal draws **one `[goal check]` per turn**: confirm the goal is met, finish with a one-line status of what remains, or keep working. The model may finish again immediately — a turn is never held hostage; a sub-task answer costs at most one extra round. The finish-tool path closes the remaining batch calls first (API-validity preserved; finish-not-last-in-batch tested), and the check counts as a reminder so a boundary-timed reminder can't bury it.
- **Wire + TUI**: `{"command":"goal","action":"set"|"clear"}` on the serve engine. Idle sets append directly; a live turn takes them through the lossless steer queue (model-visible at the next boundary); a pending or running **compaction holds them in the ordered work queue** so the summary never covers a goal that arrived after the compact command. Every durable path — and no queue-time path — emits `goal_updated`, so the TUI's mirrored state only ever reflects persisted goals. The TUI seeds the mirror from the store-loaded session at startup (resume shows the goal before any engine spawns) and routes `goal_updated` untagged so idle confirmations survive the stale-run guard, decoding `goal: null` (clear) distinctly from an empty string.

## Tests

Layered like the rest of the repo, no live API: marker/scan unit tests (set, clear, revive, multiline, compaction-covered markers); mock-HTTP agent-loop tests for turn-start reminder, fresh-set suppression, the 30-request cadence, mid-turn steered sets, the goal check on both finish seams, its once-per-turn cap, and the reminder-boundary interaction; serve decode tests (blank/unknown rejections) and pending-queue helpers; TUI wbtests driving `update()` for the grammar, oneshot rejection, mirror discipline, and null-aware `goal_updated` decode.

Plan and every commit reviewed with codex CLI; its findings shaped the grammar (`set` keyword), the suppression rule, the wire-order handling around compactions, the reminder/check interaction, and the untagged event routing. `moon check --deny-warn`, `moon info && moon fmt`, full `moon test` (989/989) + cram + the JS viz bundle all clean.

## Out of scope (possible follow-ups)

- Auto-continue on goal (compose with `/loop` instead)
- `--goal` CLI flag; goal in the TUI status line; multiline goals via the TUI composer (wire supports them)

---

## Update: structured goal status + engine auto-continue (follow-up commits)

The original "out of scope: auto-continue" is now shipped, redesigned around a structured completion signal (both Claude Code and Codex make harness-actionable transitions tool calls, never prose):

- **`goal` status tool** (`agent_tool/goal`): the model records `met` (durably clears the goal — the loop defers the `[goal cleared]` tombstone to a batch-safe boundary so replay never splits a tool batch) or `continuing` with a required `remaining` summary (which stands in for the finish check). The loop intercepts these calls — only it knows whether a goal stands — so `met` with no goal is an honest tool error. Cadence mutates only after the result append persists.
- **Engine auto-continue** (opt-in `"auto": true` on the goal set wire command): after a cleanly finished turn (`TurnDone` now threads `finished~` from the turn's terminal), a fully idle, still-serving engine with the goal standing starts a continuation turn — 20-turn budget per armed goal, `goal_continue`/`goal_budget_exhausted` events, stopped by the met tombstone, `/goal clear`, any `Cancel` (zeroes the budget), queued user work, or shutdown. Headless-first by design: the TUI never arms auto (engine-initiated turns are invisible to its run-id tracking; interactive users compose `/loop`) and surfaces the lifecycle as notices.

## Live validation

Dogfooded on a real long-horizon goal before the auto-continue commits: a goal-branch engine built a **5,048-line C compiler in MoonBit (95 tests green) in 3 turns / ~100 min / 478 requests**, with every goal mechanism observed working in production — fresh-set suppression, 30-step reminders, turn-start re-surfacing, three goal checks each catching a premature finish (the last forcing a verify-before-declare), mid-turn steer recovery from a 24-minute toolchain hang, and set/clear confirmations. The run's turn-3 "goal is fully met" existed only as prose — exactly the gap the goal tool closes. Findings filed as #419–#423.

### DIFF
--- a/agent/README.mbt.md
+++ b/agent/README.mbt.md
@@ -139,6 +139,7 @@ async test "standard tools are registered in dispatch order" {
         #|  "multi_edit",
         #|  "write",
         #|  "plan",
+        #|  "goal",
         #|  "finish",
         #|]
       ),
--- a/agent/auto_compact.mbt
+++ b/agent/auto_compact.mbt
@@ -173,6 +173,12 @@ async fn AgentLoop::yield_at_context_ceiling(
   session : @agent_session.Session,
   steers : ArrayView[@agent_runtime.SteerInput],
 ) -> @agent_session.Session {
+  // A goal(met) recorded in the turn's final batch defers its durable
+  // [goal cleared] tombstone to a batch-safe boundary; every yield site
+  // has the batch closed, and without this flush the tombstone would be
+  // lost — current_goal() would resurrect a goal the model already met,
+  // and auto-continue would keep chaining toward it.
+  let session = self.flush_goal_tombstone(session)
   let to_sequence = session.last_sequence()
   let (session, compacted) = self.checkpoint_at_ceiling(session, steers)
   // The summary request takes seconds; steers submitted while it was in
--- a/agent/goal.mbt
+++ b/agent/goal.mbt
@@ -1,0 +1,307 @@
+///|
+/// The turn-start reminder is a DECISION POINT, not just an alignment
+/// note: each turn — under auto-continue, each context-window cycle —
+/// opens with a brief three-way assessment. Catching "already met" or
+/// "cannot proceed" here saves a whole window of work; the finish check
+/// still gates the exit so a completion cannot slip out unverified.
+fn goal_reminder_text(goal : String, tool_registered~ : Bool) -> String {
+  let met_action = if tool_registered {
+    "record that by calling the goal tool with status \"met\", then finish"
+  } else {
+    "finish and say so"
+  }
+  (
+    $|\{@agent_session.GoalReminderPrefix}
+    $|The standing goal for this session:
+    $|\{goal}
+    $|Assess it briefly at this turn boundary:
+    $|- Fully met? Verify per the goal's own criteria first, then
+    $|  \{met_action}.
+    $|- Unable to proceed (blocked, impossible, or superseded)? Record that
+    $|  — call the goal tool with status "blocked" and a reason if it is
+    $|  available, otherwise say why — then finish; the user decides what
+    $|  is next.
+    $|- Otherwise continue working toward it. If the current task conflicts
+    $|  with the goal, the current task wins; note the conflict briefly.
+  )
+}
+
+///|
+fn goal_check_text(goal : String, tool_registered~ : Bool) -> String {
+  if tool_registered {
+    (
+      $|\{@agent_session.GoalCheckPrefix}
+      $|The standing goal for this session:
+      $|\{goal}
+      $|You attempted to finish. If the goal is fully met, record that by
+      $|calling the goal tool with status "met", then finish. If this task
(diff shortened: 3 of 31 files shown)


bobzhang:
## Live validation round 2: the full auto-continue lifecycle

Three staged experiments against the final branch build, same engine, real DeepSeek:

**1. Unconstrained goal (C compiler, 5–7K lines target).** The goal tool's mechanics all worked — the model recorded an honest `continuing` ("~3,300 vs 5,000-7,000 target") at the goal check — but 13 minutes later declared `met` at an unchanged 3,279 lines and exited. **Structured status is actionable but not automatically truthful.**

**2. Same goal re-armed with an evidence contract** ("only record met after running the line count and moon test in the same step and quoting both numbers"): the model worked to 5,031 lines / 99 tests, ran both verification commands (visible in the log at 14:13:59 and 14:14:18), and only then recorded `met` (14:14:24). **The contract fixed the honesty gap, n=1 but the contrast is stark** — and it motivates the `--until` external-verification follow-up for quantifiable goals.

**3. Forced chaining (max-steps 6, two-files-per-turn protocol, ten files):** `goal_continue` chained five turns autonomously (budget 19→18→17→16), the model recorded `continuing` with remaining files at each boundary, and on completion `goal(met)` → tombstone → **no further `goal_continue`** — structural termination exactly as designed, 10/10 artifacts on disk.

Also observed across the runs: turn-start reminders, 30-step cadence, fresh-set suppression, three finish-check interceptions, mid-turn steer recovery, and set/clear/met all confirmed by `goal_updated` only after durable appends.

PR #429, opened 2026-07-09
refactor(shell): local simplification + dedupe sweep

bobzhang:
Applies all 13 findings of the shell-family simplification sweep (agent_tool/shell, shell_exec, shell_output, bgjobs, and shell internals). Behavior-identical: error messages, output ordering, and timing budgets are preserved. `moon check --deny-warn` clean, full `moon test` green (1001/1001).

## Dead code
- **shell.mbt**: fold the `detachable` dead branch — the foreground spawner closure only exists inside `bg_runtime.map`, where `detachable` is definitionally true (#424 changed the condition without folding), so it always spawns retained.
- **bgjobs.mbt**: delete the now-unused `BgJobRuntime::spawn_foreground`, its `spawn_foreground_exec` field, and the closure in `new` (same commit as the fold above — the dead else-arm was its only caller; tree re-grepped to confirm). Pub API shrinks by one method; `pkg.generated.mbti` regenerated.

## Dedupe
- **shell.mbt**: one `source_tree_operation_target` helper (direct-then-dynamic short-circuit preserved) replaces four near-identical match blocks in `run_agent_shell_with_tree_precheck` and `start_background`.
- **shell_exec/execution.mbt**: `kill_for_time_limit` and the monitor's output-limit watchdog reuse `request_stop()` instead of hand-inlining `cancel_requested = true; process.cancel()`.
- **shell_exec/sink.mbt**: `discard()` reuses `close()` instead of re-inlining its body.

## Pattern-match idioms
- **shell.mbt**: if-is-Some early returns for the command-policy error and the source-write redirect (drops empty `None => ()` arms; matches the file's own style).
- **shell_output.mbt**: guarded `Stopped if ...` arms replace the nested if/else chain for the status label.
- **command_policy.mbt**: view-pattern scanner for sed flag clusters replaces the chars array + mutable index loop.
- **shell_exec/sink.mbt**: declarative range/tuple match tables for `utf8_sequence_len` and `continuation_range`.
- **decode.mbt**: single-use `optional_positive_int_or_default` wrapper inlined into its one caller.

## Test-helper dedupe
- **shell/shell_test.mbt**: one `exec_shell` helper behind the four `run_shell*` wrappers (which become one-liners) and the two inline Async/Respond unwrap blocks.
- **shell_output/shell_output_test.mbt**: `run_shell_output` generalized to take Json arguments, absorbing the four re-inlined `wait_ms` unwrap blocks.
- **bgjobs/bgjobs_wbtest.mbt**: `wait_for_first` helper for the three notice poll loops (identical 200 x 25ms budget; the on_job_exit test keeps its trailing sleep).

## Skips
None — all 13 findings applied. The reviewer's verified non-findings (wait/wait_or_invalid merge, effective_timeout_ms nested match, dual JSON schema literals, git_policy mirrors, lexer state machine, sandbox predicate catalogs) were deliberately not touched.

### DIFF
--- a/agent_tool/bgjobs/bgjobs.mbt
+++ b/agent_tool/bgjobs/bgjobs.mbt
@@ -43,11 +43,6 @@ pub struct BgJobRuntime {
   // inline-preview budget, spill path) -> execution. Captured so the runtime type
   // is not generic.
   priv spawn_exec : async (String, Array[String], String?, Int, String?) -> @shell_exec.ShellExecution
-  // Spawn a *foreground* execution on the same session group: memory-only output
-  // capped at the budget and killed if it fills, so the shell tool's foreground
-  // path shares the execution model (and can be detached on timeout) without
-  // re-deriving generic spawning at each call site.
-  priv spawn_foreground_exec : async (String, Array[String], String?, Int) -> @shell_exec.ShellExecution
   // Spawn a *retained* foreground execution (may be detached on timeout):
   // file-backed with a large hard cap, but still killed on exceeding the inline
   // cap so a timed command that floods output is an immediate output-limit error;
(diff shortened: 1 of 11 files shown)


bobzhang:

`codex review --base main` (codex-cli 0.143.0), run in the PR worktree after all commits:

> No actionable patch-introduced regressions were found in the changed code. Targeted shell/background-job tests pass; full-suite failures observed were external-network smoke tests unrelated to this diff.

Validation alongside the review: `moon fmt` clean, `moon check --deny-warn` clean, `moon test` 1001/1001, `moon info` mbti diff limited to the intended `spawn_foreground` pub-API removal.

PR #431, opened 2026-07-09
refactor(agent): local simplification + dedupe sweep

bobzhang:
One PR of the repo-wide simplification sweep, covering `agent`, `agent_review`, and `agent_skill`. All 19 reviewer findings applied; behavior-identical by construction (no public API change — `pkg.generated.mbti` files are untouched by `moon info`).

## turn_loop dedupe
- Fold the 3 copy-pasted append-Assistant + mirror-push blocks (steer-beats-answer, hard ceiling, `handle_tool_calls`) into a file-local `AgentLoop::record_assistant`; the optional `tool_calls` forwards verbatim.
- `salvage_finish_at_ceiling`: replace the `-1`-sentinel index loop with `response.tool_calls[from:].search_by(...)` on a zero-copy slice.
- `finish_turn` buffer re-seed: `messages.append(session.chat_messages())` instead of a push loop.
- `auto_compact`: `at_context_share` becomes a token-based `AgentLoop` method shared by both ceiling tiers — bit-for-bit the same comparisons, removing the soft tier's inline re-implementation.
- Inline the single-use, cross-file `print_usage` at its one call site.

## auto_compact test dedupe
- Reuse the shared `bind_test_server` bind-or-skip helper (the file postdated it and missed it).
- `sse_tool_call` delegates to `sse_tool_calls` (byte-identical singleton payload).
- `flood_definition(chars)` / `finish_definition()` shared across the 3 flood registries and 2 finish registries; registration order unchanged; the clamp test reuses `flood_tools(70_000)` (its assertions never reference the fill char).
- `finished_answer` / `summary_before_terminal` helpers replace the ~20 repeated 5-line terminal/summary guard blocks; `loc~` autofill keeps failures pointing at the test line.
- Drop a dead `ignore(result)` two lines before `result` is used.

## idiom rewrites
- Comprehensions for `sse_tool_calls` entries and `flood_batch`; `String::repeat` for flood strings.
- `plan_reminder_test`: merge `stale_plan_test_server` / `recorded_plan_test_server` into one `plan_test_server(record_first~)` — the scripted responses stay byte-identical per request number; `reminder_count` uses `Iter::count` instead of materializing the split.
- `agent_skill/parse_frontmatter`: drop the per-line `to_owned` copy; trim/`split_once`/interpolation already operate on views.
- `agent_review/engine`: `let _ : @agent_session.Session = ...` instead of bind-then-`ignore`.

## README fixes
- The `mbt nocheck` API fence used positional arguments that would not compile and an `@agent.steer` function that does not exist: now uses the labeled arguments verified against `agent/pkg.generated.mbti` and `runtime.queue_steer(Prompt(...))`; the steering prose updated to match.
- Corrected the stale claim that no tools use the runtime/scope — the shell tools use both since the background-jobs series (steer notices via `queue_steer`; `BgJobRuntime` + spill-dir cleanup owned by the scope's group).

## Skips
None — all 19 findings applied. The known trap was honored: `log_tool_result`'s brief `if/else Json::string/null` encoding is untouched (core encodes `Option` ToJson as `Some(v) => [v]`, so `brief.to_json()` would change the wire format).

## Validation
- `moon fmt` (clean; it normalized `loc~ : SourceLoc = _` to the `#callsite(autofill(loc))` form)
- `moon check --deny-warn`: 0 warnings, 0 errors
- `moon test`: 1001/1001 passed
- `moon info`: no `.mbti` changes (public surface unchanged)

### DIFF
--- a/agent/README.mbt.md
+++ b/agent/README.mbt.md
@@ -22,17 +22,17 @@ The exported surface is small:
 
 ```mbt nocheck
 @agent.run(
-  api_key,
-  Deepseek(V4Pro),
-  "fix the tests",
+  api_key~,
+  model=Deepseek(V4Pro),
+  task="fix the tests",
   system_prompt_text="You are an OpenSeek agent.",
 )
 
 let persisted = @agent.run_turn_with_append(
-  api_key,
-  Deepseek(V4Pro),
-  session,
-  "continue",
+  api_key~,
+  model=Deepseek(V4Pro),
+  session~,
+  task="continue",
   append_item=(session, item) => store.append(session, item),
 )
 
@@ -41,18 +41,18 @@ let persisted = @agent.run_turn_with_append(
   let scope = @agent_runtime.AgentTaskScope(group)
   let tools = @agent.build_tools(runtime, scope)
   let next = @agent.run_turn_in_scope(
-    runtime,
-    scope,
-    api_key,
-    Deepseek(V4Pro),
-    session,
-    "continue",
+    runtime~,
+    scope~,
+    api_key~,
+    model=Deepseek(V4Pro),
+    session~,
+    task="continue",
     append_item=(session, item) => session.append(item),
     tools~,
   )
 }
 
-@agent.steer(runtime, "also update README")
+runtime.queue_steer(Prompt("also update README"))
 ```
 
 `run` is the highest-level one-shot entry point. It creates a fresh in-memory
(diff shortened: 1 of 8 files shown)


bobzhang:

> No actionable regressions were found in the diff. `moon check` and `moon check --target all` pass; `moon test` only failed on real-world API smoke tests due restricted DNS/network access in this environment.

Validation alongside the review: `moon fmt` clean, `moon check --deny-warn` 0 warnings / 0 errors, `moon test` 1001/1001 passed locally (native), `moon info` produced no `.mbti` diffs — the public API surface is unchanged.

PR #434, opened 2026-07-09
refactor(agent_tool): edit/multi_edit/write dedupe + find() efficiency sweep

bobzhang:
Behavior-identical simplification sweep over the file-editing tools (`edit`, `multi_edit`, `write`), applying the file-tools subset of the agent-tool-core findings. A sibling PR handles the rest of that findings group.

## Applied findings

**Dedupe**
- `write/write.mbt`: deleted `manifest_write_error` — a 39-line verbatim copy of `@manifest_error.write_error` — and call the internal package instead (one import line added to `write/moon.pkg`; the now-unused `manifest_path` import removed). Also deleted the duplicated Windows-path test — `manifest_error_test.mbt` covers it verbatim; the execute-level manifest tests remain.
- `edit.mbt` / `multi_edit.mbt`: collapsed the duplicated `line_start_offset` scanner into `line_end_offset(content, line - 1)` for `line >= 2`. Verified identical on both loop pairs, including the `nobreak => length` past-EOF case.
- `edit.mbt`: the five identical 5-line respond-error scaffolds in `edit_file` now go through a local `err(message)` helper; messages are byte-identical.
- `edit_test.mbt` / `multi_edit_test.mbt`: `run_edit` / `run_multi_edit` now delegate to their `run_*_in_workspace(".", ...)` twins (`definition()` defaults `workspace_root` to `"."`).
- `edit/internal/decode/decode_test.mbt` / `multi_edit/internal/decode/decode_test.mbt`: adopted plan decode_test's `assert_decode_error(arguments, expected)` helper pattern (`assert_decode_args_error` in multi_edit) for the repeated `try/catch/noraise` scaffolds.

**Efficiency**
- `edit.mbt` / `multi_edit.mbt`: the first-occurrence scan no longer materializes every fragment via `split(...).map(to_owned).collect()` — `region.body.find(old_string)` gives the same first-occurrence offset. Error messages and offsets are unchanged; `old_string` is guaranteed non-empty at both sites (the insertion path returns earlier). The `replacement_count = 1` constant is gone (interpolated literal `1`).

**Idiom / simplify (multi_edit.mbt)**
- `compare_plan_ascending`: 10-line if-chain → tuple `Compare` (`(a.start, a.end, a.index).compare(...)`); the ascending/descending pair stays symmetric.
- resolved-edit build loop → two-var comprehension.
- `check_contiguity`: `match`-to-Bool → `current is Some(path) && path == edit.path`; `ended.get(...) is Some(_)` → `ended.contains(...)`.
- `total_edit_count` and the rejection total → `fold(init=0, ...)`; the truncated flag → `iter().any(...)`.
- shown-rejections cap → `rejections.length().min(RejectedFileDisplayLimit)`.
- `multi_edit/internal/decode`: identity comprehension over `fields.keys()` → `fields.keys().collect()`.

## Skips (within the multi_edit decode_test scope)
- The "decode lists present keys when a required field is misnamed" scaffold kept its explicit `try/catch`: it makes four compound assertions (including an `assert_false`) that the single-expectation helper cannot express.
- The single `decode_edits_json` reject scaffold kept as-is: one use, different entry point — a dedicated helper would be a net add.

Cross-package dedupes flagged in the findings notes (edit vs multi_edit gate cores, shared line helpers, decode field helpers) are intentionally NOT done — they grow pub API surface and are left for a deliberate decision.

## Validation
- `moon fmt` — clean
- `moon check --deny-warn` — clean
- `moon test` — 1000/1000 passed
- `moon info` — no `.mbti` changes (public API untouched)

### DIFF
--- a/agent_tool/edit/edit.mbt
+++ b/agent_tool/edit/edit.mbt
@@ -79,35 +79,28 @@ async fn edit_file(
   // Every failure names the file it tried to edit so the viz can show
   // `→ edit · edit <file> 🚩` on the folded call line.
   let fail_brief = "edit \{@agent_tool.brief_basename(path)}"
-  if input.replace_all {
-    return @agent_tool.ToolAction::respond(
-      "error editing \{path}: replace_all=true is no longer supported; use multi_edit with explicit line-anchored edits",
+  fn err(message : String) -> @agent_tool.ToolAction {
+    @agent_tool.ToolAction::respond(
+      "error editing \{path}: \{message}",
       is_error=true,
       brief=fail_brief,
+    )
+  }
+
+  if input.replace_all {
+    return err(
+      "replace_all=true is no longer supported; use multi_edit with explicit line-anchored edits",
     )
   }
   if input.old_string == "" && input.new_string == "" {
-    return @agent_tool.ToolAction::respond(
-      "error editing \{path}: old_string and new_string are both empty",
-      is_error=true,
-      brief=fail_brief,
-    )
+    return err("old_string and new_string are both empty")
   }
   if input.old_string == input.new_string {
-    return @agent_tool.ToolAction::respond(
-      "error editing \{path}: old_string and new_string are identical",
-      is_error=true,
-      brief=fail_brief,
-    )
+    return err("old_string and new_string are identical")
   }
   let content = @fs.read_file(path).text() catch {
     error if @async.is_being_cancelled() => raise error
-    error =>
-      return @agent_tool.ToolAction::respond(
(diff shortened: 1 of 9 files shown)


bobzhang:

> No diff-introduced correctness issues were found in the reviewed changes. Targeted tests for the changed tool packages passed; full test failure was limited to network-dependent smoke tests unrelated to this patch.

Validation alongside the review: `moon fmt` clean, `moon check --deny-warn` clean, `moon test` 1000/1000 passed, `moon info` produced no `.mbti` diffs (public API unchanged).

PR #457, opened 2026-07-11
feat(viz): dedicated plan rendering with progress, ribbon, and deltas

bobzhang:
## What

`plan` tool calls previously rendered through the generic key/value argument rows — a wall of `title:` / `status:` lines. This gives plans a dedicated rendering in both views:

- **Folded summary line**: a per-step status ribbon (one colored segment per step, step title + status on hover), so scrolling a long session shows the plan filling up without unfolding a single card.
- **Expanded card**: a progress meter (proportional fill + `completed/total` tally; fill turns green only when the plan is done) above a status-classed checklist with ✓/◐/○ icons.
- **Delta chips**: each update diffs against the previous plan call in the same view — steps that changed get a `done`/`started`/`new`/`reopened` chip and silently dropped steps stay visible as ghost rows. Diffing only engages when at least half the new plan's titles carry over (an *evolution*); a wholesale rewrite gets a single "rewritten — replaces the previous plan" note instead of chip noise (validated against the ccomp-run6 log: exactly 3 done + 2 started chips and 4 rewrite notes).
- **Plan-cadence reminders** reuse the same checklist rows and progress meter, replacing the old `[status]` text chips.
- Malformed plan arguments keep the generic rendering; the raw JSON stays one Original-toggle away; the model view shares all of it (baselines drawn from projected messages only).

Also:
- Theme now defaults to **System** (follow the OS) instead of Light.
- A standalone export containing exactly one session **auto-opens** it, so exported archives land on the log instead of an empty pane (live servers keep the explicit click).

## Verification

- 233 js + 1048 native tests pass; `moon check --deny-warn` clean on both targets; new tests cover checklist/ribbon/progress markup, delta chips, ghost rows, rewrite back-off, cleared/malformed fallbacks, and export auto-open.
- Re-exported the 5.6 MB ccomp-run6 session and drove it in headless Chrome: all 8 plan calls render ribbons/checklists, chip counts match a hand-computed diff of the session's plan arguments, both themes eyeballed via screenshots.


bobzhang:
Two follow-up commits since the opening description:

**`0dae2979` — review hardening.** A workflow-backed high-effort review (subal's backend was down) confirmed 7 bugs in the diff logic; all fixed with regression tests: tool-rejected plan calls and empty/duplicate call ids no longer fabricate diff baselines; `plan_evolves` measures carry-over against the smaller plan so pure extensions aren't branded "rewritten"; completed→in_progress now chips "reopened" (and in_progress→pending "paused"); the reminder meter parses the leading `[status]` bracket instead of substring-matching; unknown statuses render literally; the step-number assertion is restored and the three row shapes share one `plan_step_row` builder.

**`1390689f` — shared PlanStep + unfolded plan cards.** The plan tool's decoder is promoted from `agent_tool/plan/internal/decode` to a public `agent_tool/plan/steps` package (imports only `core/json`, so js-only viz can use it while the tool package stays native-only). Viz drops its own `PlanStep` struct and lenient decoder for the tool's strict one — viewer and tool can no longer drift on what a valid plan is — and `PlanStatus::label()` becomes the single wire spelling shared by decode, the reminder's `[status]` brackets, and the viewer's CSS classes. Plan cards also render `<details open>`: a plan is the task's live state, not noisy arguments.

Verified on the ccomp-run6 export: the tool-rejected call at seq 524 (122-char title) now renders generic args matching the tool's verdict, the 7 valid plan calls are exactly the 7 unfolded tool-call cards, and chip counts still match a hand-computed diff.

PR #460, opened 2026-07-11
feat(edit): report actual inclusive line range for inserts and appends

bobzhang:
## What

Insertion summaries in `edit` used to echo the **requested** `start_line`, so appending with a past-EOF anchor (which clamps to end of file) reported a line that does not exist — e.g. `ok: inserted at line 999999` — giving the model useless coordinates for follow-up edits.

Insertions now report the **actual inclusive line range** the new content occupies, in the same 1-based inclusive coordinate system that `start_line`/`end_line` accept, so the reported numbers round-trip into follow-up anchored edits with zero arithmetic:

- mid-file: `ok: inserted line 42 in <path>` / `ok: inserted lines 42-44 in <path>`
- clamped/EOF: `ok: appended lines 3-4 at end of file in <path>` — doubling as confirmation that the past-EOF anchor was clamped as intended

The viz brief carries the same range (`edited edit.mbt:3-4 (append)`). A new `inserted_line_span` helper counts occupied lines, honestly handling the no-trailing-newline glue case (appending `c\n` to `a\nb` reports `appended line 2`, since the fragment shares line 2).

## Why bless the append idiom

MoonBit top-level order does not matter, so new functions/tests/types can always be **appended at end of file** — a zero-conflict edit with no anchor to mismatch and no context block to go stale (most observed edit misses come from mid-file anchored matching). This PR routes the model toward that idiom:

- `edit` / `multi_edit` descriptions: any `start_line` past the last line appends (e.g. `999999` — no need to know the file's length); append new top-level code, reserve mid-file insertion for grouping.
- `default_prompt.mbt.md` (+ regenerated `.mbt`): new tool-routing bullet with the same guidance.

`multi_edit` needed no reporting change — its success summary is per-file counts only.

## Verification

- `moon check --deny-warn` clean; `moon fmt` applied.
- All 83 `agent_tool/edit` + `agent_tool/multi_edit` + `prompt` tests pass, including four new ones: single-line insert, multi-line insert range, multi-line clamped append, and the glued no-trailing-newline append.
- `subal review --base a26456b0`: "The insertion and append summaries correctly calculate actual inclusive line ranges, including EOF and missing-trailing-newline cases. Targeted tests and MoonBit checks pass; full-suite failures were unrelated network-dependent smoke tests." (The 4 full-suite failures are the network-dependent realworld API smoke tests, expected offline.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #461, opened 2026-07-11
docs(prompt): moon run -e cannot test local packages; bless #skip/#cfg(false)

bobzhang:
## Verified behavior (all claims tested against the current toolchain)

**`moon run -e` imports are registry-only.** Snippets may open with a multi-line `import { ... }` block, but resolution goes through the mooncakes.io registry index — never the local workspace:

- An unpublished local module fails outright: `module 'scratch/localtest' not found ... must be resolvable from local registry index`.
- The insidious case: if the module **is** published (as `bobzhang/openseek` is), the import **silently binds the published snapshot**, not the working tree. Proven: the cached published 0.2.2 contains zero occurrences of a symbol added locally the same day (7 occurrences in the working tree). An agent "testing" local code via `-e` runs stale code with no error.

**`#skip("reason")`** — body **is** type-checked (a type error under it fails `moon check`); not run by plain `moon test`; **not run by `--filter` alone** ("no test entry found"); runs only with `--include-skipped` (or `-i/--index`, which implies it).

**`#cfg(false)`** — body must still **parse** (a parse error under it fails `moon check`) but is **not type-checked** (a type error under it passes) and never runs: a structured alternative to commenting code out.

## Prompt changes

- `moon test` bullet: black-box `_test.mbt` + `--filter` is THE way to exercise local package code; documents `#skip` (with the `--include-skipped` caveat) and `#cfg(false)`.
- `moon run -e` bullet: imports resolve only from the registry; names both failure modes (module-not-found, stale published snapshot) and routes to `moon test --filter`.

## Verification

- `moon check --deny-warn` clean; `moon test prompt` 19/19; generated prompt regenerated via the `md_to_mbt_string` rule.
- Branch rebased onto current main (includes the merged #460 append bullet; sections are disjoint).

### DIFF
--- a/prompt/default_prompt.mbt.md
+++ b/prompt/default_prompt.mbt.md
@@ -112,7 +112,7 @@ Common `moon` subcommands:
   module fails with `module not found`, and if the module IS published the
   import silently binds the STALE published snapshot instead of your working
   tree. To exercise local package code, write a black-box test and run
-  `moon test <pkg> --filter` instead.
+  `moon test <pkg> --filter '<name glob>'` instead.
   `moon run -e` defaults to native on current MoonBit nightlies, but `moon run -`
   can still default to wasm-gc; pass `--target native` when stdin snippets need
   native or async support.
--- a/prompt/default_prompt.mbt.md
+++ b/prompt/default_prompt.mbt.md
@@ -92,13 +92,27 @@ Common `moon` subcommands:
   `--target` and `--diagnostic-limit <N>`.
 - shell `moon test`: targeted or full tests; run plain `moon test` before
   `moon test --update`. Example: `moon test parser --filter "Parser::*"
-  --diagnostic-limit 5`. Filters support glob syntax.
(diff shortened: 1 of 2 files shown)


bobzhang:
CLI review (`subal review --base origin/main`, at de92567d):

> The prompt recommends an invalid `moon test` command because the required filter argument is omitted.
>
> - [P2] Pass a pattern to `--filter` — `prompt/default_prompt.mbt.md:115`: `moon test <pkg> --filter` exits with code 2 without running any tests. Include a test-name or glob placeholder so the newly prescribed local-package workflow is executable.

Addressed in 6d3707fc: the example now reads `moon test <pkg> --filter '<name glob>'`. No other findings; `moon check --deny-warn` clean and prompt tests 19/19 after the fix.

bobzhang:
CLI review (`subal review --base de92567d`, at 6d3707fc):

> The updated example now supplies the required filter pattern, and the generated prompt remains synchronized with its source. Targeted checks and tests pass.

Both commits now have clean subal sign-offs.

PR #462, opened 2026-07-11
refactor(viz): dedup render fixtures and fold error-count loops

bobzhang:
Scoped simplification/dedup of the `viz` package, which grew ~+520 test lines across the plan-rendering (#457) and step-scrubber (#459) work. Private helpers only — **no public API change** (`pkg.generated.mbti` unchanged). Net **−107 lines**.

## Test dedup (viz_test.mbt)
- **`render_tool_call(name, arguments)`** — the "single tool call → RawLog" fixture was copied verbatim across **9 sites** (session build + render), differing only in name/arguments. The session id and call id never appear in any of these assertions, so they are fixed inside the helper.
- **`two_plan_calls_raw(before, after)`** — the two-successive-`plan`-call fixture the plan-diff tests (rewrite note, delta chips) share, across **3 sites**. Does not subsume the 3-plan-call tests (rejected-baseline / duplicate-id), which keep their own shape.
- **`compact_second_event(session)`** — the identical `compact(content="covered", from_sequence=2, to_sequence=2)` across **3 sites**, named for intent.

## Product code (view.mbt) — hand-rolled loops → core combinators, behavior unchanged
- `count_turn_tool_errors` and both `rendered_error_count` branches → `count_if`. The `events()` branch keeps its `.iter()` hop (Vector has no `count_if`); `chat_messages()` is an `Array` and drops it.
- `plan_evolves` / `dropped_steps` → `Array::any` (drop the redundant `.iter()`).

## Verification
- `moon check viz --target js --deny-warn` clean; `moon test viz --target js` **53/53**.
- `moon info` → `pkg.generated.mbti` unchanged (no API surface change).
- `moon build cmd/viz_app --target js` (the shipped bundle, which compiles view.mbt) still builds.
- `moon fmt` applied.
- `subal review --base origin/main`: *"The production refactors preserve existing iteration and counting semantics, while the new test helpers reproduce the prior fixtures. The targeted viz test suite passes."*

### DIFF
--- a/viz/view.mbt
+++ b/viz/view.mbt
@@ -228,13 +228,7 @@ fn turn_block(
 ///|
 /// Number of failed tool results in a turn, for the turn header's error badge.
 fn count_turn_tool_errors(turn : Array[@agent_session.SessionEvent]) -> Int {
-  for event in turn; count = 0 {
-    if event.item() is Tool(result) && result.is_error() {
-      continue count + 1
-    }
-  } nobreak {
-    count
-  }
+  turn.count_if(event => event.item() is Tool(result) && result.is_error())
 }
 
 ///|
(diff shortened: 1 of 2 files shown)


bobzhang:
CLI review (`subal review --base origin/main`, at 1b9b00c5):

> The production refactors preserve existing iteration and counting semantics, while the new test helpers reproduce the prior fixtures. The targeted viz test suite passes.

No findings. (`viz` is a JS-only package, so subal's incidental `moon test --target native` attempt is expected to no-op; `moon test viz --target js` is 53/53.)

PR #465, opened 2026-07-12
feat(agent_tool): record per-session file provenance in a FileStateMap

bobzhang:
**PR 1 of 2.** Foundational state layer for a future `remove` tool: a per-session record of how the agent has touched files, so `remove` can later tell a file the agent **created this session** (safe to delete — removing it only undoes the agent's own work) from a pre-existing one. No behavior change — pure instrumentation.

## What

- **`FileState` (`Created` | `Modified`)** and **`FileStateMap`**, a thin handle over a shared mutable `Map[path, FileState]`. `Map` is a reference type, so the shared handle observes every tool's updates with **no `Ref`**. `record_created` always wins; `record_modified` never downgrades a `Created` entry (editing a file the agent itself created keeps that provenance). It lives in `agent_tool`, which the file tools already import — zero new package deps.
- **`write` / `edit` / `multi_edit`** gain an optional `file_state?` parameter (defaulting to a fresh empty map, so **every existing call site and test is unchanged**) and record on a successful write: `write` marks `Created` for a new file / `Modified` for an overwrite; `edit`/`multi_edit` mark `Modified`.
- **`build_tools`** constructs one map per session and shares it across those three tools — the same per-session wiring `bg_runtime` already uses, **not a global**: `serve.mbt` and `concurrent.mbt` run multiple sessions per process, and async tests run in parallel, so a global would cross-contaminate provenance (a correctness bug for a delete gate, not just test flakiness).

## Design notes

- **In-memory only**: it reflects this process's session, not history restored from a durable log. Consumers must read "absent" as **"unknown"**, never "not created" — the PR-2 `remove` gate will pair this with a git-tracked fallback so a resumed session still deletes safely.
- `multi_edit` records `Modified` per written path even if a later auto-revert restores content: the map tracks tool writes, and only the `Created`-preservation behavior is load-bearing (the gate keys on `Created`, which `multi_edit` never produces).

## Verification

- `moon check --deny-warn --target native` clean across the whole workspace (no downstream breakage).
- `moon test` green: agent_tool + write + edit + multi_edit **111/111**, agent (build_tools) **67/67**. New tests cover the map merge logic (unit) and end-to-end recording through `write` (Created on create, preserved on overwrite, Modified on pre-existing) and `edit` (Modified).
- `moon info` regenerated the 4 `.mbti` files; the diff is purely additive (`FileState`, `FileStateMap`, and the `file_state?` param).
- subal sign-off posted below.

### DIFF
--- a/agent/tool_definition.mbt
+++ b/agent/tool_definition.mbt
@@ -121,13 +121,18 @@ pub async fn[X] build_tools(
   on_background_wake? : () -> Unit,
 ) -> @agent_tool.Tools {
   let workspace_root = runtime.workspace_root()
+  // One session-scoped file-state map, shared by the file-writing tools so they
+  // record which paths the agent created vs. modified this session — the record
+  // a future `remove` reads to tell an agent-created file (safe to delete) from
+  // a pre-existing one.
+  let file_state = @agent_tool.FileStateMap::new()
   Tools(
     shell_tools(runtime, scope, on_background_wake?) +
     [
       @read.definition(workspace_root~),
-      @edit.definition(workspace_root~),
-      @multi_edit.definition(workspace_root~),
-      @write.definition(workspace_root~),
+      @edit.definition(workspace_root~, file_state~),
+      @multi_edit.definition(workspace_root~, file_state~),
+      @write.definition(workspace_root~, file_state~),
       @plan.definition(),
       @goal.definition(),
       @finish.definition(),
--- a/agent_tool/edit/edit.mbt
+++ b/agent_tool/edit/edit.mbt
@@ -33,6 +33,7 @@
 ///   invalid arguments.
 pub fn definition(
   workspace_root? : String = ".",
+  file_state? : @agent_tool.FileStateMap = @agent_tool.FileStateMap::new(),
 ) -> @agent_tool.AgentToolDefinition {
   AgentToolDefinition(
     name="edit",
@@ -53,13 +54,16 @@ pub fn definition(
       },
       "required": ["path", "old_string", "new_string", "start_line"],
     }),
-    execute=Async(arguments => execute_with_workspace(workspace_root, arguments)),
+    execute=Async(arguments => {
+      execute_with_workspace(workspace_root, file_state, arguments)
+    }),
   )
 }
 
 ///|
 async fn execute_with_workspace(
   workspace_root : String,
+  file_state : @agent_tool.FileStateMap,
   arguments : Json,
 ) -> @agent_tool.ToolAction {
   let input = @decode.decode(arguments) catch {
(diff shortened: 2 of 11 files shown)


bobzhang:
CLI review (`subal review --base origin/main`, at 27ba252f) returned two findings — both legitimate; addressed in d3f428cc:

**[P2] Keys weren't normalized across equivalent spellings** — confirmed real: `workspace_path.resolve` returns paths verbatim under the default `.` root (`resolve(".", "src/main.mbt") == "src/main.mbt"`), so `note.txt` and `./note.txt` became distinct keys for one file, and a modify/lookup through a second spelling missed the entry. **Fix:** `record_*`/`get` now key on the lexically normalized path (`./a`, `a//b`, `x/../y` collapse) via `moonbitlang/x/path`, with a test. Lexical only — symlink aliases stay a consumer concern (realpath at delete time), documented.

**[P1] `Created` treated as a permanent deletion authority** — valid: a recorded path can be rebound afterward (the sandbox still allows `mv`/`rm` on non-source files; a session spans turns), so `Created` proves only that the agent once created *a* file at this path. **Fix:** reframed the docs — `FileState` is an *advisory* signal, and a deletion consumer must revalidate file identity at delete time and combine it with independent guards (git-tracked / soft-delete), treating `Created` as necessary, not sufficient. The safety *enforcement* is the `remove` tool's job (PR 2: `created-OR-git-tracked` gate + realpath revalidation + soft-delete into a session trash), which is exactly where re-stat naturally happens.

No public API change (`pkg.generated.mbti` unchanged); `moon check --deny-warn` clean, 112/112 tests. Re-review of the fix commit is running.

_(Note: subal's token was returning `HTTP 401` intermittently this session; the verdict above did land on a retry.)_

bobzhang:
Two more subal rounds on the key-normalization, each a real narrowing bug — both fixed:

- **Symlink-unsound `..` collapse** (in the P2 fix): lexically resolving `..` merges different filesystem objects when a preceding component is a symlink (`link/../victim` touches `/outside/victim` but keyed as `victim`), falsely marking a plain workspace file as agent-created. → Fixed to preserve `..`.
- **`..` substring false-positive**: guarding on `contains("..")` mis-fired on legitimate dotted filenames (`snap..bak`), so `./snap..bak` and `snap..bak` keyed differently and the created file read as untouched. → Fixed to a **component-based** normalize: split on `/`, drop `.`/empty components, keep every other component including one that is *exactly* `..`. Self-contained, so the `moonbitlang/x/path` dep is dropped.

Net: `normalize_key` unifies equivalent spellings (`a`/`./a`/`a//b`) while never merging across a real `..` segment and never mis-normalizing ordinary dotted names. `moon check --deny-warn` clean, 114/114 tests, mbti unchanged. Final re-review running.

bobzhang:
Final review after the pivot to verbatim keys (`subal review` on the last commit) — **clean**:

> Verbatim keying is applied consistently across all map operations, and the updated tests match the deliberately conservative semantics. No actionable correctness issues were found.

Summary of the normalization arc: subal's original P2 asked the map to unify equivalent path spellings; four lexical-normalization attempts each misattributed provenance for some supported path form (symlink `..`, dotted names like `snap..bak`, native Windows separators, the POSIX `//` root) — every one a *false `Created`*, strictly worse for a deletion gate than a missed spelling. The map now keys **verbatim**: a safe under-approximation (an unrecognized spelling reads `None`, never a fabricated `Created`), with OS-level canonicalization (realpath) deferred to PR 2's `remove` at delete time, where the file is live and realpath pairs with the identity revalidation that confirms provenance anyway.

All five commits now carry clean subal sign-offs. `moon check --deny-warn` clean, 112/112 tests, mbti additive-only.

PR #472, opened 2026-07-12
feat(agent_tool): add the remove tool for agent-created source files

bobzhang:
**PR 2 of 2** (builds on the `FileStateMap` from #465). Completes the tool-mediated file API: `read`/`write`/`edit` existed, but **delete did not** — the shell sandbox blocks `rm` on source paths, so there was no in-workflow way to delete a source file.

## What

`remove` deletes a `.mbt`/`.mbt.md` file **only** when the session's `FileStateMap` records it as `Created` — i.e. one the agent itself created this session, so removing it merely undoes the agent's own work. It refuses:
- a pre-existing source file (never created, or only `Modified`),
- a non-source path (→ delete with `shell`),
- a missing file or a directory.

## Why the `Created` gate is sound without soft-delete

The sandbox **blocks the shell from writing or moving onto a source path** (`SandboxSourceWriteFeedback`, `source_tree_mv_target`), so a source path recorded `Created` **cannot be rebound** to a different file after `write` recorded it — the rebind attack that would otherwise defeat a path-keyed provenance signal is closed for source files specifically. The map is keyed by the same resolved path `write` used, so an unrecognized spelling reads as unknown and is conservatively refused (never a false delete).

## Integration

- Wired into `build_tools`, sharing the session `file_state` with `write`/`edit`/`multi_edit`.
- Registered in the dispatch registry (registry tests + the ordered README doc-test updated).
- Taught in the default prompt's tool-routing section.
- 8 tests: create→remove, created-then-edited still removable, refuse pre-existing / modified-only / non-source / missing / directory, and arg validation.

## Deferred (noted for follow-up)

A git-tracked fallback (to allow removing pre-existing *tracked* files during refactors — recoverable via git), soft-delete into a session trash, and manifest/`.mbti` scope.

## Verification

`moon check --deny-warn --target native` clean across the workspace; `moon test` green (remove **8/8**, agent registry + prompt included, **94/94** in that set); `moon info` mbti is additive-only; `moon fmt` applied; prompt regenerated via the build rule.

### DIFF
--- a/agent/README.mbt.md
+++ b/agent/README.mbt.md
@@ -138,6 +138,7 @@ async test "standard tools are registered in dispatch order" {
         #|  "edit",
         #|  "multi_edit",
         #|  "write",
+        #|  "remove",
         #|  "plan",
         #|  "goal",
         #|  "finish",
--- a/agent/moon.pkg
+++ b/agent/moon.pkg
@@ -12,6 +12,7 @@ import {
   "bobzhang/openseek/agent_tool/multi_edit",
   "bobzhang/openseek/agent_tool/plan",
   "bobzhang/openseek/agent_tool/read",
+  "bobzhang/openseek/agent_tool/remove",
   "bobzhang/openseek/agent_tool/shell",
   "bobzhang/openseek/agent_tool/write",
   "bobzhang/openseek/deepseek",
--- a/agent/tool_definition.mbt
+++ b/agent/tool_definition.mbt
@@ -102,7 +102,8 @@ async fn[X] shell_tools(
 ///
 /// The returned registry contains the built-in local tools: `shell` (plus
 /// `shell_output` and `shell_stop` where background jobs are supported — see
-/// `shell_tools`), `read`, `edit`, `multi_edit`, `write`, `plan`, and `finish`.
+/// `shell_tools`), `read`, `edit`, `multi_edit`, `write`, `remove`, `plan`, and
+/// `finish`.
 /// Each tool lives in its own subpackage of `agent_tool` and exposes a
 /// `definition()` constructor; this function bundles them into the `Tools`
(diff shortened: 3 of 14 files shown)


bobzhang:
`subal review` of the first commit found two real issues — both fixed in the follow-up:

**[P1] `Created` provenance outlived the file.** After `remove` deleted a path, the map entry stayed `Created`; if that path was later recreated by something other than the agent's tools, a second `remove` would delete the *different* file. This is exactly the "advisory, not a capability" hazard `FileStateMap`'s own docs warn about — I'd rationalized it away with an incomplete "sandbox blocks rebind" argument (the sandbox blocks shell rebind, not remove-then-external-recreation). **Fix:** `remove` now calls a new `FileStateMap::forget(path)` after deleting, so a recreated path reads as unknown and is refused. Test added.

**[P2] The `kind` check followed symlinks.** `@fs.kind` follows the final symlink by default, so a `Created` path replaced by a symlink to a regular file classified as `Regular` and would have been removed — contradicting the comment. **Fix:** `@fs.kind(path, follow_symlink=false)`. Test added.

The doc no longer claims the sandbox alone makes `Created` a safe delete capability — it now lists the three actual guards (forget-on-delete, no-follow kind check, sandbox source-write block) and names the remaining residual (external *in-place* replacement of an un-removed file) as deferred hardening (stored identity / soft-delete). `moon check --deny-warn` clean, 77/77 in the agent+remove set, mbti additive (`forget`). Re-review running.

bobzhang:
Re-review of the fix commit — **clean**:

> The patch correctly clears stale file provenance after successful deletion and explicitly prevents following final-component symlinks. The changed package compiles and its targeted tests pass.

Both commits now carry clean subal sign-offs. PR ready for review.

PR #475, opened 2026-07-12
test(desktop/package): inline wbtests; drop windows wbtest import block

bobzhang:
Part of the project-level simplification dropping `*_wbtest.mbt` files that need no wbtest-only imports.

The packaging/macos/windows white-box suites exercise package internals but need **no** wbtest-only imports, so they become inline tests beside the code they cover:
- `internal/packaging/packaging_wbtest` → `packaging.mbt`
- `macos/main_wbtest`, `windows/main_wbtest` → `main.mbt`

`windows/main_wbtest` never used the `for "wbtest"` `core/string` import, so the now-dead import block is dropped from `windows/moon.pkg`.

Tests moved **verbatim** (no reformatting) to stay within the desktop module's canonical formatting. No production code changes.

Verification (needs `git submodule update --init desktop/lepus desktop/editor`): `moon check` → 0 errors; packaging 4/4, macos 7/7, windows 3/3 pass.


bobzhang:
**Codex (subal) review — signoff ✅**

> The patch moves existing white-box tests verbatim into inline test blocks and removes an unused wbtest-only import. The affected desktop package tests pass in native debug and release builds without changing production behavior or public interfaces.

CI note: the root `check (nightly)` job does not run the `openseek_desktop` module, and the `desktop.yml` jobs fail on **every** open PR (including ones that do not touch desktop) due to a private-submodule-token infra issue — so CI does not exercise this change. Verified locally instead: `git submodule update --init desktop/lepus desktop/editor`, then `moon check` (0 errors) + `moon test` on all affected packages pass; tests moved verbatim.

PR #497, opened 2026-07-15
feat(protocol): typed engine event stream as a cross-backend module

bobzhang:
**Stack: #497 (this) → #498 → `protocol-desktop-decoder`.** This PR adds the module and converts the writers; the readers adopt it in the two above. Scope here is the engine side only.

## What

The engine's stdout JSONL stream is the wire contract between `openseek run`/`serve` and everything that reads it — the TUI, the desktop host, the desktop frontend, and any script consuming stdout. It lived as **~60 anonymous JSON literals** at `@xlog.info() <? {…}` call sites, with a hand-written decoder per client and nothing tying the two directions together.

They had already drifted:

| Event | Drift |
|---|---|
| `tool_result` | emitted with `brief` from 1 site, without it from 2 |
| `mcp_connect_failed` | emitted with `error` from 1 site, without it from another |
| `compaction_failed` | reported at **warn** from 1 site, **error** from 2 — same event, two severities |

And of 49 event names emitted, **~16 were decoded by nobody**; the desktop decoded **zero** `goal_*` and **zero** `mcp_*` events, silently, through a `_ => None`.

## This PR

`bobzhang/openseek_protocol` — a **leaf module with no openseek dependencies**, split so the decoder is portable:

| Package | Contents | Targets |
|---|---|---|
| `bobzhang/openseek_protocol` | `Event` (47 variants), `Usage`, `parse` | js, wasm, wasm-gc, native |
| `bobzhang/openseek_protocol/emit` | `emit` | native (`xlog`) |

One variant per event, owning its payload **and its level**. All 60 emit call sites converted across `agent/`, `agent_session/compact/`, `cmd/openseek/`, `mcp/tools/` — net **−60 lines** at the call sites.

**Why a module, not a package:** `desktop/frontend` compiles to **js** and `xlog` is native-only, so a shared decoder must exclude the writer; and `desktop` is a separate module, which only another *module* can be a `moon.work` member of. Verified: `"../protocol"` binds the working tree, not a registry snapshot.

**`Usage` is owned here**, not borrowed from `@deepseek.Usage`. The vendor's response struct *was* the wire format for that event — rename a DeepSeek field and every client's parse changes. `agent`'s `wire_usage` is the single place the two meet. `deepseek` is untouched by this PR.

## Normalized, not byte-identical

The three drifts above are normalized: `brief` and `mcp_connect_failed.error` are always written (`null` when absent), `compaction_failed` picks one level. **No client can observe the difference** — nothing reads the envelope's `level`, and every decoder reads optional fields with a lookup that rejects a non-string, so an absent key and an explicit `null` are already the same value downstream. cram is unchanged.

## The round-trip test earned its keep

MoonBit encodes `Some(v)` as the one-element array `[v]`. A `String?` payload field would have emitted `"brief":["…"]` and read as **absent** to every decoder. That is why the old `turn_loop` site hand-wrote `Json::string`/`Json::null`; `or_null` is the same rule in one place — and the test caught it before it shipped.

## Notes

- `agent` and `agent_session/compact` no longer import `@xlog` at all — they report events, they do not log.
- **Zero `.mbti` drift outside the new module.** No existing public API changes.

## Verification

| Gate | Result |
|---|---|
| `moon check --deny-warn` | clean (js, wasm, wasm-gc, native) |
| `moon cram test tests/cram` | 23/23 |
| protocol tests | js 252, wasm 172, wasm-gc 172, native — all pass |
| full native suite | 1173/1179 |

The 6 failures are the known `agent_tool/shell` timing tests: they pass 119/119 in a fresh worktree off `main`, and `agent_tool/shell` imports none of the packages touched here, so this diff cannot reach them.

## Known gaps at this commit (both closed later in the stack)

- The clients still hand-decode — #498 and `protocol-desktop-decoder` fix that, and doing so is what surfaced two dead features and a lying comment.
- `MOON_XLOG=warn openseek serve` silences the stream — fixed in `protocol-desktop-decoder`.
- Reviewed by codex: no findings against the protocol work.

### DIFF
--- a/README.mbt.md
+++ b/README.mbt.md
@@ -15,6 +15,7 @@ entry point so request encoding can be tested without network access.
 | `bobzhang/openseek/agent_session` | Typed durable conversation state and DeepSeek message projection. | `agent_session/README.mbt.md` |
 | `bobzhang/openseek/agent_session/store` | Native filesystem-backed append-only session store. | `agent_session/store/README.mbt.md` |
 | `bobzhang/openseek/agent_tool` | Tool registry, executor, output, and control-action types. | `agent_tool/README.mbt.md` |
+| `bobzhang/openseek_protocol` | Typed engine event stream (own module): the `openseek run`/`serve` stdout wire contract, decodable on every backend. | `protocol/README.mbt.md` |
 | `bobzhang/openseek/agent` | Native-only OpenSeek agent loop and local tool dispatch. | `agent/README.mbt.md` |
 | `bobzhang/openseek/cmd/openseek` | Native-only command-line entry point. | `cmd/openseek/README.md` |
 | `bobzhang/openseek/cmd/tui` | Native-only terminal UI library, the default mode of `openseek` (and `openseek tui`). | `cmd/tui/README.md` |
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -311,7 +311,7 @@ pub async fn[X] run_turn_in_scope(
     }
   } catch {
     error => {
-      @xlog.error() <? { "event": "agent_setup_failed", "error": "\{error}" }
+      @emit.emit(AgentSetupFailed(error="\{error}"))
       return session
         |> append_item(Terminal(Failed("agent setup failed: \{error}")))
     }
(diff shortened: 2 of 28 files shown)


bobzhang:

PR #498, opened 2026-07-15
feat(tui): decode the engine stream through the typed protocol

bobzhang:
**Stack: #497 → #498 (this) → `protocol-desktop-decoder`.** Review #497 first; this diff is only the TUI adoption plus the `parse` correction it forced.

## What

#497 built the typed protocol, but no client used it — so its headline property, *adding a variant is a compile error in every client*, was not yet real. This is the first client to buy it.

The TUI hand-decoded the engine's JSONL by `"event"` string and ended in `_ => None`, so any event it did not handle vanished silently. It now matches **exhaustively** over `@protocol.Event`: every event is either mapped to an `AgentEvent` or named in the ignore arm, with the reason written down.

`AgentEvent` stays a view model — several engine events fold into one UI concept (`agent_setup_failed`, `turn_failed`, `compaction_failed` all read as `AgentError`) and most carry less than the wire does (`usage` keeps only the derived total). The indirection is the point: the match is what the compiler checks.

## It immediately found a dead feature

`a85d5682` ("Suppress streaming reasoning delta logs", **2026-06-21**) removed `on_reasoning_delta` from the engine's `StreamHandler`. `reasoning_delta` had not been emitted for three weeks. The TUI still had:

- `ReasoningDelta` in `AgentEvent` + its decode branch
- `state.streaming_reasoning`
- a `status_view` branch rendering a dimmed **"Thinking …"** activity line
- **two tests that passed** — because they fabricate the JSON the engine no longer sends

and `README.mbt.md` still advertised *"streamed thinking … appear live on the activity line."*

Users saw `Working (Ns)` instead of live thinking since June 21. Nothing caught it, because the wire had no schema and a client's tests were free to invent one.

The dead path is removed and the README now says what the TUI does. **Restoring live thinking is a separate question** — it needs the engine to emit deltas again, which `a85d5682` deliberately stopped for log noise. That trade only exists because *the log IS the protocol*.

## It also caught `parse` being too strict — before any client depended on it

Both found by existing TUI tests:

| Field | Why it cannot be required |
|---|---|
| `steer_applied.kind` | Engines older than 2026-06-25 (`9b2f6c43`) emit only `content`. The TUI launches whichever `openseek` is on **PATH**, so new-reader/old-engine is a real pairing — a strict parse would silently drop those steers. Defaults to `"prompt"`, which is what every client did by hand. |

The rule was refined once more in `protocol-desktop-decoder` after two desktop fixtures failed: **a field is defaulted iff it was added after its event already existed.** "No reader uses it" is *not* a reason — that reasoning had also defaulted `names`, which is reverted there. Only `kind` qualifies.

Leniency never weakens the round-trip: `emit` always writes every field, so it only affects lines `emit` never produced.

## Verification

Clean worktree, full suite green:

| Gate | Result |
|---|---|
| native | **1168/1168** |
| js | 254/254 |
| wasm-gc | 7/7 |
| `moon cram test tests/cram` | 23/23 |

(The 6 `agent_tool/shell` failures seen in the main checkout do not occur here — checkout-specific, as #497 notes.)

`.mbti` drift: one line, the removed dead variant.

Reviewed by codex: no findings — *"preserves the TUI's supported event behavior, handles the documented legacy fields, and exhaustively classifies protocol events."*

### DIFF
--- a/README.mbt.md
+++ b/README.mbt.md
@@ -131,9 +131,9 @@ tool. This keeps the surface small; revisit if a concrete need appears.
 The terminal UI is the **default** mode of the single `openseek` binary (the
 `cmd/tui` library, also reachable as `openseek tui`): a scrolling transcript with
 a live composer. It spawns the `openseek` engine (by default this same binary in
-`serve` mode) and renders its JSONL event stream — streamed thinking and answer
-text appear live on the activity line, and each turn's reasoning is kept as a dim
-`✻` transcript aside above its answer.
+`serve` mode) and renders its JSONL event stream — streamed answer text appears
+live on the activity line, and each turn's reasoning is kept as a dim `✻`
+transcript aside above its answer.
 
 ```bash
 export DEEPSEEK=sk-...
--- a/cmd/tui/internal/event/decode.mbt
+++ b/cmd/tui/internal/event/decode.mbt
@@ -1,128 +1,136 @@
-// Mapping from the engine's decoded JSONL values to typed `AgentEvent`s. Reading
-// lines, skipping blanks, and parsing each as JSON is `bobzhang/jsonl`'s job
-// (`@jsonl.each` in `cmd/tui`); this module only maps an already-parsed value,
-// keyed by its `"event"` discriminator.
+// Mapping from the engine's typed protocol events to the TUI's `AgentEvent`.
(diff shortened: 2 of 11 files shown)


bobzhang:

PR #499, opened 2026-07-15
feat(desktop): decode the engine stream through the typed protocol

bobzhang:
**Stack: #497 → #498 → this.** Review those first; this diff is the desktop adoption and the fixes it forced. With this, **every client of the engine stream reads it through one typed decoder** — no `_ => None`, no hand-keyed `"event"` strings anywhere.

## Commits

| Commit | What |
|---|---|
| `aa4c10ad` | drop the unused `Event::name` — zero callers since written, cost an arm per variant |
| `6947a60c` | desktop **host** decodes via `@protocol.parse`; wires `"../protocol"` into `desktop/moon.work` |
| `18e640b7` | **one encoder** (`Event::to_json`); the host's synthetic event uses it |
| `6a06318c` | desktop **frontend** (js) decodes via `@protocol.parse` |
| `d603b01c` | a logging env var must not silence the protocol |
| `3c0b7424` | hand-update the js-only transcript interface |

## What adoption found

**A second dead subsystem.** `4b1ec831` ("remove unused moon_check runtime events", **2026-07-05**) deleted `runtime_update_text`; `runtime_update` had not been emitted for ten days. The host still decoded it into a `RuntimeUpdate` variant, under a test that passed by fabricating the event. (The *frontend's* `RuntimeUpdate` is **not** dead — it is also fed by `background_notice`, so its view model and 67-line renderer stay. Only the legacy wire name is gone.)

**The frontend's half of the dead live-thinking feature**, matching #498's TUI half: `ReasoningDelta` (unemitted since `a85d5682`, 2026-06-21), the `streaming_reasoning` buffer it was the only writer of, the bubble `view.mbt` rendered from it, and `reasoning_view`, which existed only for that bubble. Committed reasoning renders through `activity_view` and is unaffected.

**A comment that lied.** The host synthesized `compaction_failed` with `"reason"` while the engine emits `"error"`, under a doc promising *"the same wire shape the engine emits"*. It worked only because the frontend's decoder happened to try both spellings. Rather than fix the word, the wire shape now has **one author** — `Event::to_json` — and the host goes through it, so the promise is enforced by construction. `emit` is now level + `to_json` + write instead of 47 literal arms. **Proof the refactor is faithful: the byte-level snapshots written against the old literals pass unchanged, field order included.**

**Three more invented comments**, each checked against git rather than believed:

| Claim in code | Reality |
|---|---|
| *"a build that does not number its steps"* | `agent_step` and `step` arrived in the same commit (`1f761102`) |
| *"older engines wrote `error` where current ones write `reason`"* | `agent_aborted` has carried `reason` since `1f761102`, never `error` |
| `names` defaulted because "no reader uses it" | `names` shipped with `mcp_tools_registered` (`13dc6e03`) — reverted |

## The leniency rule, settled

Two desktop fixtures failed because `parse` required `tool_call_id` and `usage`'s cache counters. The tempting fix — default anything no reader uses — is the same wrong reasoning that had defaulted `names`. `git log -S` settles it:

> **A field is defaulted iff it was added after its event already existed**, because engines older than the field still emit the event without it — and the clients launch whichever `openseek` is on `PATH`.

Three qualify — `steer_applied.kind`, `session_started.workspace_root`, `tool_result.brief` — each named in `parse`'s contract with the commits that prove it. `tool_call_id`, the cache counters and `names` all shipped with their events, so their absence means the line is not what it claims.

A later holistic review caught that `workspace_root` was required despite qualifying: stating the rule is not auditing against it. That fix is in this branch, with the wire pinned byte-for-byte so making a field optional cannot move it for current engines.

## A logging env var could silence the protocol

`MOON_XLOG=warn openseek run` emitted **zero lines** — reproduced against the real binary (3 by default, 0 under `warn`, 3 again after). Every event is info/warn/error and `@xlog`'s root level comes from `MOON_XLOG`, so a TUI attached to such an engine rendered nothing, with no error to explain it. The engine pins its own level; a cram case pins `warn` and `error`.

**This is a guard, not a cure.** The stream still has no identity apart from the log — the same coupling that let `reasoning_delta` be dropped for log noise and silently take a UI feature with it in two clients. A cure gives the protocol its own sink. Out of scope; the README says so.

## Interfaces

`3c0b7424` is by hand and deliberately so. A `pkg.generated.mbti` is a per-backend artifact, so a module supporting several nominates one (`preferred_target: "native"`), and `frontend/transcript` — `supported_targets = "js"` — falls outside it: neither `moon -C desktop info` nor `--target native` touches it, and `--target js` refuses to write. `viz` sits the same way in the root module. The edit is verified byte-identical to moon's own js check output. **The real fix is giving a js-only package its own module with `preferred_target: "js"`** — a layout change well outside this stack, noted rather than done.

## Verification

| Gate | Result |
|---|---|
| root | 1168 native / 27 js / 6 wasm-gc |
| desktop | 223 native / 210 js |
| cram | **24/24** |
| `moon check --deny-warn` | clean, both workspaces, both targets |
| `moon fmt --check` | clean (CI's scoped `. frontend internal package`) |
| `.mbti` drift | none |

## Review

Codex found six real issues across eight passes, all fixed here: three stale generated interfaces (`moon info` at root reaches neither a separate module nor a separate workspace, and skips js-only packages entirely); `session_started.workspace_root` required despite qualifying for the leniency rule; `usage` counters truncating `1.5` to `1`; and `Usage` still deriving `FromJson` after that fix, which left a *second public decoder* disagreeing with `parse`.

One fmt finding at `engine_event.mbt:82` was rejected on evidence — CI's exact command exits 0 — and codex withdrew it on re-review.

Final verdict: *"No actionable correctness regressions were found."* See #500 for the whole change reviewed as one thing; that is where the last three were found.


bobzhang:

PR #502, opened 2026-07-15
refactor(protocol): delete the third copy of the rule it kept getting wrong

bobzhang:
Sweep of the `protocol` module, which had never had one. **Its five source files came back clean** — the finder explicitly declined to collapse the repeated compaction literals in `to_json`, on the grounds that the literal table *is* the contract and a helper branching on an absent `summary` is how you'd get a byte change nobody reviews. Agreed.

Everything below is documentation that drifted out from under the code **within a month**, in the package built to stop exactly that.

## The README kept a third copy of the absence rule

It sat beside the two in `parse.mbt` and `command.mbt`. Commit `3906fcea` corrected those two and never touched it, so it still claimed:

> A field is defaulted **iff it was added after its event already existed** … Exactly one qualifies today

while `parse.mbt` nine files away says **three** qualify, and `command.mbt` documents a **second case** that the README's "iff" explicitly excludes.

**It is deleted, not corrected.** A rule copied is a rule that drifts, and this one proved it in under a month. The README now names the rule and points at the doc comment beside the code that enforces it.

## The README documented half the module

`Command`, `SteerKind`, and the whole command direction (added in `f11f384e`) were absent from the opener and the package table. A reader looking for the command wire contract would conclude there isn't one **and hand-roll a second encoder** — precisely what `f11f384e` deleted from the desktop host.

## The invariant's stated proof was a dangling pointer

`parse`'s doc comment cited `protocol_test.mbt` as where the round-trip is pinned. **That file has never existed in this repo** (`git log --all` confirms). The test is `emit/emit_test.mbt` — it needs both halves, so it lives with the writer.

## A comment claimed a check that isn't running

It said MoonBit's exhaustiveness covers `emit`/`parse`/`name`. There is no `name` (`aa4c10ad` removed it), and `emit` matches nothing — it delegates to `to_json`. Worse, it reassured about the check where it matters most: **`level`'s `_ => Info` means a new variant that should warn ships at `Info` with no error.**

One variant still cannot hold two levels — what the package set out to stop. But *being made to choose one* is a weaker promise than that comment implied, and it now says so. I did not expand the catch-all into 40 arms; that's a design call, and longer.

## The one code change

`parse_steer` restated `"text": String(..)` in three arms, and the precedence between "no text" and "bad kind" was **emergent from arm order**. A guard hoists it once and flattens the nested match.

All seven error strings are wire contract — they reach a controller as `command_error` — and `command_test.mbt` pins each verbatim.

## Verification

| Gate | Result |
|---|---|
| protocol | 259 js / 179 wasm / 179 wasm-gc |
| root | 1186 native |
| `.mbti` drift | none |

**Codex: clean** — *"The executable change preserves the existing `parse_steer` behavior while simplifying validation."*

### DIFF
--- a/protocol/README.mbt.md
+++ b/protocol/README.mbt.md
@@ -1,15 +1,16 @@
 # OpenSeek Protocol
 
-`bobzhang/openseek_protocol` owns the engine's stdout event stream — the wire
-contract between `openseek run` / `openseek serve` and everything that reads
-them: the TUI, the desktop host, and any script consuming `run`'s stdout.
+`bobzhang/openseek_protocol` owns **both directions** of the serve protocol: the
+stdout event stream the engine reports (`Event`), and the stdin command stream it
+is told (`Command`). Between them they are the whole wire contract with the TUI,
+the desktop host, the desktop frontend, and any script driving `run` or `serve`.
 
 It is a **leaf module with no openseek dependencies**, split so the decoder is
 portable:
 
 | Package | Contents | Targets | Deps |
 | --- | --- | --- | --- |
-| `bobzhang/openseek_protocol` | `Event`, `Usage`, `to_json`, `parse` | js, wasm, wasm-gc, native | `core/json` |
+| `bobzhang/openseek_protocol` | `Event`, `Usage`, `Command`, `SteerKind`, `to_json`, `parse` | js, wasm, wasm-gc, native | `core/json` |
 | `bobzhang/openseek_protocol/emit` | `emit` (level + `to_json` + log) | native | `xlog`, above |
 
 Only the *writer* needs `@xlog`, which is native-only. Keeping it in its own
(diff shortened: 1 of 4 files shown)


bobzhang:

Each commit reviewed in isolation (tree at *N*, `--base` at *N−1*). **This is why:** the isolated pass on `4a8d07f5` found a bug that two full-diff reviews had passed over.

**`4a8d07f5` — delete the third copy of the rule it kept getting wrong**
> [P2] Parse the JSONL string before passing it to `Command::parse` — `line` is a `String` because `to_jsonl()` serializes the command and appends a newline, while `Command::parse` accepts a `Json`. The new command-direction example therefore cannot type-check.

Correct, and the irony is total: a README example that **could not have compiled**, inside an `mbt nocheck` fence, in the commit whose subject is that the docs stopped matching the code. Fixed in `b3999195`. The event example above it was already right — `to_json()` returns the `Json` that `parse` takes; only the command half, written later, was wrong.

**`585c3c32` — drop the redundant String() in parse_steer's patterns**
> The simplified JSON string patterns preserve existing behavior. Protocol tests pass across wasm-gc, wasm, JavaScript, and native targets.

**`b3999195` — the command example could not have compiled**
Docs-only; the fix for the finding above.

Combined-diff reviews (both clean, both having missed the above): *"The parse_steer rewrite preserves existing behavior, and the remaining changes are documentation-only."*

PR #503, opened 2026-07-15
refactor(mcp/streamhttp): one error shape, one decision, no unused self

bobzhang:
Sweep of the MCP transports. **`mcp/stdio` came back clean** — its four cancellation blocks look duplicated but each carries different progressive cleanup and returns out of the enclosing function, which is exactly the code to leave alone. So this is `streamhttp` only, in two commits: tests, then source.

## Tests (−31)

- **A dead 13-line reply.** The mock endpoint built a `tools/list` Json and then `ignore(reply)`d it — the bytes it appears to produce are hand-written below, deliberately split at a token boundary, which is the *point* of that test. It was 14 lines pretending to be the fixture.
- **Four connect-or-fail prologs** byte-identical modulo the url, and **three `initialize` arms** byte-identical outright → `connect_or_fail` / `answer_initialize`. The unreachable-endpoint test keeps its own call: different timeout, and it wants the `None`.

## Source (all private, no `.mbti` drift)

- **`deliver_error`** — the synthesized JSON-RPC error was spelled out four times, once per way an exchange can fail a pending request. Each site now names only its message, the one thing that ever differed.
- **`run_post`'s two tail blocks were one decision.** The first fired on deadline expiry and set `answered = true` for no reason but to make the second skip. They're mutually exclusive on `bounded`, nothing suspends between them, and the flag was pure sequencing. Case check: `(None, !answered)` → timeout (before: via the flag); `(Some, !answered)` → ended-without-answering (before: via the first guard failing); `answered` → neither, both ways.
- **`read_sse_stream`** was a method opening with `ignore(self)` and never touching it — the `ignore` existed only to silence a parameter it didn't need. Plain function now.
- **`reserved_header`** was a 12-line match whose only job was `true`/`false`. Now the `is` expression it always was — the idiom already used in `cmd/viz_server` and `desktop/internal/shellenv`.

## Declined

`delete_session`'s three identical cancellation catches collapse into one `try`/`catch`, and the finder's reasoning was sound. It's the only finding touching cancellation flow, and a sweep is the wrong place to spend that risk. Left for someone with a reason to be in that function.

## Verification — and a caveat about it

| Gate | Result |
|---|---|
| `.mbti` drift | none |
| `mcp/streamhttp` | 5/5, **all actually executed** |

**Read the codex verdict on the tests with care.** Its sandbox can't bind TCP, so four of the five streamhttp tests were skipped in its run — its clean verdict validates the source refactor, not the test one. Locally all five bind and run (`-v` confirms no `skipping` lines), which is where the test-helper change is actually exercised.

Codex: *"The production changes are behavior-preserving refactors, and the test helper extraction preserves the previous setup."*

### DIFF
--- a/mcp/streamhttp/streamhttp_test.mbt
+++ b/mcp/streamhttp/streamhttp_test.mbt
@@ -87,19 +87,6 @@ async fn mock_mcp_endpoint(
                 _ => ()
               }
             }
-            let reply : Json = {
-              "jsonrpc": "2.0",
-              "id": id,
-              "result": {
-                "tools": [
-                  {
-                    "name": "fetch",
-                    "description": "Fetch a URL",
-                    "inputSchema": { "type": "object" },
-                  },
-                ],
-              },
-            }
             conn.send_response(200, "OK", extra_headers={
               "Content-Type": "text/event-stream",
             })
@@ -111,8 +98,9 @@ async fn mock_mcp_endpoint(
             )
             conn.flush()
             // Split at a token boundary: SSE joins the lines with a newline,
-            // which JSON tolerates only between tokens.
-            ignore(reply)
+            // which JSON tolerates only between tokens. The reply is spelled
+            // out here rather than built and stringified, because the split is
+            // the point of the test.
             conn.write(
               "data: {\"jsonrpc\":\"2.0\",\"id\":\{id.stringify()},\r\ndata: \"result\":{\"tools\":[{\"name\":\"fetch\",\"description\":\"Fetch a URL\",\"inputSchema\":{\"type\":\"object\"}}]}}\r\n\r\n",
             )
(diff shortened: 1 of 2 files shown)


bobzhang:

**`3c582fb3` — delete a dead reply, collapse four copied prologs**
> The refactoring preserves the repeated connection and initialization behavior, and the removed JSON value was unused. The project checks successfully and no actionable regression was found.

**`13cfd2a3` — one error shape, one decision, no unused self**
> The refactoring preserves existing behavior, compiles successfully, and introduces no identifiable functional regressions.

This is the commit that merges `run_post`'s two tail blocks and drops the `answered` write whose only job was making the second block skip — the one worth the closest read.

**`15bba493` — drop the redundant String() in Json literal patterns**
> The string literal patterns are equivalent to the removed explicit Json `String` constructor patterns. The project type-checks across all targets, and the streamhttp tests pass.

⚠️ **Caveat on codex's test evidence here:** its sandbox cannot bind TCP, so four of the five streamhttp tests *skip* in its runs (`local TCP bind unavailable; skipping …`). Its verdicts validate the source refactor; the test refactor is exercised locally, where all five bind and run.

PR #504, opened 2026-07-15
refactor(mcp): one optional-string read, one string-map decode, one assert

bobzhang:
Sweep of the MCP core and its config decoder. Both are strict where they should be, and **every change preserves that exactly** — no `guard … else raise Protocol` is touched, and every error message is byte-identical.

## Source

- **`optional_string`** — `types.mbt` read "the string at this key, or nothing" **six times**, in two spellings: four wanting an `Option`, two defaulting to `""`. Stated once; the defaulting sites now say `.unwrap_or("")`, which is the difference they actually have. It reads only the fields MCP lets a server omit — the required ones still raise.
- **`decode_string_map`** — `env` and `headers` were the same twelve lines twice, differing only in the field name, *which already appeared verbatim inside both of their error messages*. Only two sites, under the usual 3+ bar, but the parameterization is exact and costs nothing.
- **Two push loops** are what they meant: `entries.map(decode_tool)` (`map` is raise-polymorphic, so `Protocol` propagates identically) and `tools.append(page)`.
- **`is_supported_version`** respelled `"2025-11-25"` nine lines under the `ProtocolVersion` const that defines it — the one line a version bump would silently leave behind. It names the const now.

## Tests

`config_wbtest.mbt` copied the same try/catch/assert chain into **six** tests, differing only in the fixture. `assert_invalid` takes it; each test is now the configuration it's about. The `_ => "other"` arm — the thing that keeps an unexpected error from reading as a pass — exists once.

## For the next sweep

`Json::value` is **deprecated** (suggestion: `if json is Object(obj) { obj.get(key) }`). The finder verified it exists but not that it's live; `--deny-warn` caught it, so both new helpers pattern-match instead.

## Verification

| Gate | Result |
|---|---|
| `mcp` | 23/23 |
| `mcp/config` | 13/13 |
| `.mbti` drift | none |

**Codex: clean** — *"The refactors preserve existing decoding, pagination, validation, and protocol-version behavior."*

### DIFF
--- a/mcp/client.mbt
+++ b/mcp/client.mbt
@@ -14,7 +14,9 @@ pub const ProtocolVersion : String = "2025-11-25"
 /// The tool surface (`tools/list`, `tools/call`) is stable across these
 /// revisions.
 fn is_supported_version(version : String) -> Bool {
-  version == "2025-11-25" ||
+  // The preferred version is named, not respelled: this list is the one place
+  // a bump to `ProtocolVersion` would otherwise silently leave behind.
+  version == ProtocolVersion ||
   version == "2025-06-18" ||
   version == "2025-03-26" ||
   version == "2024-11-05"
@@ -79,9 +81,7 @@ pub async fn McpClient::list_tools(self : McpClient) -> Array[McpTool] {
     }
     let result = self.rpc.request(method_="tools/list", params~)
     let (page, next) = decode_tools_page(result)
-    for tool in page {
-      tools.push(tool)
-    }
+    tools.append(page)
     match next {
       Some(next) => cursor = Some(next)
       None => break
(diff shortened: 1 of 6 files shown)


bobzhang:

Each commit reviewed in isolation (tree at *N*, `--base` at *N−1*), not just the combined diff.

**`20372dd6` — one optional-string read, one string-map decode, one assert**
> The refactors preserve the existing decoding, pagination, and validation behavior without changing the public interface. Available checks and tests pass.

**`63a2b04b` — drop the redundant String() in Json literal patterns**
> The revised JSON literal patterns are semantically equivalent to the explicit `String(...)` patterns. All 23 targeted MCP tests pass.

Combined diff also reviewed: *"The refactors preserve existing decoding, pagination, validation, and protocol-version behavior."*

PR #505, opened 2026-07-15
refactor(mcp/tools): flatten with append, name the provider alphabet

bobzhang:
Sweep of the MCP tool registry. Two findings; the rest of the package — the name capping, uniquifying, and the schema/description budgets — is guard code and stays exactly as written.

- **`build`** delegated to `build_grouped` and then copied each group's tools out one at a time. `append` says what it meant: flatten the groups.
- **`sanitize_segment`** open-coded `[A-Za-z0-9_-]` as three range comparisons. `is_ascii_alphabetic`/`is_ascii_digit` are exactly those ranges, so the accepted set is **preserved verbatim rather than re-derived** — which matters, because this is the guard keeping a server's tool name inside the provider's function-name rules. It's the idiom `tui/internal/composer/text_cursor.mbt` already uses; core has no `is_ascii_alphanumeric`.

## Considered and left

- `CallTimeoutMs`/`DefaultDiscoveryTimeoutMs` are `pub` with no consumer outside the package — but so are the sibling tunables in `mcp/stdio`, `mcp/streamhttp` and `agent_tool/shell`. That's house convention, not dead code.
- A helper for the three stdio `build` tests would need four parameters to serve three sites: the parameterization costs what the duplication does.

Tests: **7/7**, including the sanitizer's accepted-set assertions.

### DIFF
--- a/mcp/tools/tools.mbt
+++ b/mcp/tools/tools.mbt
@@ -56,11 +56,7 @@ pub fn tool_name(server : String, tool : String) -> String {
 fn sanitize_segment(segment : String) -> String {
   let out = StringBuilder()
   for ch in segment {
-    if (ch >= 'a' && ch <= 'z') ||
-      (ch >= 'A' && ch <= 'Z') ||
-      (ch >= '0' && ch <= '9') ||
-      ch == '_' ||
-      ch == '-' {
+    if ch.is_ascii_alphabetic() || ch.is_ascii_digit() || ch == '_' || ch == '-' {
       out.write_char(ch)
     } else {
       out.write_char('-')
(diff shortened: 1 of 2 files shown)


bobzhang:

**`9167fc56` — flatten with append, name the provider alphabet**
> The refactoring preserves the sanitizer's accepted ASCII character set and flattening order. MoonBit checks and all seven targeted MCP tool tests pass.

**`b9a36b7b` — drop the redundant String() in Json literal patterns**
> The changes replace explicit JSON string-constructor patterns with equivalent overloaded string-literal patterns. The project checks successfully, and all seven targeted mcp/tools tests pass.

The sanitizer's accepted set being *preserved rather than re-derived* is the point of the first one — it's the guard keeping a server's tool name inside the provider's function-name rules.

PR #506, opened 2026-07-15
test(jsonrpc): one failing-request helper, not six copies of the plumbing

bobzhang:
Sweep of the `jsonrpc` package. **Its library code came back clean** — `protocol.mbt` and `client.mbt` have no duplication, no dead public API (every `.mbti` item is reached; `set_notification_handler` and `is_closed` from `desktop/internal/lsp`), and no hand-rolled loop core would replace. This is tests only.

## The win

Six tests each carried the same eight lines: issue a request, stringify the outcome, catch `Failed`/`Closed`, re-raise cancellation, fall through to `"other error"`. Only the method name differed.

**The point isn't the ~30 lines.** It's that the cancellation guard — the arm you forget when adding the seventh of these, and whose absence turns a cancelled test into a *hang* rather than a failure — now exists once.

Two sentinels unify: two sites said `"returned"` where four said `"no error raised"`. Both are only reachable when the test is already failing, and every site asserts a different string, so pass/fail is identical at each.

Also: a 4-line `match { Unknown => (); _ => fail(...) }` became the `assert_true(… is Unknown)` the same file already uses two tests earlier. Two more like it have multi-line arguments and are left — 4 lines each, not worth a fragile rewrite.

## Verification

| Gate | Result |
|---|---|
| `jsonrpc` | 29/29 |
| `.mbti` drift | none |

### DIFF
--- a/jsonrpc/client_wbtest.mbt
+++ b/jsonrpc/client_wbtest.mbt
@@ -137,8 +137,8 @@ async test "concurrent requests each get their own reply even when answered out 
     let second = group.spawn(() => {
       client.request(method_="second", params=Json::empty_object())
     })
-    assert_true(first.wait() is { "method": String("first"), .. })
-    assert_true(second.wait() is { "method": String("second"), .. })
+    assert_true(first.wait() is { "method": "first", .. })
+    assert_true(second.wait() is { "method": "second", .. })
   })
 }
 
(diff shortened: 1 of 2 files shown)


bobzhang:

**`c49e7827` — one failing-request helper, not six copies of the plumbing**
> The helper refactor preserves the prior test outcomes and cancellation handling, and the protocol assertion is equivalent. All 29 targeted jsonrpc tests pass.

*"preserves … cancellation handling"* is the one that mattered: the guard whose absence turns a cancelled test into a hang now exists once.

**`17fc1fd0` — drop the redundant String() in Json literal patterns**
> The changes use equivalent JSON string literal patterns without altering test behavior. The jsonrpc package compiles and all 29 targeted tests pass.

PR #511, opened 2026-07-16
fix: migrate to the 2026-07 toolchain (with-pattern parens + ToDoc extend)

bobzhang:
Makes the repo build and check clean on the current toolchain (`moon 0.1.20260713` / `moonc 2026-07-15`, which is what stable `latest` now serves). Two independent toolchain-migration fixes, one commit each; #512's `extend` commit was folded in here so this PR fixes CI whole.

## 1. Parenthesize with-patterns (parser change)

`p | q with x = v` no longer parses — the `with` clause must sit inside parens, `p | (q with x = v)`. Every existing `with`-pattern used the old spelling, so `main` didn't compile (the nightly `unexpected token with` failure). All 28 sites across 9 files moved to the parenthesized form; three shapes, the subtle one being `Some(String(c)) | (Some(Null) with c = "")` (the clause binds the branch, not the payload). Pure spelling change; net −12 lines after fmt collapses.

## 2. `extend Input with ToDoc` (trait-promotion deprecation)

The one remaining `--deny-warn` error: `input.to_doc()` relied on `impl @doc.ToDoc for Input` being *implicitly* callable as a method, now deprecated. One `extend Input with @doc.ToDoc::{to_doc}` next to the impl makes it explicit. Call site untouched, no `.mbti` drift. (The message's suggested `{to_doc, ..}` doesn't parse; `{to_doc}` does.)

## On the "stable formatter will break" concern (codex P1, refuted)

CI runs `moon fmt --check` on both stable and nightly. Codex flagged that stable might strip these parens. **Investigated:** `version.json` shows stable `latest` = `moon 0.1.20260713` + `moonc 2026-07-15` — the new parser, identical to the build toolchain, where fmt *keeps* the parens. Stable was bumped to the 07-15 moonc; both jobs now want the parenthesized form. Codex's re-review confirmed and withdrew it.

## Verification (root + desktop, all targets)

`moon check --deny-warn` clean · `moon fmt --check` clean · desktop native+js clean · tests 1175 native / 27 js / 12 wasm-gc. Both commits codex-signed-off (see #511/#512 comments).

### DIFF
--- a/agent_tool/internal/auto_check/auto_check.mbt
+++ b/agent_tool/internal/auto_check/auto_check.mbt
@@ -112,11 +112,11 @@ pub fn tally_diagnostics(text : String, truncated~ : Bool) -> CheckErrors {
       {
         "level": String("error"),
         "path"? : Some(String(path))
-        | _ with path = "",
+        | (_ with path = ""),
         "loc"? : Some(String(loc))
-        | _ with loc = "",
+        | (_ with loc = ""),
         "message"? : Some(String(message))
-        | _ with message = "",
+        | (_ with message = ""),
         ..
       } => {
         error_count = error_count + 1
--- a/agent_tool/internal/auto_check/parse_gate.mbt
+++ b/agent_tool/internal/auto_check/parse_gate.mbt
@@ -209,11 +209,11 @@ fn decide_parse_gate(
         "level": String("error"),
         "error_code": Number(code, ..),
         "loc"? : Some(String(loc))
-        | _ with loc = "",
+        | (_ with loc = ""),
         "message"? : Some(String(message))
-        | _ with message = "",
+        | (_ with message = ""),
         "context"? : Some(String(context))
-        | _ with context = "",
+        | (_ with context = ""),
         ..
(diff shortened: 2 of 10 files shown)


bobzhang:
## Codex sign-off

Two passes.

**First pass** raised one P1 — that CI's `stable/latest` `moon fmt --check` would strip these parens and fail, trading a nightly parse failure for a stable format failure. I investigated rather than complied: `cli.moonbitlang.com/version.json` shows `latest` now resolves to `moon 0.1.20260713` + `moonc v0.10.4+ (2026-07-15)` — the new-parser toolchain, byte-identical to the one this branch was formatted on, where `fmt --check` keeps the parens. The finding was reasoning from pre-07-15 stable, which no longer matches what `latest` serves.

**Second pass** (against the pushed single-commit branch) confirmed it, running formatting on both current and nightly toolchains:

> The parenthesization preserves the existing default-pattern semantics while restoring compatibility with the new parser. Cross-target checks, formatting, interface generation, and targeted tests passed on current and nightly toolchains. No regression attributable to the patch was found.

So the stable-fmt concern is resolved by the toolchain itself having moved, not by any change here.

PR #514, opened 2026-07-16
refactor(desktop): last-element patterns instead of get(length-1)

bobzhang:
Companion to #513, desktop workspace. Four spots that reach for the last element via `get(length - 1)` or index the tail of a `split` — each reads more directly as a pattern. Function-body-internal, **no `.mbti` drift**, behavior-identical; desktop native + js suites green.

| file / fn | before | after |
|---|---|---|
| `transcript/sessions.mbt` `ends_with_assistant` | `match items.get(length-1) { Some(item) => …; None => false }` | `items is [.., item] && …` |
| `model.mbt` `last_assistant_content` | `messages.get(length-1) is Some(item)` | `messages is [.., item]` |
| `fileeditor/files.mbt` `basename` | `segments => segments[length-1].to_owned()` | `[.., last] => last.to_owned()` |
| `internal/extension/fs_watch.mbt` `ignored_watch_path` | `[] => false; segments => segments[length-1] is (".git" \| …)` | `[.., ".git" \| …] => true; _ => false` |

## Behavior-identical

- `[.., item]` binds the last element and fails on empty — exactly what `get(length-1)` returning `None` did. `sessions`/`model` are Bool/Option-returning, so the whole `None => false` / `else None` arm collapses out.
- `basename`: the `[]` arm already handled empty, so `[.., last]` is precisely the non-empty case it replaces.
- `fs_watch`: empty and non-matching-last both fall to `_ => false`; a matching last yields `true` — the same truth table.

All last-element, on `String`/struct elements — no Char/code-unit concern (that hazard only applies to string-*character* patterns, which none of these are).

Verification: desktop `--deny-warn` clean (native + js) · `fmt --check` clean (CI's scoped form) · tests 230 native / 216 js · no `.mbti` drift.

### DIFF
--- a/desktop/frontend/fileeditor/files.mbt
+++ b/desktop/frontend/fileeditor/files.mbt
@@ -147,7 +147,7 @@ pub fn read_file(
 pub fn basename(path : @pathx.Relative) -> String {
   match path.0.split("/").collect() {
     [] => path.0
-    segments => segments[segments.length() - 1].to_owned()
+    [.., last] => last.to_owned()
   }
 }
 
--- a/desktop/frontend/model.mbt
+++ b/desktop/frontend/model.mbt
@@ -1083,8 +1083,7 @@ fn Conversation::sweep_pending_steers(self : Conversation) -> Conversation {
 
 ///|
 fn Conversation::last_assistant_content(self : Conversation) -> String? {
-  if self.messages.get(self.messages.length() - 1) is Some(item) &&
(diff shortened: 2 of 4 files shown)


bobzhang:
## Codex sign-off

> The last-element pattern refactors preserve the prior empty-collection and final-element behavior. MoonBit checks and desktop tests, including the JavaScript target, pass.

PR #524, opened 2026-07-18
feat(agent_tool): catch tool execution failures at the dispatch boundary

bobzhang:
## Summary

Add a single error-handling boundary in `execute_tool_call` so tools don't each have to wrap their own IO, then use it to drop `remove`'s now-redundant local `@fs.*` catches.

### The boundary (`agent_tool/agent_tool.mbt`)

`execute_tool_call` now wraps `tool.execute.run(...)` in a `catch`:

```moonbit
tool.execute.run(call.arguments) catch {
  error =>
    if @async.is_being_cancelled() || @async.is_cancellation_error(error) {
      raise error                                                  // cancellation must abort the turn
    } else {
      ToolAction::respond("error: \{call.name} failed: \{error}", is_error=true)
    }
}
```

A tool's execution failure becomes a **recoverable** error `ToolAction` the model sees and continues from, instead of propagating to the terminal handler in `AgentLoop` (which ends the turn as `Failed`).

**Why not `noraise`?** Since `async` defaults to raising in MoonBit, marking `execute_tool_call` `noraise` would force *cancellation* to be caught and swallowed here too — the turn would no longer be cancellable mid-tool-call. So cancellation is explicitly re-raised and the function stays raise-able. This rationale is captured in a doc comment on the function.

### `remove` cleanup (`agent_tool/remove/remove.mbt`)

Drops the four local `@fs.*` catches (`exists`/`kind`/`read_file`/`remove`) — the boundary now converts those IO faults uniformly. The expected-case **gate** messages (no such file, not a regular file, not created this session, content changed) stay local, since those are `guard` branches, not `catch`es.

Behavior change worth calling out: a genuine IO fault during `remove` now yields `"error: remove failed: <e>"` (recoverable) instead of the previous per-call phrasing. The **disposition** is the win — recoverable tool error vs. the terminal turn failure that removing the catches *without* this boundary would have caused.

## Test plan

- New dispatch test: a raising tool yields an `is_error` result rather than propagating.
- `moon fmt --check`, `moon check agent_tool agent --deny-warn` — clean
- `moon test -p agent_tool` — **486 pass**; `moon test -p agent_tool/remove` — **16 pass**
- No `.mbti` drift

## Follow-up (not in this PR)

The other file tools (`read`, `write`, `edit`, `multi_edit`) still carry ~40 local IO catches that this boundary makes redundant. Stripping them is a larger, per-site change (some have messages worth keeping) — best done separately.

### DIFF
--- a/agent_tool/agent_tool.mbt
+++ b/agent_tool/agent_tool.mbt
@@ -237,12 +237,35 @@ pub fn Tools::function_tools(self : Tools) -> Array[@deepseek.ToolDefinition] {
 ///|
 /// Dispatches a parsed tool call against `tools`. Returns an error response
 /// action when no definition matches.
+///
+/// A tool's own execution failure is caught HERE and turned into an error
+/// `ToolAction`, so the model sees it as an ordinary (recoverable) tool result
+/// and the turn continues — one boundary that spares every tool from wrapping
+/// each of its own filesystem/IO calls just to report a failure.
+///
+/// Cancellation is the exception: it is re-raised so the loop can abort the
+/// turn (see the terminal handler in `AgentLoop`, which maps it to
+/// `Interrupted`). That is also why this function stays raise-able rather than
+/// `noraise`: since `async` defaults to raising, marking it `noraise` would
+/// force cancellation to be caught and swallowed here too, and the turn would
+/// no longer be cancellable mid-tool-call.
 pub async fn execute_tool_call(
   call : AgentToolCall,
   tools : Tools,
 ) -> ToolAction {
   match tools.find(call.name) {
-    Some(tool) => tool.execute.run(call.arguments)
+    Some(tool) =>
+      tool.execute.run(call.arguments) catch {
+        error =>
+          if @async.is_being_cancelled() || @async.is_cancellation_error(error) {
+            raise error
+          } else {
+            ToolAction::respond(
+              "error: \{call.name} failed: \{error}",
+              is_error=true,
+            )
+          }
+      }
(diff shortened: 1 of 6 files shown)


bobzhang:
Reviewed the boundary design — it's sound (recoverable tool faults, cancellation re-raised via both checks, the noraise rationale is right per the async lib's own docs on `is_cancellation_error` accuracy). Generalized it on top (commit e633aa5e):

- **read/write/edit** drop their pure-rephrase IO catches (write and edit unwrap whole-body try/catch); the raw OS error already names the path.
- **Boundary improvement**: it now derives a `brief` from a `path` argument, so IO-fault results still label the transcript line ("edit ghost.mbt") — the edit tests caught that the local catches were providing this and the boundary wasn't.
- **multi_edit is deliberately untouched**: its catches are load-bearing (per-file failure aggregation, the "earlier files in the batch may already be written" partial-apply warning, the revert sweep, edits-file vs target-file attribution). The generic boundary message cannot express those.
- IO-fault tests now route through `execute_tool_call` — the production path — instead of poking executors directly.

Full suite 1173/1173.

bobzhang:
@codex review

bobzhang:
**Codex sign-off** (local `subal review --base a398a994`, head e633aa5e):

> The centralized dispatch boundary correctly converts non-cancellation tool failures into recoverable error actions while preserving cancellation propagation. Targeted agent and agent_tool tests pass, and no regressions were identified.

Merged on that verdict + 8/8 CI green.

PR #526, opened 2026-07-18
test(agent_tool): scope the always-failing executor inside its test

bobzhang:
Follow-up to #524: `async_boom` moves from file scope into the dispatch-boundary test as a local fn, so the always-failing executor cannot be referenced from anywhere else in the package. `moon test -p agent_tool` 486/486.


bobzhang:
**Codex sign-off** (local `subal review --base origin/main`):

> The change only scopes an always-failing test helper locally without altering behavior. MoonBit checks and all targeted agent_tool tests pass.

Will merge once CI settles.

PR #527, opened 2026-07-18
docs+refactor(shell): call-flow map, cwd guards, strip_fd_prefix

bobzhang:
## Summary

Documentation and two small refactors in the `shell` tool package.

### Docs
- **`shell`** gets a one-line summary plus a compact ASCII call-flow skeleton — the two forks that make the flow deep (background/foreground, session-group/per-call) and the `ShellRunResult` variants. Plain `///` lines (no code fence) so the doc-test harness ignores it and it reads as monospace in source and `git diff`, matching the repo's in-source-diagram convention.
- **`direct_source_write_redirect_target`** documented: what it flags (the write-redirect forms `redirect_writes_file` recognizes), the `Simple`-only parse limitation, and that the sandbox's `deny file-write*` rules are the runtime backstop for what the parse-based pass can't see.

### cwd validation (`shell`)
Replaced the two `exists`/`kind` if-blocks with two `guard … else` statements. This keeps the **specific** error messages (`cwd does not exist` / `cwd is not a directory`) while dropping the `let exists`/`let kind` temporaries and the negations. `@fs.kind` still runs only after `exists` passes.

> A `guard exists && kind is Directory` one-liner was considered but merges the two diagnostics into one message (`does not exist or is not a directory`) — that both loses signal for the agent and breaks `shell_test.mbt:1956`, which asserts the specific `cwd is not a directory` string. Two guards keep both messages, so the tests pass untouched.

### `strip_fd_prefix` (`sandbox.mbt`)
Reimplemented with `lexscan` / `re"^[0-9]+"` returning a `StringView` — drops the manual digit-loop and the `.to_owned()` allocation. Behavior is unchanged across the redirect ops (`"2>"→">"`, `">"→">"`, `"22>>"→">>"`, `""→""`, `"2>&"→">&"`).

## Test plan
- `moon fmt --check`, `moon check agent_tool/shell --deny-warn` — clean
- `moon test -p agent_tool/shell` — **190/190**
- No `.mbti` drift

### DIFF
--- a/agent_tool/internal/auto_check/moon.pkg
+++ b/agent_tool/internal/auto_check/moon.pkg
@@ -5,6 +5,7 @@ import {
   "moonbitlang/async",
   "moonbitlang/async/fs",
   "moonbitlang/core/json",
+  "moonbitlang/core/string",
 }
 
 import {
--- a/agent_tool/internal/auto_check/parse_gate.mbt
+++ b/agent_tool/internal/auto_check/parse_gate.mbt
@@ -326,7 +326,7 @@ fn loc_line_span(loc : String) -> (Int, Int)? {
       let end_ = @string.parse_int(end_) catch { _ => return None }
       Some((start, if end_ < start { start } else { end_ }))
     }
-    (re"^[0-9]+" as start) + re"(:.*|-)?$" => {
+    (re"^[0-9]+" as start) + re"(:.*|-.*)?$" => {
       let start = @string.parse_int(start) catch { _ => return None }
       Some((start, start))
     }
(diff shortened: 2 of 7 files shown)


bobzhang:
**Codex sign-off** (local subal, four rounds across the branch, head 04054441):

- Original PR: *"refactors preserve existing redirect and cwd-validation behavior"*
- loc/validator generalization: caught one real regression (`7-x` lost its fall-back-to-start) — fixed, and a 12-form fidelity table now pins every lenient loc shape
- Final aggressive batch: *"the lexscan refactors preserve the prior matching behavior; targeted tests pass across all supported targets"*

Branch totals: 9 hand-rolled scanners → lexscan (probe-verified against the old implementations on adversarial tables), full suite 1174/1174. Principled keeps documented in the last commit message (shared-constant prefixes, the shell lexer itself).

PR #528, opened 2026-07-19
refactor: =~ and lexscan for remaining find/slice path and marker code

bobzhang:
## Summary

High-ROI continuation of #527's lexscan work using the single-case `xx =~ (re"...", before~, after~)` form where one pattern suffices:

- **Path helpers** (5 sites, 2 helper fns deleted): `brief_basename`, write's `parent_dir`, viz_server's `path_tail`, viz_app's `path_basename`/`parent_dir` — the dual-separator `rev_find` index dances become one greedy `^(.|\n)*[/\\]` pattern each.
- **`redact_url`**: six chained find/slice steps → one `=~` cut at the first `?`/`#` plus a three-segment lexscan dropping `userinfo@`; the scheme arm requires `://` before any `/`, so an `@` in the path never redacts. Probe-verified equivalent on 14 adversarial URLs.
- **`failure_message` family unified**: the two remaining copies (tool_error, agent's goal copy) now share cmd/openseek's greedy-to-last-marker lexscan shape.

Every conversion was probe-verified old-vs-new on an adversarial table before landing; js target confirmed (viz_app bundle builds).

Considered and kept as-is: `goal_marker` and session-filename parsing (shared named constants), the shell lexer, single `strip_prefix` one-liners.

## Test plan
- Full suite 1174/1174; `moon check --deny-warn` clean; `moon build cmd/viz_app --target js` clean

### DIFF
--- a/agent/goal.mbt
+++ b/agent/goal.mbt
@@ -323,12 +323,12 @@ fn decode_goal_status(arguments : Json) -> Result[@goal.GoalStatus, String] {
 /// Tool packages normally share `agent_tool/internal/error`, but the agent is
 /// outside that internal visibility boundary.
 fn goal_decode_failure_message(error_text : String) -> String {
-  let marker = " FAILED: "
-  guard error_text.rev_find(marker) is Some(cut) else { return error_text }
-  let message = error_text[cut + marker.length():]
-  match message.strip_suffix(")") {
-    Some(trimmed) => trimmed.to_owned()
-    None => message.to_owned()
+  lexscan error_text {
+    // Greedy across lines to the LAST ` FAILED: ` marker — the same piece
+    // rev_find chose.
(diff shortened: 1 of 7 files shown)


bobzhang:
**Codex sign-off** (local subal, two rounds, head 0e6762c5):

Round 1 caught a real P2: the rewritten `redact_url` was fail-open for malformed-but-accepted endpoints (`/https://user:secret@host/mcp` would have echoed the credential). Fixed — first-`://` search restored, userinfo arm start-anchored (an unanchored lexscan arm SEARCHES; that footgun is exactly what the anchoring convention prevents), equivalence table extended to 17 URLs.

Round 2: *"The regex and lexscan replacements preserve the previous path, error-message, and URL-redaction behavior. No actionable regressions were identified."*

PR #531, opened 2026-07-20
feat(agent_tool): run_moonbit — run a MoonBit snippet in an isolated package

bobzhang:
## Summary

New registry tool (**twelfth built-in**): compile and run a self-contained MoonBit program and return its merged stdout/stderr and exit status.

**Motivation:** make MoonBit the agent's *default automation language* over shell `python`/`node` — it keeps automation in MoonBit (a dogfooding fluency flywheel), avoids shell quoting, and is the natural home for the planned **safe wasm backend**. Also serves language/API probes.

**Design (simplified after review + discussion):** the tool routes to moon's own `.mbtx` single-file runner rather than hand-scaffolding a module. Its whole job is: write `source` to a throwaway temp `.mbtx`, run `moon run <file>.mbtx --target <target>`, clean up (cancel-safe), and report.
- **`source`** is a `.mbtx` script: an optional inline `import { "a", "b" }` block (comma-separated), then the program with its own `main`. `async fn main` + `"moonbitlang/async"` for `@fs`/`@stdio` IO.
- **`target`** (optional, default `native`; validated `native`/`wasm`/`wasm-gc`/`js`/`llvm`) — the seam for the safe wasm backend.
- **60s timeout** so a runaway loop or stuck build cannot block the turn (cancellation tears down the subprocess).
- **Cleaner errors:** the throwaway temp path is rewritten to `source`, so diagnostics read `source:LINE:COL` about the model's input (line:col verified faithful even with an import block); moon's own compiler diagnostics are the message.
- **Honest isolation:** it runs outside the working-tree module, so a local-package import binds the published *registry snapshot* (not uncommitted edits) — for working-tree code, use a `*_test.mbt` + `moon test`.

**Safety note:** the snippet runs on the native backend today with no capability sandbox — isolation is the throwaway package. Hard containment is deferred to the planned wasm backend (safe by construction; policy applied there), per the design direction. Process spawning is simply not reachable (the shell tool remains the sandboxed command surface).

Also: **system prompt** updated to steer automation/probes to `run_moonbit` over shell `python`/`moon run -e`; READMEs for the tool and its decode package; tool lists updated to 12.

## Test plan

- 12 tool + 5 decode tests: pure run, compile-error via moon diagnostics, file read via inline import block, non-native target, timeout-arg validation, -path rewrite.
- `moon check --target all`, `--deny-warn`, full suite (1186/1186), cram (24/24), desktop check — clean.


bobzhang:
@codex review

PR #534, opened 2026-07-21
fix(agent): salvage any control-declared completion at the context ceiling

bobzhang:
## What

The context-ceiling salvage honored only a tool literally named `finish`. Two consequences:

1. A completion arriving through a **custom control tool** was skip-closed at the ceiling. Concretely: `agent_review`'s `submit_review` — a review that hits the ceiling on its submitting step loses its finished report and fails with `ReviewError`.
2. A **lone `goal(met)` verdict** without a trailing `finish` in the batch was skip-closed, resurrecting a completed goal for auto-continue — the exact hazard the prefix-execution comment already warned about, but only defended when a `finish` followed in the same batch.

## How

- `salvage_finish_at_ceiling` → `salvage_control_calls_at_ceiling`: a single walk over the unexecuted remainder that executes every control-registered, decodable call (bounded to a few bytes by the `control` contract) and skip-closes the rest. A sealing control (`Control(Finish)`/`Abort`) owns the terminal exactly as before; controls that only record results (goal verdicts, `Respond`-answering overrides) execute, then the turn yields at the ceiling with their durable results carried into the checkpoint. Returns `None` (caller's plain-yield path) only when the batch carries no control-registered call.
- `submit_review` now declares `control=true`.

Existing semantics preserved: `[probe, finish, probe]` salvage, `[goal, finish]` salvage-both, the non-control `finish` shadow being skip-closed, and steer-beats-finish all pass unchanged.

Behavior change (intended): control calls in a ceiling batch now execute even when no `finish` follows — previously they were skip-closed with the rest.

## Tests

- `a pressured custom control completion is salvaged at the ceiling` — seals with the custom tool's answer; probes never execute; checkpoint still runs.
- `a lone control verdict without a finish executes then yields` — verdict recorded durably, non-control call skip-closed, `[context ceiling]` yield.
- `submit_review declares itself a control tool`.
- `moon check --target all` clean; `agent` suite 69/70 (the 1 failure is the known-flaky background-job timing test, fails identically on clean main); `agent_review` 10/10; `moon fmt`/`moon info` clean (no public API change).

Context: PR A1 of the subagent-substrate series (explore tool + review gate), sequenced and design-reviewed via subal.

### DIFF
--- a/agent/auto_compact_test.mbt
+++ b/agent/auto_compact_test.mbt
@@ -586,6 +586,149 @@ async test "a pressured finish batch salvages the completion" {
 }
 
 ///|
+/// A completion tool under a custom name: control-declared, seals via
+/// Control(Finish) exactly like the built-in `finish` — the salvage matches
+/// the control declaration, not the name (a sub-run's submit tool is the
+/// production shape).
+async test "a pressured custom control completion is salvaged at the ceiling" {
+  @async.with_task_group() <| group => {
+    let checkpoint_requests : Array[Json] = []
+    let mut main_requests = 0
+    let handle = request => {
+      if is_checkpoint_request(request) {
+        checkpoint_requests.push(request)
+        Body(checkpoint_body("SUBMIT-SUMMARY"))
+      } else {
+        main_requests += 1
+        Sse(
+          sse_tool_calls(
+            [("call_1", "probe"), ("call_2", "submit_report")],
+            900_000,
+          ),
+        )
+      }
+    }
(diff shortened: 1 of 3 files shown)


bobzhang:
Addressed the subal review finding (P2: salvaged control output unaccounted): the salvage walk now threads a running `spent_result_chars` total, measuring each non-sealing control's recorded result after `ContinueToolBatch` — the same accumulation pattern `handle_tool_calls` uses — so a later sealing control sizes its compact-on-finish checkpoint against actual appended growth. Controls remain exempt from the pre-execution fit check only.

PR #535, opened 2026-07-21
feat(agent): route all turn events through a per-loop EventSink

bobzhang:
## What

PR A2 of the subagent-substrate series. Every protocol event the loop produces currently goes straight to the process-wide log. A nested agent loop (a sub-run such as the review engine running inside a live session) therefore cannot exist safely: its ordinary `AgentFinished` lands in the host-facing stream, and the TUI engine client (`engine_client.mbt:171`), TUI loop, and desktop host (`engine.mbt:363,748`) all treat any `AgentFinished` as the **outer** run's terminal. Event isolation is the prerequisite for shipping any subrun tool.

## How

- New `@emit.EventSink`: wraps `(Event, SourceLoc) -> Unit`; its `emit` autofills the **caller's** location so forwarded log lines keep pointing at the reporting code, never the sink plumbing. `EventSink::global()` forwards to `emit` — the behavior-preserving default.
- `AgentLoop` carries an `events` sink used by every emit site in the loop (`turn_loop`, `auto_compact`, `goal`). `GoalCadence::observe_notice` and the two non-method helpers (`log_and_reasoning_content`, `log_tool_result`) take the sink as a parameter.
- `run_turn_in_scope` gains `event_sink?` (default global) and routes the pre-loop `AgentSetupFailed` through it — a setup failure must not leak past an isolating sink either.
- The global xlog handler is never swapped; isolation is per-loop by construction.

No behavior change for existing callers: the default sink is exactly today's global emit.

## Tests

- `protocol/emit`: a custom sink receives the event + the caller's location; the global sink writes the same wire line with caller `source` attribution.
- `agent/event_sink_test.mbt`: a full mock-server turn's events (`agent_step` / `assistant_message` / `usage` / terminal `agent_finished`) all arrive at a collecting sink; `tool_result` events from a tool batch route through it too.
- `moon check --target all` clean; agent suite 71/72 (the 1 failure is the known-flaky background-job timing test, pre-existing on main); emit 8/8; agent_review 10/10; `fmt`/`info` clean (`agent/pkg.generated.mbti` picks up the new optional parameter).

### DIFF
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -226,6 +226,12 @@ async test "run_turn_with_append closes failed persisted turns" {
 /// a fresh registry for this call only. `api_url` is forwarded to the DeepSeek
 /// client. The caller owns the lifetime of `runtime`, `scope`, and any shared
 /// `tools` registry passed here.
+///
+/// `event_sink` is where every protocol event this turn produces goes —
+/// including the pre-loop setup failure. The default forwards to the process
+/// log (today's behavior). A nested sub-run passes an isolating sink so its
+/// lifecycle events (most critically `AgentFinished`) stay out of the
+/// host-facing stream, where they would read as the OUTER run's terminal.
 pub async fn[X] run_turn_in_scope(
   runtime~ : @agent_runtime.AgentRuntime,
   scope~ : @agent_runtime.AgentTaskScope[X],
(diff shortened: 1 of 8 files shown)


bobzhang:
Superseded by the process-based subagent architecture (#541): a sub-run now lives in a dedicated child process, so host-stream isolation is structural and the per-loop sink is unnecessary. Branch retained.

PR #536, opened 2026-07-21
feat(agent_subrun): extract the sub-run substrate; refactor agent_review onto it

bobzhang:
## What

PR B of the subagent-substrate series (stacked on #535 / PR A2 — the diff here is the top commit only). Extracts the nested-ephemeral-turn shape out of `agent_review` into a reusable `agent_subrun` package; `explore` (PR D) and the goal-review gate (PR F) are its next clients.

## Design (per the subal-reviewed plan)

- **`SubrunSpec[T]`** — session id, prompts, a tools *builder* that receives the runner-owned capture ref (so the registry and result channel cannot disagree), step ceiling, optional wall deadline.
- **`SubrunResult[T]`** — captured value, terminal (`Captured | NoReport | MaxSteps | ContextYield | TimedOut | Failed`), plus the steps and provider tokens the child actually spent, observed from the child's own event stream via the A2 sink — the data budget reconciliation and eval attribution need.
- **Event isolation by default** — the child's events forward to the caller's `event_sink`, defaulting to a suppressing sink. A child's `AgentFinished` must never reach the host stream (TUI/desktop treat any `AgentFinished` as the outer run's terminal).
- **Cancellation contract** — external cancellation re-raises, never folds into a terminal; the runner-owned deadline cancels the child internally and absorbs into `TimedOut`; a value captured just before the deadline still reports `Captured`.
- **`capture_tool[T]`** — the generalized submit tool: parse/validate, reject-with-retry, capture + `Control(Finish)`, `control=true` so the A1 ceiling salvage honors a pending capture.

`agent_review` refactors onto the substrate with its public contract unchanged: same report type, same `ReviewError` text, global sink (the CLI's `DiscardLog` still governs visibility), no deadline (its historical unbounded behavior).

## Tests

Mock-endpoint lifecycle suite in `agent_subrun` (9 tests): capture + exact cost accounting + forwarding, reject-then-retry, `NoReport`, `MaxSteps`, **ceiling salvage of a pending capture** (the A1 interlock), `ContextYield`, wall-deadline `TimedOut`, **external-cancellation re-raise**, toolset-failure `Failed`. `agent_review` 10/10 unchanged; `moon check --target all` clean; `fmt`/`info` clean.

Note: subal per-commit review is pending — quota exhausted until Jul 26 (#535 has the same caveat).

### DIFF
--- a/agent_review/engine.mbt
+++ b/agent_review/engine.mbt
@@ -8,7 +8,7 @@ pub suberror ReviewError {
 
 ///|
 /// Default step budget for one review run.
-let default_review_max_steps : Int = 120
+pub let default_review_max_steps : Int = 120
 
 ///|
 /// Run a code review of the changes between `base` and HEAD and return the
@@ -19,6 +19,12 @@ let default_review_max_steps : Int = 120
 /// pre-build source generation, so the checkout is not guaranteed byte-for-byte
 /// unchanged. Raises `ReviewError` if the model finishes without submitting a
 /// report.
+///
+/// Runs on the `@agent_subrun` substrate: an ephemeral child session whose
+/// report is captured through the `submit_review` control tool. The child's
+/// events go to the process log as before (the review CLI installs a
+/// discarding handler; a future in-session gate passes an isolating sink and
+/// a wall deadline instead).
 pub async fn run_review(
   api_key : String,
   model : @deepseek.Model,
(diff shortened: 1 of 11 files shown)


bobzhang:
Superseded by #541: same contracts (capture_tool, SubrunResult terminals, lifecycle test suite) rebuilt on the process runner instead of the in-process core.

PR #548, opened 2026-07-21
Subagent transcripts: persist child sessions and visualize them (chips, inline nesting, sidebar tree)

bobzhang:
Two halves, one feature: make subagent (sub-run) transcripts real, then make them visible.

**Substrate** (`da5a7df6`): a child of a durable parent persists its own transcript as a sibling session `<parent id>-sr-N` — the runner appends `--session`/`--session-root` to the child's argv after allocating the sub-run id, and the child mode wires a store-backed `append_item` into `execute_kind`. Transcripts append as the turn runs (crash-honest: a child that dies at its ceiling leaves the partial log you need to diagnose it). The sub-run id is stamped into the explore/review result briefs — display-only; the model never sees briefs. Ephemeral parents keep children in-memory as before.

**Viz** (`5728700a`): explore/review result cards grow a `↳ subagent` chip navigating to the child session (`#s=<child id>`, with a hashchange listener + bare-id→opaque-key resolution so it works in the live server AND standalone exports); the raw view nests the child's full transcript inline inside the parent's tool card (prefetched on session load, folded behind a step/tool-count summary); the model view stays chip-only (the child's turn is nothing the parent's model was fed — projection honesty); the sidebar nests `-sr-N` sessions under their parent.

Validation: native 1259/1259, js 54/54 (new snapshot test pins chip/nesting/model-view-abstinence/scanner), cram 26/26, plus an end-to-end headless-Chrome check on a standalone export of a synthesized parent+child store — chip href, nested transcript, sidebar nesting, and bare-id hash restore all render server-less.

### DIFF
--- a/agent_explore/moon.pkg
+++ b/agent_explore/moon.pkg
@@ -1,4 +1,5 @@
 import {
+  "bobzhang/openseek/agent_session",
   "bobzhang/openseek/agent_subrun",
   "bobzhang/openseek/agent_tool",
   "bobzhang/openseek/agent_tool/read",
--- a/agent_explore/pkg.generated.mbti
+++ b/agent_explore/pkg.generated.mbti
@@ -2,6 +2,7 @@
 package "bobzhang/openseek/agent_explore"
 
 import {
+  "bobzhang/openseek/agent_session",
   "bobzhang/openseek/agent_subrun",
   "bobzhang/openseek/agent_tool",
   "bobzhang/openseek/deepseek",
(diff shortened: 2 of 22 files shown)


bobzhang:
Known limitation (from K3's delta review, accepted): two engine processes concurrently resumed on the SAME durable session can both scan the same child-ordinal floor and race to allocate one child id — the loser's first append fails loudly on the store's stale-snapshot check (no clobber, no fork). Running two engines on one session is already unsupported; recorded here for the record.

PR #551, opened 2026-07-22
agent: adjacent concurrent-safe tool calls execute in parallel (parallel subagents)

bobzhang:
Two subagent delegations in one batch used to serialize (launch, wait minutes, reason, launch again). Sub-runs are isolated read-only child processes, so adjacent calls can safely overlap.

- `AgentToolDefinition.concurrent_safe` — built-in-only privilege like `control`: no workspace/loop-state mutation, no `Control` actions, no sibling-result dependence. `explore` + callable `review` declare it; explore's description invites batching independent questions.
- The loop prefetches maximal runs of ADJACENT flagged calls concurrently; the unchanged sequential batch loop then consumes cached actions — recording, budgets, ceiling handling, and result order are byte-identical, only the waiting overlaps.
- No wasted child spend: a group prefetches only when the whole batch prefix through its last call fits the running batch budget worst-case (otherwise it falls back to today's sequential path).

Validation: full suite 1263/1263 native, cram 26/26; new tests pin the overlap (`start-a,start-b,end-a,end-b` + call-order results) and the sequential default for unflagged tools.

### DIFF
--- a/agent/concurrent_tools_test.mbt
+++ b/agent/concurrent_tools_test.mbt
@@ -20,7 +20,7 @@ fn probe_tool(
   journal : Array[String],
   concurrent_safe~ : Bool,
 ) -> @agent_tool.AgentToolDefinition {
-  @agent_tool.AgentToolDefinition(
+  AgentToolDefinition(
     name="probe",
     description="Probe tool.",
     schema=JsonSchema({ "type": "object" }),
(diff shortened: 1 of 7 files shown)


bobzhang:
Correction to the PR description: the initial push was 1263/1264 — the one failure was agent_tool/read's README doc-test snapshot, which derives Debug over AgentToolDefinition and needed the new `concurrent_safe` field. Snapshot updated in the follow-up commit; full native suite is now genuinely green.

PR #554, opened 2026-07-22
run_moonbit: explain a sandbox source-write denial instead of a bare EPERM

bobzhang:
## Problem

When a `run_moonbit` snippet writes a protected source file, the sandbox kills the write and the tool result is just:

```
[run_moonbit: exited 1]
OSError("@fs.open(): \"keep.mbt\": Operation not permitted")
```

Nothing says the sandbox caused it, so the model reads a filesystem fault and either debugs the wrong thing or starts probing for ways around the block.

## Change

- A sandboxed run whose output shows a source-write denial now gets an appended explanation: the block is the tool's deliberate source-write-readonly policy (not an environment fault), the snippet must NOT try to work around it (renames, copies, temp-then-move), and source changes belong to the line-anchored `edit` tool. The run is reported `is_error` even if the snippet swallowed the failure and exited 0 — the same semantics `shell` applies.
- The text-side denial detector (`source_write_denied_in_text` + line helpers) moves from `agent_tool/shell` to the shared `agent_tool/internal/sandbox`, since both tools run under the same profile; `shell`/`shell_output` now delegate to it.
- **Real bug found while verifying**: MoonBit's `OSError` Show output escapes the quotes around the path (`keep.mbt\":`), so the detector's bare-quote pattern (`subject + '"'`) never matched run_moonbit's own denial shape. The detector now also recognizes the backslash-escaped-quote form — pinned against a live `sandbox-exec` reproduction.
- Line-level detection pins move to the internal/sandbox wbtest; the shell wbtest keeps only the sandboxed-command gate it still owns. New e2e test: an uncaught denial yields the raw OS error **plus** the sandbox explanation, and the protected file stays untouched.

## Verification

- Reproduced the bare-EPERM failure with a live `sandbox-exec` run before the fix; the new e2e test covers the exact shape.
- `moon check --deny-warn` clean, `moon fmt`/`moon info` fresh, affected package tests: sandbox+run_moonbit 20/20, shell+shell_output 129/129.

### DIFF
--- a/agent_tool/internal/sandbox/pkg.generated.mbti
+++ b/agent_tool/internal/sandbox/pkg.generated.mbti
@@ -26,6 +26,8 @@ pub async fn sandbox_exec_available() -> Bool
 
 pub fn sbpl_string_escape(String) -> String
 
+pub fn source_write_denied_in_text(String, Array[String]) -> Bool
+
 pub async fn source_write_readonly_profile_data(String, writable_lab? : String) -> SourceWriteProfileData
 
 // Errors
--- a/agent_tool/internal/sandbox/sandbox.mbt
+++ b/agent_tool/internal/sandbox/sandbox.mbt
@@ -554,6 +554,82 @@ pub fn sbpl_string_escape(text : String) -> String {
 }
 
 ///|
+/// Whether `output_text` shows a source-write denial from the sandbox: a
+/// "not permitted"/"denied" line naming a protected source suffix or one of
+/// the profile's `denial_subjects`. This is the text-side counterpart of the
+/// source-write profile above — tools that run under it (`shell`,
+/// `run_moonbit`) use it to recognize a denial in command output and attach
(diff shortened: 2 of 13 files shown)


bobzhang:
Addressed the subal P2 in cb1a71ac's follow-up: the denial scan now reads the whole on-disk log in bounded memory (chunked, line-carry) instead of the 48KB display prefix, and applies to the timeout arm too. New e2e test floods ~200KB before the denied write and asserts the guidance still lands while the displayed output shows only the truncated flood.

bobzhang:
Addressed the second subal finding (P1: unbounded post-deadline scan): the denial scan now reads at most 16MB from each end of the log via positional `read_at`. Rationale: an uncaught denial crashes the snippet so its OSError is the log's tail; a printed caught denial lands where it happened (head, for realistic probes). A wbtest pins the budget behavior with a tiny test budget (head/tail found, skipped middle missed by design, small logs scanned whole).

PR #556, opened 2026-07-22
run_moonbit: warning field, default off

bobzhang:
## Change

`run_moonbit` snippets are throwaway scripts, so compiler warnings (unused-value style noise) trip constantly, burn context, and can drown the program's real output. The tool now passes `--warn-list -a` to `moon run` by default, and gains an optional `warning` argument:

- `"off"` (default): warnings suppressed at the compiler.
- `"on"`: warnings shown — for when the warnings themselves are what's being probed (they get the same `source:LINE:COL` diagnostic rewrite as errors).

Errors are unaffected either way. Documented in the schema, tool description, doc comment, and README.

## Verification

- Probed live: `moon run snippet.mbtx --warn-list -a` suppresses the unused-variable warning; without the flag it renders before program output.
- e2e tests: default hides `Warning`, `warning:"on"` shows `Unused variable` with the `source:` rewrite; unknown values rejected at decode with a field-naming error. Decode unit tests cover default/on/off/invalid.
- `moon check --deny-warn` clean, `moon fmt` fresh, run_moonbit + decode tests 25/25.

### DIFF
--- a/agent_tool/run_moonbit/README.mbt.md
+++ b/agent_tool/run_moonbit/README.mbt.md
@@ -34,10 +34,13 @@ diagnostics are the error message when something does not build.
   the `shell` tool.
 - `warning` (string, optional, `"off"`/`"on"`, default `"off"`): whether
   compiler warnings appear in the output. Snippets are throwaway scripts, so
-  unused-value style noise is suppressed (a `--warn-list` spec disabling every
-  warn-state diagnostic) unless the warnings themselves are what you are
-  probing. Errors are unaffected — warning IDs whose default state is `error`
-  (e.g. `partial_match`) still fail compilation with warnings off.
+  unused-value style noise is suppressed — via a `--warn-list` spec derived
+  from the toolchain's own warning table (`moonc build-package -warn-help`,
+  probed once per process) — unless the warnings themselves are what you are
+  probing. Errors are unaffected: warning IDs whose default state is `error`
+  (e.g. `partial_match`) still fail compilation with warnings off, and if the
+  table cannot be probed the tool passes no flag at all, leaving warnings
+  visible rather than ever breaking the run.
 
 Only **dependency resolution** is isolated: a local-package import resolves to
 the **published registry snapshot** (not your uncommitted edits), and a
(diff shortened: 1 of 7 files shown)


bobzhang:
Addressed both subal P1s:

1. **Error-state diagnostics preserved** — reproduced the regression live (`--warn-list -a` let a non-exhaustive match compile and panic at runtime), and probed that `@id` re-enables do not survive `-a`. `warning: "off"` now uses explicit warn-state ranges (`-1..10-12..14-16..39-41..43-45..54-56..79`), keeping partial_match/unused_mut/multiline_string_escape/invalid_inline_wasm/unannotated_ffi at their default error state. New e2e test pins partial_match still failing with warnings off. IDs a future compiler adds past 79 fail open as visible warnings.
2. **decode `pkg.generated.mbti` regenerated** and committed (the `warning` field on `RunMoonbitInput`).

bobzhang:
Addressed the subal P2: the warn-off ranges now end at 83 — the bundled moonc's real max id (probed via moon run: 84 rejected, 83 accepted); ids 80..83 are all warn-state, so no new error-state exceptions. The quiet/loud e2e test now includes an `ambiguous_braces` (id 82) line to pin the 80..83 block.

bobzhang:
Addressed the subal P1 (static ceiling vs pinned desktop toolchain): the warn-off spec is now **derived at runtime** from `moonc build-package -warn-help` (probed once per process, from the same PATH toolchain as the spawned `moon`) — every listed id disabled by ranges that skip the error-state ids. On the desktop's pinned 0.10.2 this yields `..79`, on current nightly `..83`, and if the probe fails the tool passes no flag at all (fail-open to visible warnings, never a broken run). Parser pinned by wbtest incl. edge ids and garbage input.

bobzhang:
Addressed the follow-up subal findings: the warn-table probe now has its own hard 10s deadline (hard_cancel tears down a wedged moonc; falls back to passing no flag), and CRLF `-warn-help` output (Windows toolchains) parses correctly — trailing `\r` stripped per line, pinned by a wbtest case.

PR #557, opened 2026-07-22
edit: replace_all_preview — review every match as a multi_edit batch

bobzhang:
## Motivation

`replace_all` is rejected as a blind global replacement, but the agent had no way to *see* all match sites of a string before deciding what to change. That's a real gap for the edits the compiler cannot vet — comments, doc prose, `.mbt.md` text, string literals — and `moon check` after the fact only catches changes that stop compiling, not well-typed-but-wrong ones.

## Design: preview → select → apply → verify

`edit` now takes `replace_all_preview: true`. The file is **not** modified; the response lists every match of `old_string` in the `start_line`/`end_line` range — each site with 2 context lines and a `>` marker — and ends with a ready-to-review `multi_edit` `edits` array in its exact entry shape (`{file, old_string, new_string, start_line}`, anchors valid against the file as previewed). The agent keeps the entries to apply, drops the rest, and passes them to `multi_edit`, whose existing gates (pre-write parse gate, post-write `moon check` auto-revert) verify the applied batch.

The pipeline is complementary, not overlapping: the preview filters what the compiler cannot see; `multi_edit` verifies what it can. For typed-code renames the compiler loop (deprecated alias + fix what `moon check` flags) remains the better tool — the description, report footer, and README all say so, and a code-heavy match list in the preview is exactly the cue to switch to it before any damage.

## Mechanics

- Capped at 40 sites with a "narrow with start_line/end_line" nudge (one extra probe distinguishes "40+" from exact counts).
- Several matches on one line collapse into a single **whole-line** entry with all of them applied — two line-anchored edits with the same anchor would both hit the first occurrence. Noted inline in the report.
- Multiline matches render as `lines A-B:` spans; a second multiline match starting on the same line is reported as inexpressible rather than emitted wrong.
- Display lines truncate at 200 chars; the emitted entries always carry the exact strings.
- The `replace_all=true` rejection message now points at the preview.

## Verification

- e2e test round-trips the emitted array into `multi_edit`, applying a selected subset (middle entry dropped, file shows exactly the two selected replacements).
- Tests for range+same-line collapse, multiline spans, the 40-site cap, no-match and empty-old_string errors; decode snapshots updated.
- `moon check --deny-warn` clean, fmt/info fresh, edit+decode+multi_edit tests 80/80.


bobzhang:
Addressed both subal P2s: the cap now bounds emitted **sites** (same-line matches merge into the last site unbounded; only a match opening a new site past the cap stops the scan — regression test: a 45-match line no longer starves later sites, exact counts kept), and display truncation backs up one unit when the cut would split a surrogate pair (inline boundary test).

bobzhang:
Addressed the subal P1 (quadratic scan on dense lines): the match scan is now linear regardless of density — `StringView` find with no tail copies, incremental line counting, and whole-line entries built once at site finalization. A 100k-matches-on-one-line regression test runs with exact counts. Bonus hardening found while fixing: a many-match line longer than 4096 chars now yields a **note instead of an entry** — embedding a minified line twice in the edits array could push the report into the outer result clamp and cut the JSON mid-array, corrupting every entry.

bobzhang:
Addressed all three findings from the full-branch subal xhigh pass:

1. **Expansion estimated, never materialized** — the whole-line guard now bounds `line + match_count × growth` arithmetically before `replace_all` runs; a 100-match line with a 2000-char replacement yields a note, not a ~200k allocation (test).
2. **Report budgeted under the 60k tail clamp** — context blocks and the entries array each carry a 24k budget with counted, stated omissions; the array always closes validly (test asserts the emitted array parses even with entries omitted).
3. **Line-offset index** — one O(file) pass; finalization and context rendering look lines up in O(1) instead of rescanning from the start (~400 full scans worst case before).

bobzhang:
Addressed the follow-up subal P1: the expansion estimate is now Int64 — the 32-bit product of `match_count × growth` could wrap negative (4096 × 2^19 = 2^31) and pass the guard it implements, letting `replace_all` attempt a ~2GiB string. Test pins the exact overflow shape.

PR #562, opened 2026-07-23
Add editor as a monorepo module

bzy-debug:
## Summary

- add the editor source tree directly under `editor/` and register it in the root `moon.work`
- keep the editor import as regular monorepo source, without importing its `vscode` or `codemirror` gitlinks
- vendor the nine pinned VS Code mouse-wheel traces as MoonBit fixtures so classifier reference tests no longer depend on a VS Code checkout at runtime

## Why

This makes editor development part of the openseek workspace without a Git submodule boundary, while keeping its wheel-classifier reference coverage self-contained and reproducible.

## Validation

- `moon info && moon fmt`
- `moon check --target all`
- `moon test --target js` — 1821 passed, 0 failed
- `moon test editor/internal/viewer/ui/scrollbar/mouse_wheel_classifier_reference_wbtest.mbt --target js` — 9 passed, 0 failed


.github/workflows/publish-editor.yml
tonyfettes, line 1: I don't think this will work b/c moonbitlang does not share CI secrets with moonbit-community?

bzy-debug: Will remove it

PR #571, opened 2026-07-24
fix(desktop): allow inline composer completion

tonyfettes:
TL;DR: allow `#` and `$` to trigger symbol completion in a half-way drafted composition.

image

image

## Summary

- allow trailing, whitespace-delimited `$` skill and `#` symbol completion tokens after ordinary prompt text
- preserve the draft text before the trigger token when a completion is accepted
- use the same token parser for asynchronous symbol requests and replies while keeping slash commands whole-draft-only

## Root cause

Desktop completion only recognized a trigger when the entire draft began with `/`, `$`, or `#`. Accepting a skill or symbol also cleared the entire composer draft, so simply broadening the trigger check would have discarded the text before an inline token.

## User impact

Users can now type prompts such as `Please use $review-pr` or `Inspect #parse_hover`, accept the completion, and keep the preceding prompt text intact.

## Validation

- `moon -C desktop fmt --check frontend/completion.mbt frontend/model.mbt frontend/update.mbt`
- `moon -C desktop info`
- `moon -C desktop check --target js`
- `moon -C desktop test --target js` (226/226 passed)
- `moon -C desktop check --target native`
- `git diff origin/main...HEAD --check`

### DIFF
--- a/desktop/frontend/completion.mbt
+++ b/desktop/frontend/completion.mbt
@@ -1,8 +1,8 @@
-// The composer's completion menu: typing a leading `/` lists built-in
-// conversation commands, a leading `$` lists installed skills whose name is
-// dropped into the prompt. Both filter as the user types and are driven from
-// the model so the keyboard (↑/↓, Enter/Tab, Esc) and the popup view stay in
-// sync.
+// The composer's completion menu: a leading `/` lists built-in conversation
+// commands, while a trailing `$` or `#` token lists installed skills or
+// workspace symbols after ordinary prompt text. All three filter as the user
+// types and are driven from the model so the keyboard (↑/↓, Enter/Tab, Esc)
+// and the popup view stay in sync.
 
 ///|
 /// The built-in slash commands, in menu order. `command` is the id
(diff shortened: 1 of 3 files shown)


desktop/frontend/completion.mbt
```diff
@@ -21,14 +21,55 @@ fn slash_command_items() -> Array[CompletionItem] {
...
+/// mention preserves.
+priv struct CompletionTrigger {
+  kind : CompletionKind
+  query : String
+  task_without_token : String
+}
+
+///|
+fn CompletionTrigger::parse(text : String) -> CompletionTrigger? {
+  if re"^/(?<query>[^[:space:]]*)$".execute(text) is Some(result) &&
```
tonyfettes: Can we use `text =~ re".."` here?

desktop/frontend/completion.mbt
```diff
@@ -40,53 +79,68 @@ fn compute_completion(
...
-          range: None,
+    SkillMentions => {
+      let items : Array[CompletionItem] = []
+      for skill in installed {
+        if completion_matches(trigger.query, [skill.name, skill.description]) {
+          items.push({
+            label: "$" + skill.name,
+            detail: skill.description,
+            command: "",
+            insert: skill.name + " ",
```
tonyfettes: Use string interpolation.

PR #574, opened 2026-07-25
simplify(desktop/frontend): dedup the resize clamp; inline single-use helpers

bobzhang:
Part of a project-level simplification/dedup sweep, one PR per package group. This one covers `desktop/frontend` — the desktop module has never been swept.

## What

The editor side-panel resizer and the file-tree resizer are structural twins. Their clamp arithmetic was copied **three times verbatim** (editor width, tree width, tree height), carrying the same *"floor wins on a narrow window"* comment each time; `end_resize` and `end_tree_resize` were identical modulo which drag flag they read.

- extract `drag_size(raw, floor~, max~)` — replaces the three copies of the clamp
- extract `end_drag(active)` — replaces the two identical drag-enders
- both land in `editor_resize.mbt`, which already hosts the shared root-style writer, `set_root_class` and `schedule_remeasure`
- inline `parsed_markdown_nodes` into its one caller `markdown_nodes` (3 lines, reads as well inside the `@js.Error_::wrap` it guards)
- inline `scroll_transcript` into `scroll_transcript_after_render`, its only caller and a bare one-line wrapper

Net **-18 lines**.

## Deliberately kept

Scanned every single-use private fn in the package; these are house-rule exceptions, not misses:

- **symmetric families** — the nine `save_*` storage setters, the `*_msg` bridge decoders, `plain_text`/`tights_text`, the three `set_*` size setters, `request_hover`/`request_lsp_open`
- **documented policy predicates** — `is_submit_shortcut`, `is_absolute_path`, `is_idle_event`, `host_prefers_dark`, `host_is_windows`, `file_extension`, `diagnostic_severity`
- **model seams** used cross-file — `Model::settings_status`, `Conversation::last_assistant_content`, `Model::any_conversation_busy`

## Verification

- `moon -C desktop check --target js` — clean
- `moon -C desktop check --target native` — clean
- `moon -C desktop info --target native` — **no .mbti drift** (all changes are private)
- `moon -C desktop test --release` — **1511/1511 passed**

Note: `--deny-warn` is not clean on `main` right now, independent of this PR — the 2026-07-24 nightly toolchain deprecated two call sites (`agent_session/json.mbt:168`, `agent_tool/multi_edit/multi_edit.mbt:883`). Gated without the flag and confirmed this branch adds no new warnings.

### DIFF
--- a/desktop/frontend/dom_helpers.mbt
+++ b/desktop/frontend/dom_helpers.mbt
@@ -5,16 +5,16 @@ fn is_submit_shortcut(event : @dom.KeyboardEvent) -> Bool {
 
 ///|
 fn scroll_transcript_after_render() -> @cmd.Cmd {
-  @cmd.custom_cmd(_ => scroll_transcript(), kind=@cmd.after_render)
-}
-
-///|
-fn scroll_transcript() -> Unit {
-  match @dom.document().get_element_by_id("transcript").to_option() {
-    Some(transcript) =>
-      transcript.set_scroll_top(transcript.get_scroll_height())
-    None => ()
-  }
+  @cmd.custom_cmd(
+    _ => {
+      match @dom.document().get_element_by_id("transcript").to_option() {
+        Some(transcript) =>
+          transcript.set_scroll_top(transcript.get_scroll_height())
+        None => ()
+      }
+    },
+    kind=@cmd.after_render,
+  )
 }
 
 ///|
--- a/desktop/frontend/fileeditor/editor_resize.mbt
+++ b/desktop/frontend/fileeditor/editor_resize.mbt
@@ -64,7 +64,7 @@ fn request_resize_handle_mount() -> @cmd.Cmd {
       )
       let document : @js.Value = @js.globalThis.get_with_string("document")
       @interop.add_capture_listener(document, "mousemove", on_resize_move)
-      @interop.add_capture_listener(document, "mouseup", end_resize)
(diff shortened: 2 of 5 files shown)


bobzhang:

**VERDICT: APPROVE**

- `drag_size` is character-for-character identical to all three removed clamp copies; floor/ceiling ordering and the `.to_int()` conversion are preserved, including the narrow-window floor-wins edge case. Argument-order evaluation change is inert — `rect.get_right()/get_width()/get_height()` and `mouse.get_client_*()` are pure DOM getters with no interleaved mutation inside the synchronous handler.
- `end_drag(active)` faithfully parameterizes the only difference (which `Ref[Bool]` is cleared). The `_ => end_drag(resizing)` closure is semantics-neutral: `add_capture_listener` takes `(@dom.Event) -> Unit`, the event arg was already ignored (`_event`), and the refs are module-level `let` bindings that are never re-bound. Each resizer still clears only its own flag — no new cross-talk.
- The `parsed_markdown_nodes` inline keeps the entire cmark parse+render **inside** the `@js.Error_::wrap` panic boundary that the doc comment describes; `open_file?` flows into the struct unchanged.
- The `scroll_transcript` inline is verbatim, with `kind=@cmd.after_render` intact.
- Verified no dangling references to any removed symbol; `moon -C desktop check --target js` passes on the affected packages; no `.mbti` diff.

> Readability is a net improvement: three copies of a subtle clamp (with a non-obvious floor-wins rule) are now one documented function, and the drag-end logic exists once.

PR #575, opened 2026-07-25
simplify(desktop/internal): one home for the PATH-variable handling

bobzhang:
Part of a project-level simplification/dedup sweep, one PR per package group. Follows #574 (`desktop/frontend`).

## What

`internal/engine` and `internal/moonbit` both spawn children with an explicit environment block, and both had grown their own copy of the same three things:

| duplicated | copy 1 | copy 2 |
|---|---|---|
| `#cfg(platform="windows")` `PathSep` const pair | `engine/engine.mbt` | `moonbit/toolchain_paths.mbt` |
| `prepend_to_path_env` | `engine/engine.mbt` | `moonbit/toolchain_paths.mbt` |
| Windows `Path`-vs-`PATH` reconciliation loop | `engine`'s `prepend_path_env_entry` | body of `@moonbit.bundled_tool_env` |

The first two were byte-identical. The third was duplicated down to near-identical doc prose — and it is the risky kind: it encodes a subtle rule about `inherit_env=false` blocks (find the spelling that already exists, drop duplicate keys, write back under that spelling), so a fix to one copy would silently leave the other spawning children with two path variables and a coin toss over which one wins.

All three move into `internal/env` — already native-only, already imported by both packages, already the home for process-environment reads. `bundled_tool_env` keeps its signature and drops to four lines.

## Tests

The unit test for `prepend_to_path_env` moves with the code it now covers, and **gains coverage for the reconciliation rule it never had**: the existing-spelling-wins case, duplicate-key removal, and the empty block. Net +2 tests.

## Verification

- `moon -C desktop check --target native` — clean
- `moon -C desktop check --target js` — clean
- `moon -C desktop info --target native` — the **only** `.mbti` change is `internal/env` gaining the three shared symbols; `engine` and `moonbit` interfaces are untouched, confirming no public surface moved
- `moon -C desktop test --release` — **1513/1513 passed**

Note: `--deny-warn` is not clean on `main` right now, independent of this PR (two 2026-07-24 nightly deprecations in `agent_session` / `agent_tool`). Gated without the flag; this branch adds no new warnings.

## Also examined, deliberately not changed

Every single-use private fn in `desktop/internal/*` and `desktop/package/*` was scanned. The rest are house-rule keeps, not misses:

- **op-handler families** — the `*_op` handlers in `internal/extension` are a registration table
- **platform families** — the three `reset_bundle_dirs` (linux/macos/windows) share a name but have genuinely per-platform bodies; not duplication
- **path-accessor family** — `moonbit_bin_dir`, `moonbit_lsp_command`, `moonbit_seed_version_path`, the two bundle-probe paths
- **documented named concepts** — `openseek_client_token`, `resolved_run_cwd`, `resolved_global_skills_dir`, `prepare_engine_stdin`, `path_env_value`, `unquote_xdg_value`, `parse_outcome`

### DIFF
--- a/desktop/internal/engine/engine.mbt
+++ b/desktop/internal/engine/engine.mbt
@@ -467,26 +467,9 @@ fn resolved_global_skills_dir() -> @path.Path? {
 }
 
 ///|
-#cfg(platform="windows")
-const PathSep : String = ";"
-
-///|
-#cfg(not(platform="windows"))
-const PathSep : String = ":"
-
-///|
-fn prepend_to_path_env(bin_dir : String, current_path : String?) -> String {
-  if current_path is Some(path) && !path.is_empty() {
-    "\{bin_dir}\{PathSep}\{path}"
-  } else {
-    bin_dir
-  }
-}
-
-///|
 fn first_path_entry(path : String?) -> String {
   guard path is Some(path) else { return "" }
-  match [..path.split(PathSep)] {
+  match [..path.split(@env.PathSep)] {
     [first, ..] => first.to_owned()
     [] => ""
   }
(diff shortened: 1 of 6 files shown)


bobzhang:

**VERDICT: APPROVE**

- `@env.prepend_path_entry` is line-for-line equivalent to both removed copies (loop, duplicate removal, write-back key choice, return value); `bundled_tool_env` preserves the path-update → `MOON_HOME`-removal ordering.
- `#cfg` platform gating is evaluated at the defining package's compile for the target — moving the const across packages changes nothing.
- All call sites repointed (`first_path_entry`, engine test, toolchain env); no dangling references; `moon.pkg` deps already in place.
- `.mbti` diff is confined to `internal/env`; +3 symbols on an internal package is the fair price of removing two duplicate implementations.
- Tests assert the reconciliation rule correctly.

> Verified by inspection only; per the read-only constraint I did not run `moon check`/`moon test` (they would write to `_build/`).

### One nit raised — fixed in 34ed700f

> the "drops duplicate keys" test never creates a duplicate-key block, so the dedup branch remains uncovered — suggest adding a `{"Path": ..., "PATH": ...}` case in a follow-up

Correct, and the test name overclaimed. Rather than defer it, the case is now split in two: one test for the existing-spelling-wins rule, and a new one that really carries both `Path` and `PATH`. Since which spelling survives follows the block's own iteration order, the new test asserts the invariant — exactly one key left, holding the prepended value — instead of a particular winner.

Re-gated after the fix: `moon -C desktop check` clean, `moon -C desktop test --release` **1514/1514**.

PR #576, opened 2026-07-25
simplify(editor/syntax): hoist the lexer helpers the lang_* packages copy

bobzhang:
Part of a project-level simplification/dedup sweep, one PR per package group. Follows #574 and #575 (desktop). This is the first pass over `editor`, scoped to `syntax`.

## Why this shape

`syntax` **already** keeps a shared lexer helper for exactly this reason. `is_capitalized`'s doc comment reads:

> Shared by the language lexers (`lang_*`), which otherwise carry identical copies.

Two more helpers had drifted into that same copied state, so this PR follows the precedent the package already set rather than inventing a new home.

## What

| helper | was duplicated in | notes |
|---|---|---|
| `push_token` | `lang_moonbit`, `lang_javascript` | byte-identical; the two copies documented the same rule differently |
| `decode_stack` | `lang_moonbit`, `lang_javascript` | byte-identical |
| `encode_stack` | `lang_javascript` | the inverse of the above |

`push_token` encodes the coalescing rule — contiguous same-tag lexemes become one span, so a quoted string arrives as quote/content/quote and leaves as a single `String` token. Worth having once.

The stack codec operates on `TokenizerState`, which is a `syntax` type, so the encoding of that state belongs beside it rather than in whichever lexer needed it first. It is renamed **`decode_mode_stack` / `encode_mode_stack`**: `decode_stack` was clear enough as a file-local name but says nothing on a shared public surface, and the pair now travels together instead of leaving the inverse stranded in one lexer.

`lang_json` is deliberately untouched — it pushes tokens directly and does not coalesce.

Per this module's docs-are-contracts convention, `syntax/README.md` documents the shared helpers.

## Verification

- `moon check --target all --warn-list +73` — clean (zero warnings under `editor/`; the warn-73 hits in the output are pre-existing ones in `desktop/`, now visible because the workspace was unified)
- `moon fmt --check` on the touched packages — clean
- `moon test` over `syntax` + all three `lang_*` — **22/22 on `js` and 22/22 on `native`**. These are the golden token-stream tests: they render every token's offsets, tag and text for each lexer, so a behavior change in `push_token` or the state codec could not pass them unchanged.
- `moon info` — the only `.mbti` change is the three hoisted symbols appearing in `syntax`, beside `is_capitalized`

## Deliberately not changed

The three `lang_*/lexer_test.mbt` files carry a **35-line golden-rendering harness that differs by exactly one line** (which tokenizer it constructs). Deduping it needs a shared test-support package plus a `moon.mod` publish-exclude entry — a structural change to this module that I'd rather propose separately than smuggle into a dedup PR. Flagging it here as a known follow-up.

Also examined and kept as house-rule exceptions: `plain.mbt`'s `is_space` / `is_ident_continue` / `is_ident_start` / `is_digit` (char-class family), `lang_json`'s `colon_follows` (documented, and cited by the README), and the per-lexer `normal_step` / `comment_step` / `template_step` / `string_step` / `dollar_step` families.

### DIFF
--- a/editor/syntax/README.md
+++ b/editor/syntax/README.md
@@ -41,6 +41,14 @@ implementing `LineTokenizer`. Concrete languages are selected by hosts, examples
 or tests; reusable viewer core packages must not import them. There is no runtime
 grammar-loading or `setMonarchTokensProvider` equivalent.
 
+Helpers every stateful lexer needs live in this package rather than in each
+`lang_*`, which otherwise carry identical copies: `is_capitalized` (the
+identifier-is-a-type heuristic), `push_token` (append a token, coalescing it
+into the previous one when the tag and offsets are contiguous), and
+`decode_mode_stack` / `encode_mode_stack` (the nesting-mode stack a lexer
+carries between lines, packed into `TokenizerState`; an empty state decodes to
+the base mode `'n'`).
+
 When translating a Monarch or CodeMirror grammar:
 
 - Use `lexmatch ... with longest`; equal-length matches choose the earliest arm.
--- a/editor/syntax/lang_javascript/lexer.mbt
+++ b/editor/syntax/lang_javascript/lexer.mbt
@@ -37,7 +37,7 @@ pub impl @syntax.LineTokenizer for JavascriptTokenizer with fn tokenize_line(
   line_text,
   state,
 ) {
-  let stack = decode_stack(state)
+  let stack = @syntax.decode_mode_stack(state)
   let tokens : Array[@syntax.LineToken] = []
   let mut at = 0
   let mut view = line_text[:]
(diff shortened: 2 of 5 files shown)


bobzhang:

**VERDICT: APPROVE**

- Hoisted `push_token` / `decode_mode_stack` / `encode_mode_stack` bodies are byte-identical to the removed lexer copies (only redundant `@syntax.` qualifiers dropped); all call sites repointed, no leftover local definitions.
- `lang_moonbit`'s direct `TokenizerState("n")` return correctly left untouched; no behavior change anywhere.
- Public placement in `syntax` is necessary (cross-package visibility) and follows the `is_capitalized` precedent; `pkg.generated.mbti` and README updated in step; the `*_mode_stack` rename is well-justified.
- `lang_json` exclusion is correct — it has a two-state flag, not a mode stack.

> Tests not executed (read-only review); existing `lang_*/lexer_test.mbt` suites cover the hoisted paths end-to-end — CI should confirm.

(Locally those suites are green: 22/22 on both `js` and `native`.)

### Non-blocking observation worth recording

> note a pre-existing quirk: `lang_json`'s `comment_step` emits adjacent same-tag `Comment` tokens it never coalesces; a follow-up could adopt `@syntax.push_token` there.

Agreed, and deliberately out of scope here — adopting `push_token` in `lang_json` would **change its token output** (merging adjacent `Comment` tokens into one span), so it is a behavior change rather than a dedup, and it would rewrite that lexer's golden test. Worth a separate PR if the coalescing inconsistency between JSON and the other two lexers is unintended.

PR #579, opened 2026-07-25
fix: keep #cfg consts out of generated interfaces

bobzhang:
Self-review follow-up to the merged #575, extended after cross-model review. Fixes a whole hazard class rather than the one instance I introduced.

## The problem

`moon info` bakes a `pub const`'s **value** into the checked-in `pkg.generated.mbti`. When the const is `#cfg`-gated, only the branch that survived the current target's compile exists at generation time — so regenerating on another platform rewrites the line and the committed interface stops matching. CI's interface check runs on ubuntu, so the effect is a Windows contributor carrying a spurious diff, or turning the check red if they commit it.

A systematic sweep of every `pub const` reaching a committed `.mbti` finds exactly **two** `#cfg` cases:

| symbol | committed `.mbti` says | on Windows | introduced by |
|---|---|---|---|
| `desktop/internal/env` `PathSep` | `":"` | `";"` | #575 (mine) |
| `agent_tool/internal/sandbox` `PlatformShellProgram` | `"sh"` | `"pwsh"` | pre-existing |

The convention the repo already follows elsewhere is to keep such consts **private** — `ExecutableSuffix` in `desktop/internal/moonbit/toolchain_paths.mbt` sits under the identical `#cfg` pair and is private for exactly this reason. #575 broke that convention without noticing.

## The fix

Both become a private `#cfg` const behind a public accessor, so the interface records a *signature* rather than a value:

```diff
-pub const PathSep : String = ":"
+pub fn path_sep() -> String
```
```diff
-pub const PlatformShellProgram : String = "sh"
+pub fn platform_shell_program() -> String
```

Call sites repointed: `engine.mbt`'s `first_path_entry`, one engine test and the `internal/env` tests; `agent_tool/shell/shell.mbt` and `agent_tool/shell/sandbox.mbt`. No behavior change — the accessor returns the same const the callers were reading.

## Verification

- `moon check` (root) and `moon -C desktop check --target native` / `--target js` — clean
- `moon info` + `moon -C desktop info --target native` — both interfaces trade the const for the accessor; nothing else drifts
- `moon test` — **1514/1514 passed**

## Review note

Kimi K3 approved the first commit and caught that its "only such case in the repo" claim was wrong — my original grep for the hazard had been narrowed to separator characters and missed `PlatformShellProgram`. The second commit fixes that instance too and corrects the claim.

### DIFF
--- a/agent_tool/internal/sandbox/pkg.generated.mbti
+++ b/agent_tool/internal/sandbox/pkg.generated.mbti
@@ -2,11 +2,11 @@
 package "bobzhang/openseek/agent_tool/internal/sandbox"
 
 // Values
-pub const PlatformShellProgram : String = "sh"
-
 pub const SandboxExecPath : String = "/usr/bin/sandbox-exec"
 
 pub fn platform_shell_args(String) -> Array[String]
+
+pub fn platform_shell_program() -> String
 
 pub fn sandbox_exec_args(String, String) -> Array[String]
 
--- a/agent_tool/internal/sandbox/sandbox_exec.mbt
+++ b/agent_tool/internal/sandbox/sandbox_exec.mbt
@@ -152,22 +152,25 @@ async fn cleanup_sandbox_probe(
 }
 
 ///|
-/// The platform shell program used to interpret command text.
-///
-/// This is `pwsh` on Windows and `sh` on other targets. It is used by both
-/// `sandbox_exec_args` and callers that need the same unsandboxed command
-/// semantics.
 #cfg(platform="windows")
-pub const PlatformShellProgram : String = "pwsh"
+const PlatformShellProgram : String = "pwsh"
 
 ///|
-/// The platform shell program used to interpret command text.
-///
-/// This is `pwsh` on Windows and `sh` on other targets. It is used by both
-/// `sandbox_exec_args` and callers that need the same unsandboxed command
-/// semantics.
 #cfg(not(platform="windows"))
-pub const PlatformShellProgram : String = "sh"
+const PlatformShellProgram : String = "sh"
+
+///|
+/// The platform shell program used to interpret command text: `pwsh` on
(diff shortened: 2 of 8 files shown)


bobzhang:

Reviewed independently by **Kimi Code CLI (kimi-code/k3)**, read-only.

**VERDICT: APPROVE**

- The defect is real: `moon info` emits the surviving `#cfg` branch's value for a `pub const`; the pre-fix committed line `pub const PathSep : String = ":"` proves it, and a Windows regen would produce `";"`. K3 cross-checked the mechanism against non-conditional examples (`AppVersion = "0.1.4"`, `ExtensionId = "openseek-desktop"`) to confirm values really are emitted.
- The fix (private `#cfg` const + `pub fn path_sep() -> String`) is minimal, idiomatic, and matches the existing `ExecutableSuffix` convention; the `.mbti` now carries only a signature.
- All former `@env.PathSep` call sites (`engine.mbt:472`, `engine.mbt:943`, five in `path_test.mbt`) are repointed with identical values — no behavior change, no stragglers repo-wide.

### The finding I acted on

> Caveat (non-blocking, needs follow-up): the "only such case in the repo" claim is wrong — `agent_tool/internal/sandbox` has `#cfg` `pub const PlatformShellProgram` with `"sh"` baked into its committed `.mbti`. Same hazard.

Correct, and my fault: I'd grepped for the hazard with a pattern narrowed to separator characters (`"[;:]"`), which structurally could not have found `"sh"`. A proper sweep — every `pub const` in a committed `.mbti`, then back to its definition to check for `#cfg` — finds exactly two cases, and `PlatformShellProgram` is the other (`"pwsh"` on Windows).

Rather than reword the claim, the PR now fixes both: `PlatformShellProgram` gets the same private-const-plus-accessor treatment, its two callers in `agent_tool/shell` are repointed, and the PR is retitled around the hazard class. Re-gated: `moon check` clean, both interfaces regenerated, `moon test` **1514/1514**.

PR #584, opened 2026-07-25
ci: deny warnings on stable only, and give desktop a stable channel

bobzhang:
Two changes with one motive: nightly moves ahead of us, so it should report without gating.

## `--deny-warn` on stable only

Nightly introduces new warnings before there is anything to fix on our side, so denying them there turns the job red for code that is fine — and a job that is red for reasons outside the diff is a job people stop reading. Stable stays the gate that keeps warnings out of the tree.

Expressed as a matrix value rather than a duplicated step:

```yaml
- moonbit-channel: nightly
  moonbit-version: nightly
  deny-warn: ""
- moonbit-channel: stable
  moonbit-version: latest
  deny-warn: "--deny-warn"
```

## A stable channel for desktop

`desktop.yml` had **no stable coverage at all** — all five jobs installed nightly outright, so the desktop module's stable compatibility has never been tested, even though `ci.yml` has run both channels for the root module. `desktop-check` and `desktop-test` now run the same two-channel matrix, with the same `fail-fast: false` so a green-stable/red-nightly split is readable at a glance.

Resulting channels per job:

| job | channels |
|---|---|
| `check` (ci.yml) | nightly, stable **+deny-warn** |
| `desktop-check` | nightly, stable **+deny-warn** |
| `desktop-test` | nightly, stable |
| `desktop-appimage` / `desktop-macos` / `desktop-windows` | nightly only (unchanged) |

The packaging jobs stay nightly-only deliberately: they build shipping artifacts from one toolchain, and a second pass on the macOS and Windows runners costs real wall-clock for no signal about the code. Say the word if you want them doubled too.

The desktop `moon info` interface check becomes nightly-gated, matching the root CI — `moon info` renders abstract types differently across channels and the committed `.mbti` are nightly-generated, so on stable it would fail on formatting alone.

## Two things to know

1. **This PR's own run is the first time the desktop module is built on stable.** If it goes red there, that is genuine new information rather than a regression from this diff — and it is exactly the coverage gap this PR exists to close.

2. **CI is already red on `main`, unrelated to this.** On `ce80b4aa` (and on `de635e93` before it), `check (stable)` passes and `check (nightly)` fails at `moon fmt --check`: the nightly formatter disagrees with the committed `agent_tool/run_moonbit/README.mbt.md`. That is the same class of nightly churn this PR addresses for warnings, but `fmt --check` is a separate policy call, so I have left it alone.

## Test plan

- Both workflows parse as YAML, and the expanded matrix was verified job by job
- The three packaging jobs were confirmed to still install a literal `nightly` — they have no matrix, so a templated version there would have expanded to an empty string and broken the install
- The real verification is this PR's own CI run

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,10 +18,16 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          # Warnings are denied on stable only. Nightly adds new warnings ahead
+          # of anything we could have fixed, so denying them there turns the job
+          # red for code that is fine and trains us to ignore it; stable is the
+          # gate that keeps warnings out of the tree.
           - moonbit-channel: nightly
             moonbit-version: nightly
+            deny-warn: ""
           - moonbit-channel: stable
             moonbit-version: latest
+            deny-warn: "--deny-warn"
     steps:
       - name: Checkout code
         uses: actions/checkout@v5
@@ -104,8 +110,9 @@ jobs:
       - name: Check formatting
         run: moon fmt --check
 
-      - name: Check (deny warnings)
-        run: moon check --deny-warn
+      - name: Check
+        # `--deny-warn` on stable only; see the matrix note.
+        run: moon check ${{ matrix.deny-warn }}
 
       - name: Check generated interfaces
         # `moon info` formats abstract types differently across toolchain
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,6 +108,14 @@ jobs:
           nm -D "$dist/lib/libproton.so" | grep -q proton_runtime_set_menu_json
 
       - name: Check formatting
+        # Stable only, for the same reason as `--deny-warn` below. `moon fmt` is
+        # a moving target on nightly: its output for markdown `mbt` blocks
+        # changed between 2026-07-15 and 2026-07-25, which red-lined this step
+        # for a file no one had touched. And because the step ran on both
+        # channels, the repo could only ever satisfy it while the two
+        # formatters agreed — once they disagree there is no committed form
+        # that passes, which is not a state CI should be able to reach.
+        if: matrix.moonbit-channel == 'stable'
         run: moon fmt --check
 
       - name: Check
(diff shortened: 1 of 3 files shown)


bobzhang:
## CI results from the first run, and a second commit

### The `--deny-warn` change is confirmed working

`desktop check` was red on `main` at the `Check (deny warnings)` step, and the cause was exactly the case this PR is about — a nightly deprecation promoted to a build error:

```
Error: [0020]
 ╭─[ agent_tool/multi_edit/multi_edit.mbt:883:29 ]
 883 │   (a.start, a.end, a.index).compare((b.start, b.end, b.index))
     │                             ───┬───
     │                                ╰───── Error Warning (deprecated): Use `compare` instead
Failed with 0 warnings, 2 errors.
```

With the gate moved to stable, `desktop check (nightly)` passes here.

### The stable desktop failures were network, not incompatibility

Both new stable jobs failed, but neither compiled a line of MoonBit:

- `desktop check (stable)` → died in `Seed MoonBit registry index` (the mooncakes.io `git clone`, already retried 3×)
- `desktop test (stable)` → died in `Test (release)`, but inside dependency resolution:
  ```
  Error: Failed to resolve the module dependency graph
    1: error sending request for url (https://download.mooncakes.io/user/moonbitlang%2Fx%2F0.4.46.zip)
    2: operation timed out
  ```

`desktop test (release) (nightly)` ran the identical steps and passed. Worth noting that this PR doubles the desktop jobs and therefore doubles concurrent registry traffic — if these keep flaking, the dependency install wants the same retry wrapper the index-seed step already has.

### Second commit: formatting

`check (nightly)` failed at `moon fmt --check` on `agent_tool/run_moonbit/README.mbt.md`. Diagnosis: nightly's formatter changed between 2026-07-15 and 2026-07-25 and now normalizes markdown `mbt` blocks — inserting `///|` and expanding single-line `import { ... }`. Local `moon fmt --check` and `check (stable)` both pass on the same tree, so this is toolchain drift, not a bad commit.

The structural problem is that `moon fmt --check` ran on **both** channels, which is only satisfiable while the two formatters agree — once they disagree about a file, no committed form passes both and CI is wedged. So the step is now stable-gated, same as `--deny-warn`.

The flagged block is fixed too, since nightly's preferred output is what the next block in that same README already looked like. Verified the new form survives the older local formatter (`moon fmt --check` passes, `moon fmt` does not revert it).

PR #586, opened 2026-07-25
refactor(deepseek): make api_key a Model method, match model with matches

bobzhang:
## What

`internal/agentcli` held exactly one function, `api_key(matches, model)`. This folds it into `deepseek` as `Model::api_key(matches)` and simplifies the resolution by matching the model and the parsed options together.

## Why the match merges

The old body matched twice. The first match pulled **both** provider keys out of `values`:

```moonbit
let (generic_key, deepseek_key, kimi_key) = match matches { ... }
```

The second then picked one and threw the other away, alongside the provider's display name and env var:

```moonbit
let (provider_key, provider, env_var) = match model {
  Kimi(_) => (kimi_key, "Kimi", "KIMI")
  Deepseek(_) => (deepseek_key, "DeepSeek", "DEEPSEEK")
}
```

Matching `(self, matches)` collapses that: each arm reads only the option its provider actually uses and keeps that option name next to the display name and the environment variable. Adding a provider is one new arm rather than an edit in three places, and the discarded extraction is gone.

## Why it lands on `Model`

The model tag is precisely what decides which provider a key belongs to, so the behavior belongs on the type. `internal/agentcli` is empty afterward and is deleted; every call site becomes `model.api_key(matches)`.

`deepseek` picks up a `moonbitlang/core/argparse` import for the `Matches` pattern. That is the one place in the package that reads parsed CLI options — noted in `deepseek/README.mbt.md` — and it stays pure (no I/O). The package still checks clean on js, wasm, and wasm-gc, which `internal/agentcli` (native-only) never had to.

## One behavioral non-change worth flagging

`.trim()` returns a borrowed `StringView`, which is why the original wrapped its returns in `"\{key}"`. That is now an explicit `.to_owned()` with a comment, so nobody deletes it as redundant interpolation.

## Test plan

- `moon fmt`, `moon check` — clean
- `moon test` — 1515/1515 pass (native)
- `moon check --target {js,wasm-gc,wasm} deepseek` — clean
- `moon info` — `.mbti` regenerated

The four `api_key` tests in `cmd/openseek/main.mbt` carry over unchanged in intent; two now assert against the `model` that `agent_settings` already returns rather than re-spelling the constructor.


bobzhang:
Pushed 24df18ad: `Model::api_key` now returns `String?` and the mains report the failure.

## The option is what actually shrank the body

Raising forced each arm to produce a value for the fall-through path, so every option had to be defaulted to `""` and re-tested afterwards. Returning an option lets an unset, empty, or blank option just *not match* — the arms become an ordered list of places a key may come from:

```moonbit
pub fn Model::api_key(self : Model, matches : @argparse.Matches) -> String? {
  // `trim` yields a borrowed view; the caller owns the key, so copy out.
  match (self, matches.values) {
    (_, { "api-key": [k, ..], .. }) if k.trim() != "" =>
      Some(k.trim().to_owned())
    (Deepseek(_), { "deepseek-api-key": [k, ..], .. }) if k.trim() != "" =>
      Some(k.trim().to_owned())
    (Kimi(_), { "kimi-api-key": [k, ..], .. }) if k.trim() != "" =>
      Some(k.trim().to_owned())
    _ => None
  }
}
```

Gone with it: the `Some([]) | None with k = ""` defaulting, the four-tuple, and the post-match if-chain. 30 lines to 9. Matching `matches.values` instead of `matches` is what keeps each pattern and its guard on one line — `moon fmt` will not break before a guard, and the full `Matches` pattern pushed the arms past 80 columns.

Call sites:

```moonbit
guard model.api_key(matches) is Some(api_key) else {
  fail("an API key is required for \{model}: pass --api-key")
}
```

The option still forces all 7 sites to handle the missing case — at the type level now rather than via `raise`.

## Message change to be aware of

The reported text is generic where it used to be provider-specific. It previously said `set DEEPSEEK` / `set KIMI`, which needs the provider tag; keeping that would have meant a second accessor on `Model`, which was declined in review. So the env-var hint is dropped from the first-run error.

Test fallout: the two tests asserting the message contained `DEEPSEEK` now assert `api_key(...) is None` directly. The TUI test still goes through `parse_config`, so it keeps an end-to-end check that a main turns `None` into a failure — that path is now covered where it previously wasn't.

`moon fmt` / `moon check` clean, `moon test` 1515/1515 native, `deepseek` clean on js/wasm/wasm-gc, `.mbti` regenerated.

PR #614, opened 2026-07-30
feat(shell): trust recoverable git worktree verbs in the source-write sandbox

bobzhang:
## What

The shell sandbox was too restrictive for worktree workflows: `git worktree add` fell into the source-write sandbox, where materializing the new checkout is kernel-denied. This PR classifies the recoverable `git worktree` verbs as trusted source writes:

- **`add` (any flags)** — git *unconditionally* refuses an existing non-empty (or file) target; `--force` only relaxes the branch-checked-out-elsewhere and registered-but-missing-path safeguards (verified empirically on git 2.51, including `--force --force`). So `add` only creates new paths whose content comes from the object store. A `-B` branch reset is reflog-recoverable, same as already-trusted `git reset`.
- **`remove` without force** — git refuses a worktree containing modified or untracked files, so everything deleted is tracked and clean. (Ignored files go silently — the same exposure as trusted `checkout`.)
- **`prune`** — touches only `$GIT_DIR/worktrees` admin data.

Everything else fails closed: force `remove` stays sandboxed, `list` is read-only and runs fine sandboxed, `move` relocates arbitrary worktree bytes like the excluded `mv`. The existing gates keep applying: custom environments and reconfiguring global options demote the command, and scratch-lab children remain force-sandboxed.

No preblock changes: unlike custom-env `checkout`, a repointed `worktree add` still cannot overwrite existing source (the non-empty refusal holds regardless of which repo it runs from).

## Tests

- `git_policy` unit tests for the new `worktree_args_are_recoverable` classifier.
- `sandbox_wbtest` classification cases (trusted verbs, force-remove/reconfigured/custom-env demotions, not-preblocked).
- Two e2e tests, written first and confirmed failing pre-change (negative control):
  - add → verify checkout → remove happy path runs unsandboxed and succeeds;
  - forced remove of a dirty worktree runs sandboxed: exit≠0, denial feedback present, and an untracked **source** file survives. The test deliberately pins only source survival — under a partial denial git deletes what it can first, and non-source untracked files are deletable by any sandboxed command anyway; their deletion order is filesystem-dependent (observed flaking on APFS before this was pinned down).
- Full `agent_tool/shell` suite: 123/123 in a fresh worktree (the 5 output-limit failures in the long-lived main checkout are the known pre-existing local flake, also failing there on clean main).

### DIFF
--- a/agent_tool/shell/internal/git_policy/git_policy.mbt
+++ b/agent_tool/shell/internal/git_policy/git_policy.mbt
@@ -75,6 +75,43 @@ pub fn subcommand_writes_from_object_store(subcommand : String) -> Bool {
 }
 
 ///|
+/// Whether a `git worktree` invocation (args in `words[start..end)`, starting at
+/// the verb) is a recoverable write. `add` qualifies with any flags: git
+/// unconditionally refuses an existing non-empty (or file) target — `--force`
+/// only relaxes the branch-checked-out-elsewhere and registered-but-missing-path
+/// safeguards — so it only creates new paths whose content comes from the object
+/// store, and a `-B` branch reset is reflog-recoverable like trusted `reset`.
+/// `remove` qualifies only without force: git then refuses a worktree containing
+/// modified or untracked files, so everything deleted is tracked and clean
+/// (ignored files go silently — the same exposure as trusted `checkout`).
+/// `prune` touches only `$GIT_DIR/worktrees` admin data. Everything else fails
+/// closed: `list` is read-only and runs fine sandboxed, `move` relocates
+/// arbitrary worktree bytes like the excluded `mv`.
+pub fn worktree_args_are_recoverable(
+  words : Array[String],
+  start : Int,
+  end : Int,
+) -> Bool {
+  guard start < end else { return false }
+  match words[start] {
+    "add" | "prune" => true
+    "remove" => {
+      for index in (start + 1)..<end {
+        let arg = words[index]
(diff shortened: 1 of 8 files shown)


bobzhang:
Two follow-up commits:

- **a2e6b69d** fixes a P1 from the subal review: git accepts unambiguous long-option abbreviations, so `git worktree remove --fo` acts as `--force` (verified on git 2.51 — it silently destroys a dirty worktree) and the exact-spelling check would have trusted it. `remove` now fails closed on *any* option before `--`; words after `--` are paths, so a worktree literally named `-f` stays removable non-force.
- **3b9b23e1** documents the workflow in the system prompt: do risky/exploratory changes in branch worktrees under `.worktrees/` inside the workspace, run each worktree command standalone (chains aren't trusted), clean up with non-force `remove` + `prune`. A wbtest pins the exact recommended command shape to the trust classifier.

bobzhang:
Follow-up round 2 (subal on the two previous commits):

- **P1 "restore lepus revision"**: local-environment artifact — the reviewer diffs the working tree, and this checkout has a pre-existing dirty `desktop/lepus` submodule pointer. The PR range `9e8816c6..HEAD` touches zero lepus paths.
- **P2 (real)**: fixed in **0a2c9cc3** — the prompt now has the agent append `.worktrees/` to `.git/info/exclude` once per repository before the first `git worktree add`, so the parent checkout stays clean and `git add .` can never stage the nested checkout as a gitlink. Local-only exclude, so the guidance works in any user workspace.

PR #618, opened 2026-07-31
fix(prompt): make the guard example total — nightly deny-warn breaker

bobzhang:
## Summary
`check (nightly, native)` is red for every PR: nightly moon emits guard_inexhaustive (0087) for the `guard rest is " world"` example in `prompt/base_prompt.mbt.md` (checked fence), and `--deny-warn` fails the job. The example taught guard-without-else crashing — a shape models would now be warned on themselves — so it gains an `else` and the comment says why. base_prompt is the unselected experimental prompt; generated file regenerated.

Repo `moon check` now clean of warnings; `moon fmt --check` clean.

### DIFF
--- a/prompt/base_prompt.mbt.md
+++ b/prompt/base_prompt.mbt.md
@@ -940,7 +940,7 @@ test "string indexing and utf8 encode/decode" {
   guard b0 is ('\n' | 'h' | 'b' | 'a'..='z') && s is [.. "hello", .. rest] else {
     fail("unexpected string content")
   }
-  guard rest is " world"  // otherwise will crash (guard without else)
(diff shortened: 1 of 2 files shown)


bobzhang:
Nightly-native note: this PR removes the guard_inexhaustive deny-warn breaker (nightly now gets past prompt checking), but the job still fails later on a pre-existing nightly-toolchain drift: `undefined reference to moonbit_get_cli_args` from `moonbitlang/x/sys` (linker, debug test link of desktop + cmd). That needs a moonbitlang/x bump or upstream nightly fix — tracked separately; not caused or fixable here. Stable channels green.

PR #627, opened 2026-07-31
build(desktop): unify hashed web assets on esbuild

tonyfettes:
## Summary

- replace the Browser-only Vite build and the separate Desktop minification step with one MoonBit-managed `web_assets` package
- use pinned, SHA-256-verified standalone esbuild 0.28.1 binaries on macOS arm64, Linux x64, and Windows x64
- emit content-hashed JS, CSS, codicon, and Geist assets plus the same deployment manifest for Desktop and Browser
- keep debug Web builds readable and add esbuild `--minify` only when the package command receives `--release`
- keep relative URLs by default and support a Browser CDN prefix through `OPENSEEK_ASSET_BASE`

## Why

Desktop and Browser previously followed different asset paths: Desktop only minified a stable `frontend.js`, while Browser used Vite to hash the complete asset graph. That could let their CSS, fonts, terminal runtime, and HTML wiring drift.

Both personalities now differ only in their MoonBit entry package. They share the same esbuild invocation, fixed SeekMoon title, template rendering, runtime assets, hash naming, and manifest generation. `package/browser` no longer needs npm or Vite.

Package commands preserve the debug/release behavior added on main: without `-- --release`, the nested MoonBit frontend build uses debug mode and esbuild omits `--minify`; with `-- --release`, both MoonBit optimization and esbuild minification are enabled.

## CDN behavior

The default output is a self-contained directory whose HTML and CSS use relative URLs. Setting `OPENSEEK_ASSET_BASE` to an absolute trailing-slash HTTP(S) URL applies that prefix to the HTML JS/CSS references and to CSS font references. Manifest paths remain relative so upload tooling can address the generated files directly.

When openseek-api updates its OpenSeek revision, its production frontend stage should invoke `moon run package/browser --target native -- --release`; omitting the package flag intentionally produces the readable debug distribution.

## Validation

- targeted native package tests: 23 passed
- release-mode `web_assets` tests: 2 passed
- `moon test --target js --deny-warn`: 2512 passed
- debug Browser build: esbuild command omits `--minify`, generated JS about 28.0 MB
- release Browser build: nested MoonBit build receives `--release`, esbuild receives `--minify`, generated JS about 8.6 MB
- latest Mermaid assets are retained in both generated distributions
- `moon fmt`
- `git diff --check`

The first post-rebase Browser run also verified compatibility with the editor current relative `./codicon.ttf` URL.


tonyfettes:
Not planned in near term

tonyfettes:
Will (and should) be addressed in a in-coming refactor where we will replace all these build to justfile with `#[cache(..)]`, when proton upstream provides single proton_package with command line arguments.

PR #639, opened 2026-08-04
refactor(desktop): use `http://` instead of `proton://`

tonyfettes:
TL;DR: `<iframe>` tag won't work if the origin starts with `proton://`. Therefore this PR (#639) and PR #640 tries to use `http://` to serve the frontend assets.

The desktop window leaves `proton://app` and becomes the same JSON-RPC WebSocket client a relayed browser is. Groundwork for the sidebar dock's browser tabs (`docs/plans/sidebar-dock.md`, M1.5); the dock strip hoist itself (M1) follows in a separate stacked PR.

## What changes

- The host serves the bundled frontend over `http://127.0.0.1:<port>` (`internal/remote/local.mbt`; port persisted in `local_http.json` so the origin — and the webview's localStorage — survives relaunches) and upgrades `/ws` into **the same `WsClientActor` a relayed browser gets**. The in-process `__MoonBit__` bridge envelope (`internal/extension/`) is deleted; `proton_ext` / `proton_contract` deps dropped. No lepus bump needed — a `Url` entry needs no bridge.
- `/ws` requires a per-launch entropy token (rides the entry URL) plus an exact `Origin` match; static assets resolve only plain visible names.
- Loopback-only methods stay off the shared catalog and live on the local connection: `shell.open_external` (moved) and the new `notification.show`. Native notification clicks pump through the hub as `notification.click`, delivered to loopback connections only.
- Frontend: `bridge_request` always speaks JSON-RPC WS; `ws_connect` takes the `ChannelId`; every runtime protocol sniff is gone — the shell executable names itself (`boot(desktop~)`), including the external-link capture listener (previously keyed on `location.protocol === "proton:"`, which would have turned desktop externals into CEF popups).
- Contract updated in `docs/remote-protocol.md` (Transport, Authentication, Local-only methods, v2.2 change note).

Why: the M0 probe showed CEF swallows every renderer-initiated http request from a non-http document origin, which killed both iframes and the proton-origin escape hatches. An http window fixes the dock's M2 iframes and collapses the desktop onto the one wire protocol.

## Validation

- `moon check` / `moon test`: js 736/736, native 1911/1911, both warning-free; `moon build --target native` links.
- Pre-probe: WS-from-http-origin under CEF verified end-to-end (handshake with correct `Origin`, both frame directions) before the design was committed.
- Live E2E (packaged bundle) still pending: window paint, run round-trip, terminal, notification permission/click, external links, reload-reconnect, relay coexistence.

## Known cost

The old `proton://app` origin's localStorage is unreachable: UI prefs reset once; only users jumping from pre-host-store builds must re-enter engine keys in Settings.


tonyfettes:
Currently, this PR has a huge deficit: preferences/settings lost every application reboot. Reason:

- Different port belongs to different "origin", i.e. http://localhost:8081 and http://localhost:8082 represent two different website.
- Because we are using random ports for our application, it can use a new port when user restarts application, and therefore their preferences/settings are lost.

tonyfettes:
See #657

PR #653, opened 2026-08-05
feat(shell): trust recoverable git rebase forms

bobzhang:
## Summary

Extends the shell tool's trusted-source-write git policy to `rebase`, classified **by form, not by name** — the same model used for `checkout` (safe forms trusted, dangerous forms pre-blocked).

- **Trusted (unsandboxed)**: an exact-token allowlist (`rebase_args_are_recoverable`) admits plain rebases, `--onto <ref>`/`--onto=<ref>`/`--keep-base`, `--fork-point`/`--no-fork-point`, `--autostash`/`--no-autostash`, positional revisions, and the in-progress verbs `--continue`/`--abort`/`--skip`/`--quit`. These satisfy the existing trust invariant: every worktree write materializes from the object store, the pre-rebase tip stays reachable via `ORIG_HEAD`/the reflog, git refuses to start on a dirty tree and refuses to overwrite untracked files during replay.
- **Fail closed to the sandbox**: anything not exactly known-safe (`--rebase-merges`, `--root`, abbreviations like `--cont`, …). Allowlist polarity is deliberate — opposite of the checkout guard — because rebase's option surface is large, includes command runners, and git accepts unambiguous long-option abbreviations.
- **Pre-blocked cross-platform**: the command-running forms — `-i`/`--interactive`, `-x`/`--exec`, `--edit-todo`, `-s`/`--strategy`/`-X` — including abbreviation prefixes (`--exe=cmd`), short clusters (`-iv`), and attached values (`-xcmd`). On a platform without `sandbox-exec` these previously ran unguarded (`git rebase --exec 'anything'` on Linux); now they are blocked before execution.
- **Custom-environment rebase** joins the object-store-writer pre-block on both the argv and TooComplex text paths (`subcommand_is_object_store_writer`): a repointed `GIT_DIR`/`GIT_CONFIG_*` could write workspace source from a different repository.

Conflict resolution still flows through `edit` + sandboxed `git add`, so trusting rebase opens no source-editing channel outside the tools — the object store only contains history plus edit-tool-authored content (store feeders remain pre-blocked).

## Testing

- `moon test -p agent_tool/shell/internal/git_policy`: 24/24
- `moon test -p agent_tool/shell`: 228/228
- `moon check --deny-warn` clean, `moon fmt` applied, `moon info` regenerated

### DIFF
--- a/agent_tool/shell/internal/git_policy/git_policy.mbt
+++ b/agent_tool/shell/internal/git_policy/git_policy.mbt
@@ -66,12 +66,86 @@ pub fn global_option_reconfigures(arg : String) -> Bool {
 /// a tracked file recoverable from it — reviewable through git, never sourced from
 /// an external file or stdin. Excluded: `mv` (moves arbitrary worktree bytes onto
 /// the destination) and `clean` (deletes UNTRACKED files with no object-store
-/// copy — permanent). `rm` stays: it only deletes tracked files.
+/// copy — permanent). `rm` stays: it only deletes tracked files. `rebase` also
+/// writes from the object store but is trusted only in the argument forms
+/// admitted by `rebase_args_are_recoverable`, so it is not listed here.
 pub fn subcommand_writes_from_object_store(subcommand : String) -> Bool {
   match subcommand {
     "checkout" | "switch" | "restore" | "reset" | "stash" | "rm" => true
     _ => false
   }
+}
+
+///|
+/// Whether `subcommand` writes the worktree from the object store: the
+/// unconditionally recoverable writers plus `rebase`, whose recoverable forms
+/// are gated by `rebase_args_are_recoverable`. Used by the custom-environment
+/// pre-blocks — any of these, repointed via `GIT_DIR`/`GIT_CONFIG_*`/`env`,
+/// could write workspace source from a different repository regardless of its
+/// arguments.
+pub fn subcommand_is_object_store_writer(subcommand : String) -> Bool {
+  subcommand == "rebase" || subcommand_writes_from_object_store(subcommand)
+}
+
+///|
+/// Whether a `git rebase` invocation's arguments (in `words[start..end)`) keep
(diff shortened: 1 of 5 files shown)


bobzhang:
**Cross-model review** (subal is out of credits until Aug 8, so this round ran on Kimi at xhigh-equivalent): 1 Medium, 2 Low, 1 Info — all addressed or explicitly deferred in `350de8ac6`.

1. **[Medium] Trusted rebase executes repo-configured programs** (`.git/hooks` pre-rebase/post-rewrite/commit hooks, merge drivers on conflicted replay, and — new, since no other trusted subcommand creates commits — `gpg.program` via `commit.gpgSign`), while `.git` itself is writable under the sandbox profile. *Disposition:* the argv layer can't close a config-driven channel; docs now state the exec surface honestly instead of claiming the trust extension is exec-free. The real fix is denying `.git/config` + `.git/hooks` writes in the SBPL profile — deferred as a separate decision because it would also break legitimate sandboxed `git config` use (same pre-existing exposure applies to trusted `checkout` via smudge filters/hooks).
2. **[Low] `-S`/`--gpg-sign` missing from the cross-platform pre-block.** *Fixed:* clusters, attached key ids, and abbreviations (`--gpg-s`) now pre-block; `--no-gpg-sign` deliberately stays unblocked.
3. **[Low] allowlist ∩ pre-block = ∅ untested.** *Fixed:* every admitted form is now asserted non-pre-blocked in one paired loop.
4. **[Info] `--onto -i` divergence** (allowlist mirrors git's consume-next; pre-block scans the raw token). *Fixed:* pinned by test, with the launch-ordering guarantee (`run_agent_shell_with_tree_precheck` refuses before trust is consulted) documented at the `--onto` arm.

Reviewer verified sound: fail-closed on unknown tokens/abbreviations/clusters, `--`/`-`/`--onto=` handling, argv–text–custom-env path consistency, and the `.mbti` update.

PR #664, opened 2026-08-06
refactor(desktop): split frontend update files

tonyfettes:
## Summary

- split the 12,316-line `desktop/frontend/update.mbt` into cohesive `update_xxx.mbt` files
- keep root update orchestration in a 97-line `update.mbt`, with page and device dispatch separated
- group the existing update helpers, fixtures, and tests by responsibility
- preserve all 295 top-level blocks and 244 tests without changing message semantics or the public package interface

## Why

The frontend update file mixed page dispatch, device events, transcript reconciliation, run delivery, settings, release handling, and more than eight thousand lines of tests. The new file layout keeps each file below 2,000 lines and makes those responsibilities navigable while remaining a behavior-preserving, same-package refactor.

## Impact

No user-facing behavior or public API changes are intended. This PR only reorganizes existing MoonBit top-level blocks and adds short file responsibility comments.

## Validation

- `moon check desktop/frontend --target js --deny-warn`
- `moon test desktop/frontend --target js --deny-warn` — 305/305
- `just check`
- `just test` — native 3176/3176, JS 2512/2512, cram 27/27
- `moon fmt --check`
- `git diff --check`


tonyfettes:
Should directly refactored into rabbita components

PR #677, opened 2026-08-06
fix(desktop): recover a conversation whose worktree vanished

tonyfettes:
Follow-up to #661, from its review.

## The defect

Delete a conversation's checkout outside the app (`rm -rf`, or a `git worktree remove` from a terminal) and the registry row stays bound while `session_tree_dir` quietly answers with the workspace root. The next turn then runs the agent **in the main checkout** — the opposite of what a worktree conversation asked for — and the sidebar goes on claiming the worktree.

That fallback only ever served this case. Our own removal paths drop the registry row outright, so a row that outlives its checkout means the directory vanished outside OpenSeek: the one situation the user never agreed to.

## The fix

The registry answers with a three-state `SessionTree` (`WorkspaceRoot` / `BoundTree` / `MissingTree`), so "unbound" and "bound but gone" stop sharing a branch:

- **File browsing** still degrades to the workspace root — read-only, and a broken tree beats an empty panel.
- **The run path** refuses, naming the worktree, instead of silently relocating the agent.

The composer turns that refusal into the two ways out, rather than a dead end:

- **Rebuild the checkout** on its surviving branch. `worktree.create` already binds by session, so it repairs the existing entry instead of minting a second one, and the branch keeps every commit.
- **Release the conversation** to the workspace. `worktree.remove` already prunes and unbinds when the checkout is missing, so this needs no new host op.

Both land as the usual `worktree.changed` broadcast; neither mutates optimistically.

This also gives `WorktreeInfo.present` its first reader — the client stored the field and never looked at it.

## Verification

`moon check` clean on both targets; `moon test` 3187/3187 native, 2538/2538 js. New host tests cover the three-state resolution and the repair path; new frontend tests cover the banner and both recovery commands.

The UI itself is unverified — the disabled composer, its notice, and the two buttons want a human at the app.

### DIFF
--- a/desktop/app.css
+++ b/desktop/app.css
@@ -2671,6 +2671,25 @@ main {
   display: inline-flex;
   align-items: center;
 }
+/* The notice's own text/action split, used by the vanished-worktree banner:
+   the sentence takes the slack so the actions stay right-aligned. Metrics
+   live here rather than on `.secondary` — see the generic-buttons note. */
+.composer-notice-text {
+  flex: 1;
+  min-width: 0;
+}
+.composer-notice-btn {
+  flex: none;
+  border-color: var(--line);
+  background: var(--surface);
+  color: var(--ink);
+  padding: 4px 10px;
+  font-size: 12px;
+  white-space: nowrap;
+}
+.composer-notice-btn:not(:disabled):hover {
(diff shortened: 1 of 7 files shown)


tonyfettes:
superseded by #683

PR #678, opened 2026-08-06
fix(desktop): key terminal tabs by the execution tree

tonyfettes:
Follow-up to #661, from its review.

## The defect

The host roots a terminal in the conversation's worktree — `terminal.open` resolves its cwd through the same registry seam an agent run uses — but tabs were grouped by `(channel, workspace)` alone.

Two conversations in one project therefore shared a tab group even when they sat in different checkouts. Switching from a worktree conversation to the Local one handed the user the **worktree's** shell, and its commands ran in a checkout the user was not looking at.

Manually typed commands are the hazard here: an agent run leaves a transcript, `rm` in the wrong checkout does not.

## The fix

`WorkspaceKey::Project` now carries the worktree alongside the workspace, so a group is the **execution tree** rather than the project. Conversations that genuinely share a checkout still share their shells, which is what grouping is for.

- Both constructors take labelled fields — three positional arguments of two types was already one too many to read.
- The tab label carries the worktree name (`project/wt-1`) so the strip tells the checkouts apart.
- `Ctx.worktree` resolves the way the sidebar chip already does: the local binding while the create-then-start chain owns it, the registry join afterwards — which is also what survives a page reload.

## Verification

`moon check` clean on both targets; `moon test` 3185/3185 native, 2538/2538 js.

The new blackbox test drives the public surface rather than reaching into `visible_tabs`: opening a terminal in the worktree conversation and then switching to Local must make `sync` open a *second* shell. Under the old key it reused the first one and the count stayed at 1. It also asserts a same-checkout sibling still shares, and that returning to the worktree finds its own.

### DIFF
--- a/desktop/frontend/terminal/pkg.generated.mbti
+++ b/desktop/frontend/terminal/pkg.generated.mbti
@@ -24,6 +24,7 @@ pub(all) struct Ctx {
   channel : @interop.ChannelId
   session : String
   workspace : @pathx.Absolute?
+  worktree : String?
   dark_mode : Bool
   connected : Bool
 }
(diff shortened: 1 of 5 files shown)


tonyfettes:
Fixed in #683

PR #680, opened 2026-08-06
feat(desktop): show tool call details

tonyfettes:
## Summary

- project assistant tool calls into the Desktop transcript in model-emitted order
- pair later tool results into the existing call row by `tool_call_id`
- show recorded arguments before result output while preserving legacy unmatched results
- recognize every current `cmd/openseek` built-in in activity summaries

## Why

Desktop previously projected only durable `ToolResult` events. That hid the original arguments and let result arrival order determine the visible order. Recording calls from the assistant event and updating them in place preserves the actual call timeline while keeping result state current.

## Validation

- `moon -C desktop test frontend --target js --deny-warn` — 329 passed
- `just check`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/desktop/app.css
+++ b/desktop/app.css
@@ -2392,7 +2392,8 @@ main {
   padding: 6px 0 2px 16px;
 }
 
-/* One call inside a burst: its brief, expandable to the output. */
+/* One call inside a burst: its brief, expandable to recorded arguments and
+   the result output. */
 .tool-call-label,
 .tool-call-summary {
   color: var(--ink-3);
@@ -2406,9 +2407,20 @@ main {
   color: var(--del);
 }
 
-/* Output grows the page instead of scrolling in place: the transcript
-   stays the only scroll area, so the wheel is never trapped by a
-   nested scrollbar. Long dumps are clamped in the renderer. */
+/* Arguments and output grow the page instead of scrolling in place: the
+   transcript stays the only scroll area, so the wheel is never trapped by a
+   nested scrollbar. Long result dumps are clamped in the renderer. */
+.tool-call-section + .tool-call-section {
+  margin-top: 8px;
+}
+.tool-call-section-label {
+  margin-top: 6px;
+  color: var(--ink-3);
+  font-size: 10px;
+  font-weight: 600;
+  letter-spacing: 0.08em;
+  text-transform: uppercase;
+}
 .tool-call pre {
   margin: 6px 0 8px;
   padding: 10px 12px;
@@ -2420,7 +2432,7 @@ main {
   white-space: pre-wrap;
   overflow-wrap: anywhere;
 }
-.tool-call.error pre {
+.tool-call.error .tool-call-output {
   background: var(--del-bg);
 }
 
--- a/desktop/frontend/model.mbt
+++ b/desktop/frontend/model.mbt
@@ -3358,7 +3358,8 @@ fn Conversation::freeze_durable_tail_notices_from_snapshot(
 ///|
 /// Apply one durable commit's item to the transcript. Local notices have no
 /// durable twin and are therefore never retired by item class. Returns
-/// whether the durable transcript visibly grew.
+/// whether the durable transcript visibly changed: a tool result can now fill
+/// an existing call row without increasing the array length.
 fn Conversation::apply_commit(
   self : Conversation,
   sequence : Int,
@@ -3367,8 +3368,7 @@ fn Conversation::apply_commit(
   // A canonical User is transcript data, never local-submission ownership.
   // It is not a run boundary: the same kind also stores in-run steering.
   let messages = self.messages.copy()
-  let before = messages.length()
-  @transcript.apply_session_item(messages, item)
+  let changed = @transcript.apply_session_item(messages, item)
   // Unknown terminal payloads render nothing but still close the durable run.
   let terminal = item.is_terminal()
   let conv = {
@@ -3398,7 +3398,7 @@ fn Conversation::apply_commit(
     sequence,
     messages.length(),
   )
-  (messages.length() > before, conv)
+  (changed, conv)
 }
 
 ///|
--- a/desktop/frontend/transcript/pkg.generated.mbti
+++ b/desktop/frontend/transcript/pkg.generated.mbti
@@ -8,7 +8,7 @@ import {
 }
 
 // Values
-pub fn apply_session_item(Array[TranscriptItem], @protocol.SessionItem) -> Unit
+pub fn apply_session_item(Array[TranscriptItem], @protocol.SessionItem) -> Bool
 
 pub fn apps_from_reply(@protocol.ListAppsReply) -> Array[LauncherAppItem]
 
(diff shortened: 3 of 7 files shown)


bobzhang (approved):
Rebased onto `main` and reviewed. **LGTM — approving.**

## Rebase

Force-pushed `dd13f883` → `c1ff2438`, all four commits preserved with original authorship.

Only one conflict, in `desktop/frontend/transcript/pkg.generated.mbti`, entirely against #682's typed-path refactor (`@common.Uri` workspace, `sessions_from_reply` taking `@interop.ChannelId`) — not against this PR's logic. Resolved by keeping `main`'s signatures and this PR's `apply_session_item -> Bool`, then confirmed the resolution is exactly what the generator produces: `moon info` leaves the file byte-identical.

`view.mbt` / `update.mbt` / `model.mbt` auto-merged despite `main` having moved ~590 lines across them. Sanity check that nothing was silently dropped: the per-file line counts of the net diff are identical to the pre-rebase PR for every file except the `.mbti` (`app.css` 22, `model.mbt` 8, `sessions.mbt` 220, `types.mbt` 18, `update.mbt` 32, `view.mbt` 246).

Verified locally on the rebased tip:

- `moon check --target native --deny-warn`, `moon check --target js --deny-warn`, `moon fmt --check` — clean
- `moon test --target js` — 2596 passed
- `moon -C desktop test frontend --target js --deny-warn` — 352 passed
- `moon info` — no `.mbti` drift
- `git diff --check origin/main...HEAD` — clean

## Review

The core design decision is right: recording calls from the assistant event and filling them by `tool_call_id` makes the visible order the *model-emitted* order rather than the result-arrival order, and it does so without a second index — the id lives on the row.

What I specifically checked:

**Pairing is conservative in exactly the right places.** Every path that isn't an unambiguous match — empty `tool_call_id`, no recorded call, a call that already has a result, two pending calls sharing an id — falls through to the append-only row, which is byte-for-byte the pre-PR rendering. So the worst case for a malformed or legacy log is the old behavior, never a wrong pairing or a swallowed result. The `has_result` flag rather than an "is the content empty" heuristic is the right call: a legitimate result with empty content, `is_error=false` and no brief is indistinguishable from a pending call otherwise.

**The ordering assumption holds, in the code and against real data.** The pairing depends on the assistant event committing before its results, and it does: `record_assistant` runs before the batch's `for call_index, native_call in response.tool_calls` loop on every path, and sequences are assigned serially under the store's exclusive lock. Worth noting the *live* wire stream is explicitly not ordered — `prefetch_concurrent_results` emits a prefetched group's brackets while earlier batch calls are still recording — but Desktop only projects durable commits, so that doesn't reach here. I also audited the five recorded session logs in the repo (287 tool calls): assistant-then-results in every case, 1:1 pairing throughout, zero empty `tool_call_id`s, zero duplicate ids, zero results before their call, zero orphans on either side.

**No orphan results.** Only two durable `tool_result` writers exist (`agent/turn_loop.mbt`), both fed from the same `response.tool_calls` recorded in the assistant item; the engine never synthesizes an id. Sub-agents write their own sibling session file, compaction only appends a `Summary`, and `cmd/openseek` appends only `User`/`Runtime`/`Summary` directly. So the unmatched-append branch really is a legacy/malformed path, not a live one.

**No double-rendering with the live path.** `EngineEvent::ToolResult` and `ToolCallDecodeError` both stay `(false, conv)` in `update.mbt`, so durable commits remain the sole source of tool rows. Nothing to reconcile.

**`apply_commit`'s return type change is load-bearing, not cosmetic.** An in-place fill grows the row's height without growing `messages.length()`, so the old `messages.length() > before` would have skipped `scroll_if_active` on exactly the event the reader is waiting for. Threading `changed` out of `apply_session_item` fixes that. Overlay anchoring is unaffected — `event_boundaries` still records `items.length()` per sequence, and the tests pin the boundaries staying flat across a fill.

**Argument bounding.** Clamping before `@json.parse` is the right order — `write`/`multi_edit` arguments carry whole files, and a collapsed `<details>` body is still constructed on every render. The parse falling back to the bounded raw text when truncation breaks the JSON is honest and cheap. `clamped_view` keeping cuts out of surrogate pairs is a nice catch given the `capitalize` bug the file already documents.

## Non-blocking notes

1. **`has_result` never reaches the view.** `tool_call_view` destructures `name~, arguments~, is_error~, brief~, ..` and drops it, so a call with no result renders identically to a finished one with empty output — same label, same expandable arguments, no "waiting" affordance.

   This matters more than it looks, because permanently-unanswered calls are reachable today. `Control(Abort(reason))` at `agent/turn_loop.mbt:735` appends its own result and the `Terminal(Aborted)`, but unlike the finish arm at `:666` it never calls `close_skipped_tool_calls` on `response.tool_calls[call_index + 1:]` — so aborting mid-batch leaves the remainder unanswered in the durable log forever. Interrupt/failure and a mid-batch crash do the same; `flush_pending_tool_calls` repairs those for model context at read time only, never durably. Before this PR those calls were simply invisible; now each one is a row that looks like a completed call that returned nothing. Surfacing them is the more honest rendering — it just wants a marker to say which it is.

2. **Consider preferring the most recent pending match over refusing to pair.** `pending_matches == 1` is the right *default*, but it has a reachable failure mode in combination with the above: a stale pending call from an aborted turn, plus an id collision in a later turn, makes both rows unfillable and pushes the new result out as a duplicate unmatched row. DeepSeek's `call_NN_<random>` ids won't collide, but `api_url` is user-configurable, and per-turn `call_1`/`call_2` ids are exactly what OpenAI-compatible servers commonly emit (and what this repo's own mock servers reuse across turns). Scoping the scan to rows added since the last assistant event, or taking the last pending match, would close it without giving up the conservative fallback.

   Related and already handled correctly: a provider that streams tool-call deltas without an `id` leaves `StreamToolCall.id` as `""` (`deepseek/client/stream.mbt`), which reaches the durable log as `tool_call_id: ""` via the decode-error path. The `!payload.tool_call_id.is_empty()` guard catches that and falls back to the old rendering — good.

3. **Per-render cost of the arguments path.** `truncate_output` → `@json.parse` → `stringify(indent=2)` → `truncate_output` runs for every tool row on every render, including collapsed ones, and the view isn't memoized. The 20k code-unit bound caps each row but not the row count, so a long session re-rendering per `AssistantDelta` pays ~3× the previous per-row formatting cost. Caching the formatted arguments on the item, or deferring the parse until `toggle`, would take it to zero.

4. Trivial: `trimmed != "{}"` only filters the exact compact spelling — `{ }` or a pretty-printed empty object still gets an "Arguments" section.

## CI

Green except `check (stable, native)`, which is **already red on main** — `adf9c607` (the commit this is rebased onto) fails it in [run 31209115807](https://github.com/moonbitlang/openseek/actions/runs/31209115807), as did #695 before it, with the identical signature: a deadlock in `desktop/internal/terminal/terminal_test.mbt:228` *"duplicate acknowledgements release an output chunk only once"* → SIGABRT. Unrelated package, native target; this PR touches only `desktop/frontend/**` (js) and `app.css`. Not caused by this change, and worth its own issue.

Everything else passes, including `check (nightly, native)` — the job that runs `moon info` and rejects `.mbti` drift, which is the CI-side confirmation of the conflict resolution above.

Nothing here blocks.

PR #683, opened 2026-08-07
fix(desktop): keep missing worktree chats isolated

tonyfettes:
## Summary

- keep the durable conversation-to-worktree binding when a checkout is missing or a task is archived
- remove the Release action; missing checkouts only offer Repair or Archive
- make unarchive restore the task first and leave it in the same Missing UI until the user explicitly repairs the checkout
- reject agent, filesystem, terminal, and file-editor operations while the checkout is missing instead of falling back to the repository's main worktree
- retire stale PTYs/watchers/viewer state and park the dock strip until Repair restores the checkout
- refresh the worktree registry after a rejected start so externally removed checkouts are surfaced immediately

## Why

An absent worktree record was previously indistinguishable from an unbound task, so execution could continue in the registered project's main worktree. Archive also removed the binding, which made unarchive take that same fallback path. This preserves the user's original worktree intent and requires an explicit Repair before work resumes.

## Validation

- `just check`
- `just build`
- `moon test --target js desktop/frontend` (345/345)
- `just test` (native 3182/3182, JS 2577/2577, cram 27/27)
- `git diff --check origin/main...HEAD`

### DIFF
--- a/desktop/app.css
+++ b/desktop/app.css
@@ -2626,6 +2626,25 @@ main {
   display: inline-flex;
   align-items: center;
 }
+/* The notice's own text/action split, used by the vanished-worktree banner:
+   the sentence takes the slack so the actions stay right-aligned. Metrics
+   live here rather than on `.secondary` — see the generic-buttons note. */
+.composer-notice-text {
+  flex: 1;
+  min-width: 0;
+}
+.composer-notice-btn {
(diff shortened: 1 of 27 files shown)


tonyfettes:
@codex review

tonyfettes:
@codex review

PR #706, opened 2026-08-08
fix(desktop): prevent terminal shutdown deadlocks

tonyfettes:
## Summary

- upgrade `moonbit-community/pty` to 0.3.3 so Linux PTY descriptors are created with `CLOEXEC` atomically
- close POSIX terminal sessions through the owning `Pty` before cancelling the output task
- add a regression test for shutdown when a descendant keeps the PTY slave open

## Root cause

PTY 0.3.2 marked descriptors `FD_CLOEXEC` after creating or duplicating them. A concurrent process spawn could inherit an unrelated PTY slave during that window, preventing the terminal reader from observing EOF and leaving the pump's task group blocked.

Separately, OpenSeek's explicit POSIX close path cancelled the saved child PID but still depended on every slave descriptor reaching EOF. Routing shutdown through `Pty::close` lets the PTY implementation terminate the process group and coordinate an in-flight read before the output task is cancelled.

## User impact

Terminal sessions no longer remain stuck during concurrent process spawning or when closing a shell whose descendants retain the slave side of the PTY.

## Validation

- `moon check internal/terminal --target native --deny-warn`
- `moon test internal/terminal --target native --deny-warn` (13/13)
- terminal native suite repeated 20 times (20/20)
- `moon test --target native` (3314/3314)
- `moon info` (no meaningful interface diff)
- `moon fmt --check`
- `git diff --check`

The Linux-only atomic `CLOEXEC` branch requires the Ubuntu CI job for platform-specific verification; local validation ran on macOS.

### DIFF
--- a/desktop/internal/terminal/moon.pkg
+++ b/desktop/internal/terminal/moon.pkg
@@ -3,7 +3,6 @@ import {
   "moonbitlang/core/encoding/utf8",
   "moonbitlang/async",
   "moonbitlang/async/cond_var",
-  "moonbitlang/async/process",
   "moonbitlang/core/debug",
   "openseek_desktop/internal/env",
   "tonyfettes/xlog",
--- a/desktop/internal/terminal/pump.mbt
+++ b/desktop/internal/terminal/pump.mbt
@@ -27,28 +27,18 @@ pub async fn run_terminal_pump(
 }
 
 ///|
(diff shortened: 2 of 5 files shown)


tonyfettes:
Should be fixed by upgrading pty to 0.4.0

PR #712, opened 2026-08-08
feat(desktop): standing goals via /goal

tonyfettes:
Gives a desktop conversation a standing goal: `/goal` puts the composer into goal mode, and the goal the engine persists sits above the input until it is cleared. Brings the desktop to parity with the TUI's `/goal`.

## Where goal state comes from

Entirely from the durable `[goal]` / `[goal cleared]` / `[goal blocked]` / `[goal unblocked]` runtime-notice markers (`agent_session/goal.mbt` is the authority; `frontend/transcript/goal.mbt` re-spells the grammar because that package is js-only and cannot depend on it).

Those markers reach the client through both `session.load` snapshots and `session.event` commits, and **one projector serves both** — so a conversation resumed after a restart shows the goal the engine actually holds, for free, with no second code path to keep in sync. Nothing local writes the mirror.

The engine's `GoalUpdated` / `GoalBlocked` / … **stream** events stay dropped: they duplicate the durable path this client already renders.

## Host

`agent.goal` joins the typed command catalog as `@commands.agent_goal`, handled by `goal_run`. It mirrors `compact_run` — addressed by session rather than run, spawning a serve engine on demand for a conversation with no live process — but **deliberately drops its phase gate**: serve accepts a goal in every phase (appended when idle, steered mid-turn, queued behind a compaction), so there is nothing to refuse. An absent `text` clears.

## UI

One row above the input, in the compaction notice's slot, carrying whichever of these applies: the edit in progress, a set awaiting its marker, or the standing goal.

The row is **always in the DOM**, collapsed by CSS when it has none of them. This is load-bearing, not tidiness: a row that came and went would shift `composer-inner`'s child index, and rabbita's index-based diffing would then rebuild the textarea and drop its focus at exactly the moments this row changes — entering goal mode, leaving it, and the durable `[goal]` commit that lands mid-edit.

## What review changed

Codex raised eight findings on this PR; all held up, and two were caused by earlier fixes in this same branch. Three are worth reading as design decisions rather than patches:

**The goal draft belongs to the conversation, not the app.** It started as one model-wide slot. Keeping a hidden edit alive across a conversation switch — which the clear fix required, and retention then depended on — meant opening goal mode anywhere else silently destroyed it. Patching the singleton was the wrong shape: the draft now lives on the conversation that owns it, so "whose edit is this" stops being a question anyone can get wrong, retention falls out of `draft_empty`, and every handler acts on the conversation on screen.

**A goal establishes placement.** On a fresh conversation with Worktree mode selected, sending `agent.goal` straight through spawned the engine in the workspace root, and the goal's own durable card then made `wants_fresh_worktree` false — so the first real prompt ran in the main checkout against the user's explicit choice. A goal is the first thing that pins an unstarted conversation's placement, so it now creates and binds the worktree first, exactly as the first prompt does.

**`agent.goal` returning is not persistence.** `goal_run` replies once the command reaches the engine's stdin; the append behind it can still fail, or the process can die with the command queued. The submitted text is held as `goal_pending` until **the set marker carrying that same text** commits — not just any goal marker, since a block on the goal that still stands, or an earlier set from a second submit, would otherwise retire the recovery copy while its own append could still fail.

## Deliberately not included

**Auto-continue.** `GoalSet` carries an `auto` flag that arms the engine's autonomous continuation, and the wire and payload keep the field — but the desktop always sends manual, and the toggle that was here has been taken back out.

The reason is a lifecycle gap review surfaced: the desktop's run model assumes every turn is one it started. Serve announces the turns autonomous continuation begins with `GoalContinue`, which this client drops and the host does not act on, so its `ServeEngine.phase` stays `Idle` through an autonomous turn. The composer would then offer Send rather than Stop, `start_run`'s only gate is that phase, and a prompt submitted there is accepted with a fresh run id while serve queues it behind the turn already running — whose terminal event settles it, reporting the user's prompt finished before it starts. No client had ever sent `auto=true`, so nothing had tested that assumption.

Arming returns in a follow-up that teaches the host `GoalContinue` opens a turn it did not start. That is the same missing piece that keeps `GoalContinue(remaining)` / `GoalBudgetExhausted` from showing an armed goal's remaining budget, so both land together.

`GoalCheck` / `GoalReminder` stay dropped on purpose: they are model-directed text, not UI.

## Tests

`moon test` green on both targets: **js 2647/2647**, **native 3340/3340**. Coverage for the marker grammar and its reducer, snapshot/commit projection into cards plus the mirror, the slash command, blank-submit and clear paths, cancel restoring the prompt draft, per-conversation ownership of an edit across a switch, retention of a conversation that holds only a goal draft, which marker settles a pending set, and the worktree routing decision.

Verified to the bundle boundary; the app itself has not been driven from this session.


tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

PR #718, opened 2026-08-08
fix(desktop): serialize shared state writes across processes

tonyfettes:
The desktop has never been the only writer of its durable state:

- a TUI launched in a workspace shares that workspace's session store **by design** (`workspaces.mbt` says so in its header — the two are meant to resume each other's conversations),
- a dev build shares `~/.openseek` with the installed app,
- a leftover process from a previous run outlives the launch that spawned it.

Every registry mutation nevertheless serialized only on an `@async.Mutex`, which orders this process's own tasks and nothing else. Two writers could each read the same list and let the later write silently erase the earlier one. The non-atomic `write_text` also let a concurrent reader observe a truncated file and read it as "nothing is attached".

## What this adds

Two primitives in `internal/fsx/lock.mbt`:

- **`FileLock::acquire(path, kind)` / `release()`** — an advisory `flock` on a sidecar `<file>.lock`, honored across processes. A sidecar rather than the data file itself: an atomic save replaces the target by rename, so a lock taken on the data file would be attached to the inode the rename orphans, and two writers would each hold "the lock" on a different file.
- **`write_text_atomic(path, content, permission?)`** — temp file + rename.

Shaped as an acquire/release pair rather than a scoped callback because every call site already sits next to `mutex.acquire(); defer mutex.release()` — this inserts two lines instead of reindenting the surrounding block into a closure.

## The convention

| | |
|---|---|
| **mutation** | take the exclusive sidecar lock, **re-read inside it**, then write atomically |
| **read** | no lock — the atomic write guarantees a complete version |

Leaving reads unlocked is deliberate. They are everywhere (`registered_workspaces()` alone is called from a dozen places) and they are tolerant by contract: an unreadable registry means "nothing attached", not an error. Locking them would convert that into a raising path. It also would not help the readers that matter most — an external tool or an older build never takes our lock, and what actually protects those is the atomic write.

## Applied to

- `workspaces.json`
- `worktrees.json` (create and remove; the create holds the lock across the `git worktree add` too, so another instance cannot pick the same `wt-N` after this one probed it as free)
- `engine-settings.json`
- the legacy `archived_sessions.txt` migration (probe before locking, so the common "already migrated" launch leaves no sidecar behind)

`write_settings_file` now delegates to the shared helper instead of keeping a second copy of the Windows rename fallback.

## Testing

- `moon check --target native`: clean
- `moon test --target native`: **3343/3343**
- 4 new tests in `internal/fsx`: atomic save leaves no `.tmp`, an exclusive lock makes the next holder wait, shared locks admit each other, a lock can be retaken after release

`@fs.File::lock` is `flock`, which binds the open file description rather than the process, so two acquires of one path contend inside a single process through exactly the mechanism a second instance would hit — the tests cover the cross-process path without spawning a helper.

## Note on a pre-existing flake

The first full-suite run failed one test, `package/internal/packaging` → "Proton CEF root follows the active runtime manifest", with `mkdir target: File exists`. A second full run passed 3343/3343. That package's own `ensure_dir` is check-then-mkdir, and two packaging tests race to create `desktop/target` when it does not yet exist. Pre-existing, untouched by this change, and left alone as out of scope — but it is the same bug class, if someone wants it.

### DIFF
--- a/desktop/internal/engine/settings.mbt
+++ b/desktop/internal/engine/settings.mbt
@@ -204,6 +204,12 @@ pub async fn update_settings(
   guard manager.runtime_dir is Some(dir) else {
     raise EngineError("engine settings cannot persist: no runtime directory")
   }
+  // `settings_lock` orders this process's clients; the runtime directory is
+  // shared with every other instance running against it. The load below has
+  // to follow this lock — settings read before it would carry back the very
+  // fields another instance just changed.
+  let shared = @fsx.FileLock::acquire(settings_file(dir), Exclusive)
+  defer shared.release()
   let current = load_engine_settings(manager.runtime_dir)
   guard current.version <= EngineSettingsVersion else {
     raise EngineError(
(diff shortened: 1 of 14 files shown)


tonyfettes:
@codex review

PR #722, opened 2026-08-08
Fold subagent transcripts under their parent in the sidebar

tonyfettes:
A sub-run (`explore` / `review` / `subtask`) persists its own transcript as a sibling session named `<parent>-sr-N`. The sidebar listed every one of them as a top-level conversation: in this repository's own store 16 of 29 rows are child records, so a single worker sweep buries the chats it belongs to.

Each conversation row now leads with a twisty, and the runs it delegated unfold in place beneath it — a tree, not a second kind of row. Every row carries the cell whether or not it has anything to unfold, so titles stay aligned; the button stops its click short of the row, because unfolding a conversation is not opening it. Which rows are open is window-lifetime view state keyed by `(device, session id)` — session ids only mean anything inside their channel.

`@protocol.split_child_session_id` puts the id derivation next to the `subrun_started` / `subrun_finished` events that name the same runs. The protocol module is shared by the engine and the desktop, so the sidebar and the session viewer stop reinventing the parse (the viewer's own `brief`-string parse is left alone here; it moves onto this in a later change).

The grouping is a pure `@transcript.session_tree`, tested away from the view: parents keep the listing's order, children sort by sub-run ordinal — past `sr-9` an id sort gets it wrong.

**Nothing is hidden.** A child whose parent is not in the group — archived, or a differently-scoped store — keeps its own row where it was listed, and a grandchild folds under the outermost listed ancestor rather than opening a second level of nesting. Child rows carry no archive control: a sub-run's record belongs to the turn that produced it.

## Verification

`moon check` and `moon test` on `protocol` (15), `desktop/frontend/transcript` (30) and `desktop/frontend` (506) — all green. `moon info` + `moon fmt` run; the `.mbti` diffs carry only this change.

Sidebar geometry was checked by rendering the real `app.css` against the row markup in headless Chrome and measuring boxes: every conversation title lands at the same x, nested rows are pixel-identical to before, children sit one 20px step in. That caught one bug worth naming — the empty twisty cell was first classed `row-twisty empty`, and `.empty` already means an empty-state block in `app.css` (`padding: 32px 12px`), which stretched the 14px cell to 24px and pushed childless rows 9px right.

The desktop app itself has not been launched; that E2E pass is the reviewer's.

## Follow-ups (not in this PR)

- The durable `↳ subagent` link on the tool-result card, which needs `child_session` on the tool result rather than a parsed `brief`.
- Live sub-run progress in the transcript (`SubrunProgress`), so a 10-minute `explore` is not a silent gap between two notices.

### DIFF
--- a/desktop/app.css
+++ b/desktop/app.css
@@ -1417,7 +1417,7 @@ aside {
 .conversation-row {
   display: flex;
   align-items: center;
-  gap: 8px;
+  gap: 6px;
   min-width: 0;
   min-height: 32px;
   padding: 0 8px;
@@ -1618,9 +1618,11 @@ button.composer-worktree:hover {
   height: 12px;
 }
 
-/* A conversation nested under its workspace row. */
+/* A conversation nested under its workspace row: the twisty cell and the row
+   gap make up the rest of the indent, landing the title under its
+   workspace's. */
 .conversation-row.nested {
-  padding-left: 28px;
+  padding-left: 16px;
 }
 .conversation-row.placeholder {
   color: var(--ink-3);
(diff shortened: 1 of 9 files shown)


tonyfettes:
@codex review

PR #733, opened 2026-08-09
feat(desktop): add Review launcher entry

bobzhang:
## Summary

- present Browse, Files, and Review as first-class launcher destinations
- open Review directly in the changed-files inventory and refresh the Git snapshot
- keep Files explicit when switching back from Review
- label the picker and empty prompt for review context

## Scope

This is the small entry-and-routing slice. It reuses the merged changed-file source/full-diff view. Review progress and next/previous navigation remain a follow-up PR.

## Validation

- just check
- moon test --target js desktop/frontend/dock (18/18)
- moon test --target js desktop/frontend (367/367)
- moon run desktop/package/macos (packaged and code-sign validated)

## Review notes

- self-reviewed launcher visibility, picker reuse, Files/Review mode switching, change refresh, and remote behavior
- bounded fresh Codex CLI review re-ran both targeted suites; it produced no finding before its timebox ended
- native click-through E2E is pending only because the Mac session is locked; packaging itself passed

### DIFF
--- a/desktop/app.css
+++ b/desktop/app.css
@@ -795,6 +795,9 @@ html.is-resizing * {
   color: var(--ink-2);
   font-size: 13px;
 }
+.editor.mode-picker.mode-review .viewer-stack::after {
+  content: "Select a changed file to review";
+}
 /* The strip's new-tab button and its launcher popup. The tabsbar anchors
    the popup; the full-screen transparent backdrop catches the click away
    (reusing the topbar launcher's backdrop). */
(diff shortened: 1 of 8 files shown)


bobzhang:
Independent Codex review found one issue: the checked-in dock interface had not been updated for LauncherActions.review. Fixed in 152d0215. The checked-in interface now exactly matches the JS compiler output. Revalidated just check, dock 18/18, and frontend 367/367.

PR #735, opened 2026-08-09
feat(desktop): navigate changed-file reviews

bobzhang:
## Summary

- add previous and next changed-file controls with an accessible queue position
- support ArrowUp, ArrowDown, Home, and End on changed-file rows
- skip unavailable rows while preserving exact duplicate-path Git row identity
- move focus to the destination row after keyboard navigation
- keep Home and End as no-ops at their current queue boundary

## Validation

- `moon test --target js desktop/frontend/fileeditor` (90/90)
- `moon test --target js desktop/frontend` (367/367)
- `just check`
- `git diff --check`
- the stacked macOS app packaged successfully with `moon run desktop/package/macos`

## Review

- self-reviewed the exact four-file patch
- an initial fresh Codex CLI review found that focus stayed on the previous row after keyboard navigation; fixed with stable destination row IDs
- a second fresh Codex CLI review of the rebased PR found that Home or End at the current boundary reopened the same review and reset Full Diff to Source; fixed by making current-boundary navigation a tested no-op
- native click-through E2E remains pending because the Mac session is locked; packaging itself succeeds

### DIFF
--- a/desktop/app.css
+++ b/desktop/app.css
@@ -1268,6 +1268,46 @@ html.is-resizing * {
 .review-badge.hidden {
   display: none;
 }
+.review-navigation {
+  display: inline-flex;
+  flex: 0 0 auto;
+  align-items: center;
+  gap: 2px;
+}
+.review-navigation.hidden {
+  display: none;
+}
+.review-nav-button {
+  display: grid;
+  place-items: center;
+  width: 20px;
+  height: 20px;
+  padding: 0;
+  border: 1px solid transparent;
+  border-radius: 4px;
+  color: var(--ink-2);
+  background: transparent;
+  font: inherit;
+  font-size: 16px;
+  line-height: 1;
+  cursor: pointer;
+}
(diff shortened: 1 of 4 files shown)


bobzhang:
Fresh Codex CLI review found one P2 on the published patch: Home on the first row or End on the last row could reopen the same review and reset Full Diff to Source. Fixed in 0b2e02a6 by making current-boundary navigation a no-op, with regression assertions. All focused and root frontend tests plus `just check` pass after the fix.

bobzhang:
Nightly native failure is outside this patch and covered by the standing nightly exception: the nightly async runtime deadlocked in `desktop/internal/terminal/terminal_test.mbt` (`close_all_terminals retires every live session`). This PR changes only file-editor rendering/tests and CSS. Stable native completed successfully, as did stable/nightly JS; macOS and Linux packages are green. Waiting for the Windows package before merge. cc @tonyfettes

PR #743, opened 2026-08-09
refactor(editor): bind repeated peek guard calls

bobzhang:
## What

`open_references_peek_shell` and `schedule_references_peek_preview` both interleave browser mutations with an ownership re-check, because each step can hand control to a listener that supersedes the shell being built. The re-checks were spelled out in full at every site:

- 7 × 8-line `retire_provisional_references_peek_shell(state, generation, widget, mounted_view.val, mounted_zone_id.val, mounted_overlay.val, markdown_mounted.val)`
- 12 × 7-line `references_peek_preview_commit_is_current(generation, preview_generation, location.uri.to_string(), model, result_model, widget)`

Every site passed an **identical** argument list, so the guards read as noise and the actual sequence of browser work was hard to follow.

## Change

Bind each call shape to a local closure (`retire` / `still_current`) and call it at each site. Net: **-112 lines**, one file.

## Why this is behavior-preserving

- Both closures capture only immutable bindings plus the four `Ref`s.
- `retire()` reads `mounted_*.val` at call time — exactly as each original site did.
- `still_current()` re-runs `location.uri.to_string()` at call time — exactly as each original site did.
- `guard c else { ... }` replaces `if !c { ... }` with the same body.
- The one remaining inline `references_peek_preview_commit_is_current` is the guard that *binds* `result_model`, so it necessarily precedes the closure.

`editor/viewer/pkg.generated.mbti` is unchanged — no external surface moved.

## Testing

- `moon check --target js` — 0 errors (the 9 `lexscan` deprecation warnings are pre-existing in untouched `editor/syntax/lang_*` files)
- `moon test --target all` — native 1161, js 1738, wasm 1033, all passing
- Playwright smoke/component suites left to CI (`editor.yml` triggers on `editor/**`)

### DIFF
--- a/editor/viewer/definition_peek_host.mbt
+++ b/editor/viewer/definition_peek_host.mbt
@@ -649,6 +649,20 @@ fn Viewer::open_references_peek_shell(
   let mounted_zone_id : Ref[String?] = Ref(None)
   let mounted_overlay : Ref[@editor_browser.OverlayWidget?] = Ref(None)
   let markdown_mounted = Ref(false)
+  // Every abandon path below unmounts the same shell, and each reads the mount
+  // facts as they stand at that point. Binding the call once keeps the reads
+  // late while leaving one place to extend the resource set.
+  let retire = () => {
+    self.retire_provisional_references_peek_shell(
+      state,
+      generation,
+      widget,
+      mounted_view.val,
+      mounted_zone_id.val,
+      mounted_overlay.val,
+      markdown_mounted.val,
+    )
+  }
   match self.current_code_browser_data() {
     Some(browser) => {
       let zone_dom_node = @rdom.document().create_element("div")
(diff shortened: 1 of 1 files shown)


bobzhang:
## Codex review sign-off

**Verdict: no findings.**

> The new local closures preserve late evaluation of mutable mount refs and the preview currency predicate, so behavior remains equivalent. The editor's full MoonBit test suite passed on all supported targets.

Specifically confirmed:
- `retire()` still reads `mounted_view/mounted_zone_id/mounted_overlay/markdown_mounted` `.val` **at call time**, not at closure-construction time.
- `still_current()` still re-evaluates `location.uri.to_string()` and the contribution-state identity checks at call time, including from the deferred resize-observer and animation-frame callbacks.
- `guard c else { … }` is control-flow-identical to the `if !c { … }` it replaces.

PR #744, opened 2026-08-09
refactor(editor): dedupe unified diff subtree swap

bobzhang:
## What

`UnifiedDiffBrowserView` had the same subtree-swap sequence written out three times:

```moonbit
if self.disposed { return }
let next = render_…(…)
match self.root {
  Some(previous) => self.host.remove_child(previous.as_node())
  None => ()
}
self.host.append_child(next.as_node())
self.root = Some(next)
```

`set_hunks` and `set_message` were byte-identical apart from the one `render_*` call; `clear` repeated the detach half a third time.

## Change

- Extract `install(render)` and `detach_root()`. `set_hunks`/`set_message`/`clear` become one line each.
- `install` takes a **thunk**, not an element, so a disposed view still short-circuits *before* rendering — same as today.
- `render_line`: hoist `row_class` so the commentable branch appends to it (`"\{row_class} …-commentable"`) instead of respelling both base classes — they can no longer drift.
- `render_line`: two `match … { Some(n) => …; None => () }` attribute writes become `if … is Some(n)`.
- `comment_action_label`: merge the identical `Addition` / `Context` arms into `(Addition | Context, _, Some(number))`.

Net **-24 lines**, one file.

## Testing

- `moon check --target js` — 0 errors (9 pre-existing `lexscan` deprecation warnings in untouched `editor/syntax/lang_*`)
- `moon test --target all` — native 1161, js 1738, wasm 1033, all passing
- `editor/internal/viewer/browser/unified_diff/pkg.generated.mbti` unchanged — `install`/`detach_root` are private, no public surface moved
- Playwright smoke/component suites left to CI

### DIFF
--- a/editor/internal/viewer/browser/unified_diff/unified_diff_view.mbt
+++ b/editor/internal/viewer/browser/unified_diff/unified_diff_view.mbt
@@ -26,16 +26,7 @@ pub fn UnifiedDiffBrowserView::set_hunks(
   self : UnifiedDiffBrowserView,
   hunks : Array[@unified_diff_model.UnifiedDiffHunk],
 ) -> Unit {
-  if self.disposed {
-    return
-  }
-  let next = render_unified_diff(hunks, self.on_comment)
-  match self.root {
-    Some(previous) => self.host.remove_child(previous.as_node())
-    None => ()
-  }
-  self.host.append_child(next.as_node())
-  self.root = Some(next)
+  self.install(() => render_unified_diff(hunks, self.on_comment))
 }
 
 ///|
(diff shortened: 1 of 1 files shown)


bobzhang:
## Codex review sign-off

**Verdict: no findings.**

> The refactor preserves the existing rendering, disposal, and DOM replacement behavior. MoonBit checks and the full editor test suite pass; browser smoke testing was blocked only by sandbox cache permissions.

Codex independently ran `just editor-test` during the review — 1033 wasm / 1738 js / 1161 native, all passing. Its `just editor-test-browser` attempt failed only on `Unable to lock dependency source cache … Operation not permitted`, i.e. its own sandbox could not write `~/.moon/cache/deps`; that gate runs for real in the `browser-correctness` CI job.

Key behavior-preservation point confirmed: `install` takes a thunk rather than an already-rendered element, so a disposed view still skips rendering entirely instead of building a subtree and discarding it.

PR #760, opened 2026-08-09
fix(desktop): search workspace stores before the global root in known_session_roots

tonyfettes:
## Summary

`known_session_roots` was ordering stores as **(1) global root, (2) workspace stores** — the reverse of the old `session_store_root` priority (which searched workspaces first, then fell back to global). This meant that if the same session id existed in both, `live_record_root` and `archived_record_root` would return the global record instead of the workspace record the sidebar row was targeting.

This PR swaps the order: **workspace stores first, then global root as fallback** — matching the old behavior.

## Why not a bigger change

The Codex review on #758 suggested passing workspace identity through the frontend message. That would add complexity to the `ArchiveSession` message schema and every call site, while the real fix is just reordering one array — the client hint is deliberately not trusted (the comment on `live_record_root` explains why), so the search order is all you need.

## Practical impact

Session ID collisions across stores are near-impossible in practice (fingerprint entropy), so this is correctness work, not a bugfix for a user-facing issue. But the old ordering was clearly wrong and costs nothing to correct.

Stacked on #758.


tonyfettes:
Closing without merge: main has moved past this.

- #892 gave every conversation a project and c967ff6c2 removed the legacy global session root, so the global store is no longer one of the searched session roots. `known_session_roots` became `@session_store.known_roots()` (#891) and only enumerates registered workspaces. There is no global entry left to reorder.
- The direction Codex suggested here (carry the selected store through the request instead of resolving an ambiguous id by search order) is what #867 did for delete and what #886 does for archive/unarchive/load.

Left on main from the same cleanup: the `unarchive_session` doc comment still says "the global root first", and the `claim_session_family_move` refusal still names "the global store". Worth checking whether #886 already rewrites those.

PR #800, opened 2026-08-11
fix(desktop): simplify prompt delivery settlement

tonyfettes:
## Summary

- treat an accepted `agent.start` response or matching `Started` event as the ownership receipt for an initial prompt
- remove durable-persistence recovery, unconfirmed-prompt notices, and archive salvage for accepted initial prompts
- restore the draft immediately when the start request fails or the host does not accept it
- move a cancelled engine out of `Running` before publishing `Finished`, so immediate archive observes the settled state

## Why

The desktop kept an initial prompt pending after the host had accepted it, waiting for later durable-record evidence. Interrupt, process-exit, reconnect, and archive paths then had to recover or reclassify that prompt. A narrow timing window produced a prominent “prompt delivery could not be confirmed” notice even though the start had already been accepted.

The host already provides explicit acceptance through the start response and the matching `Started` notification, so the desktop now settles ownership at that boundary.

## Impact

Interrupting or archiving an accepted run no longer restores the prompt or shows the unconfirmed-delivery warning. A failed start still restores the exact draft. Steer and goal reconciliation are unchanged.

The separate speculative `session.load` retry behavior for a brand-new conversation is intentionally outside this PR.

## Validation

- `moon -C desktop check frontend --target js`
- `moon -C desktop test frontend --target js` — 424 passed
- `moon -C desktop check internal/engine`
- `moon -C desktop test internal/engine` — 123 passed
- `git diff --check`

The frontend check still reports the repository's existing deprecated `lexscan` warnings.

### DIFF
--- a/desktop/frontend/bridge.mbt
+++ b/desktop/frontend/bridge.mbt
@@ -999,15 +999,7 @@ async fn request_agent_start(
   let reply = @interop.bridge_request(channel, @commands.agent_start, payload) catch {
     error => {
       let message = @interop.bridge_error_text(error)
-      scheduler.add(
-        dispatch(
-          if @interop.bridge_error_is_definite(error) {
-            StartRejected(session~, submission_id~, message~)
-          } else {
-            StartUndelivered(session~, submission_id~, message~)
-          },
-        ),
-      )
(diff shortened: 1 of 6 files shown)


tonyfettes:
Replaced by #882 + #940

PR #802, opened 2026-08-11
Align Markdown document font size with app UI

bzy-debug:
## Summary

- derive Markdown document prose size from the host `--ui-font-size` token
- keep the existing `14px` baseline for standalone Viewer embedders without that token
- cover prose, tokenized fenced code, heading scale, fallback behavior, scroll state, and fold-control geometry

## Why

The Markdown document article hard-coded a `14px` base size while the OpenSeek workbench defines `--ui-font-size: 13px`. As a result, rendered Markdown looked noticeably larger than the surrounding application UI.

## User impact

Markdown documents in OpenSeek now use the same base font size as the rest of the workbench. Standalone Viewer consumers retain the previous `14px` behavior unless they opt into a UI font-size token.

## Validation

- `just check`
- `just editor-test`
- `just test`
- `just build`
- `just editor-test-browser` — 143 passed
- `moon -C desktop run --target native package/macos`
- strict recursive `codesign` verification and visual inspection of the packaged `SeekMoon.app`

### DIFF
--- a/editor/internal/viewer/browser/markdown_document/markdown_document.css
+++ b/editor/internal/viewer/browser/markdown_document/markdown_document.css
@@ -35,7 +35,9 @@
   box-sizing: border-box;
   min-height: 100%;
   padding: 16px 24px 64px;
-  font-size: 14px;
+  /* Match themed workbench prose to its surrounding UI while preserving the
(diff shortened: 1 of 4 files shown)


tonyfettes:
macOS app valid on disk, merging

PR #814, opened 2026-08-11
Fix submodule handling for managed worktrees

bobzhang:
## Summary

- provision each managed worktree with the submodules already initialized in the attached main checkout, including recursive children
- keep optional, uninitialized submodules opt-in
- apply the same provisioning during checkout repair and roll back partial creation safely on failure
- remove gitlink-containing worktrees with delete-plus-prune because Git refuses `worktree remove`, while still honoring external worktree locks
- add an offline lifecycle regression covering creation, optional submodules, repair, locking, and removal

## Root cause

`git worktree add` checks out the superproject gitlinks but does not initialize their working directories. OpenSeek's root `moon.work` includes packages under `desktop/lepus`, so a fresh managed worktree could not resolve the workspace. Cleanup also needed adjustment because Git refuses to move or remove any worktree whose tree contains submodule gitlinks.

## Validation

- `moon test desktop/internal/engine` (125 passed)
- `just check`
- `just test` (native 3412 passed, JS 2822 passed, cram 27 passed)
- `just build`
- `moon info desktop/internal/engine` (no public interface diff)

### DIFF
--- a/desktop/internal/engine/worktree_submodules.mbt
+++ b/desktop/internal/engine/worktree_submodules.mbt
@@ -1,0 +1,197 @@
+///|
+/// Exact gitlink paths from `git ls-tree -z` or `git ls-files --stage -z`.
+/// Reading Git metadata keeps paths with spaces unambiguous and still sees
+/// submodules whose working directories are absent.
+fn gitlink_paths(metadata : String) -> Array[String] {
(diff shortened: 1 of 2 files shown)


bobzhang:
cc @tonyfettes

PR #820, opened 2026-08-12
fix: restore live reasoning stream

tonyfettes:
## Summary

- restore `reasoning_delta` as a typed engine event and emit each provider reasoning fragment
- show live, transient thinking in Desktop and TUI while a turn is running
- keep completed reasoning durable through the existing session commit path, without duplicating transcript items

## Why

PR #214 intentionally stopped forwarding streaming reasoning deltas to avoid logging sensitive intermediate content. Later typed-protocol cleanup removed the now-dead client handling as well. The result is that long reasoning phases look stalled, especially in Desktop.

This restores live client feedback using the current typed protocol. Desktop renders partial reasoning as subdued plain text and does not persist it as a transcript item; the complete reasoning message still comes from the existing durable session event.

## Validation

- `moon test -p bobzhang/openseek_protocol --target native` (15 passed)
- `moon test agent/steer_test.mbt --target native --filter '*reasoning*'` (1 passed)
- `moon test cmd/tui --target native` (84 passed)
- `moon -C desktop test frontend --target js` (444 passed)
- `moon -C desktop test internal/event internal/remote --target native` (17 passed)
- relevant protocol, agent, TUI, and Desktop checks pass with `--deny-warn`
- `moon fmt --check`
- generated Desktop transcript interface matches compiler output
- `git diff --check`

### DIFF
--- a/agent/steer_test.mbt
+++ b/agent/steer_test.mbt
@@ -335,7 +335,7 @@ async fn reasoning_log_test_server(group : @async.TaskGroup[Unit]) -> String? {
 }
 
 ///|
-async test "streaming reasoning deltas are skipped in logs" {
+async test "streaming reasoning deltas reach the engine event stream" {
   @async.with_task_group() <| group => {
     guard reasoning_log_test_server(group) is Some(url) else { return }
     let entries : Array[Json] = []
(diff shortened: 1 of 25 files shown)


tonyfettes:
Close b/c of #825

PR #823, opened 2026-08-12
ci(desktop): automate MoonBit toolchain updates

tonyfettes:
## Summary

- add a daily and manually dispatchable workflow for the Desktop MoonBit toolchain pin
- install the nightly toolchain and parse the compiler version from `moonc -v`
- update `desktop/.moonbit-version` only when the version changes
- keep one automation-owned branch and pull request for subsequent nightly updates

This keeps Desktop packages pinned to a reviewed compiler version while removing the need for manual version bumps.

## Validation

- parsed the workflow as YAML
- ran ShellCheck over every workflow shell block
- ran `git diff --check`
- verified the version parser against the current local `moonc -v` output

## Operational note

The repository currently disables GitHub Actions from creating pull requests. The workflow needs the repository setting **Allow GitHub Actions to create and approve pull requests** enabled before its `gh pr create` step can succeed.

### DIFF
--- a/.github/workflows/desktop-moonbit-update.yml
+++ b/.github/workflows/desktop-moonbit-update.yml
@@ -1,0 +1,79 @@
+name: Desktop MoonBit toolchain update
+
+on:
+  # GitHub cron is UTC: run daily at 10:17 Asia/Shanghai.
+  schedule:
+    - cron: "17 2 * * *"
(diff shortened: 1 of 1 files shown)


tonyfettes:
This actually needs human intervention like fix warnings when upgrading toolchains, etc.

PR #828, opened 2026-08-12
refactor(desktop): split and package application stylesheet

tonyfettes:
## Summary

- split the 4,602-line desktop stylesheet into responsibility-based source files
- keep `desktop/app.css` as an ordered development import manifest for refresh-only CSS iteration
- assemble the same sources with MoonBit during production packaging and ship the generated result as `app.css` on browser, macOS, Linux, and Windows
- document the development and packaging flow and test that the import order matches the production order

## Impact

Development continues to load CSS directly from the checkout, so CSS edits only need a page refresh. Production packages still contain one `app.css`; the concatenated contents are byte-for-byte identical to the previous stylesheet.

## Validation

- `moon info && moon fmt`
- `moon test desktop/package/internal/packaging --target native` (12/12)
- `moon fmt --check`
- `git diff --check`
- `moon -C desktop run --target native package/macos -- --target app`
- verified the packaged `app.css` matches `desktop/app.generated.css`
- launched the packaged app with Computer Use and checked the main screen, Settings page, and dark theme


tonyfettes:
@codex review

PR #833, opened 2026-08-12
refactor(desktop): use semantic command replies

tonyfettes:
## Summary

- keep `Command` at two generic parameters and give each command a concrete, semantic reply type
- catch and log operation failures at the command registration boundary, then return the command-specific error variant to the frontend
- preserve the existing JSON wire shapes while updating frontend handling, protocol tests, and remote protocol documentation
- replace generic acknowledgement and success/failure wrappers with outcomes such as `GitChanges`, `NotRepository`, `TerminalOpened`, and command-specific error variants

## Why

Native operation failures were raised through Proton, so the UI could collapse them into messages such as `op ext:openseek/agent.start failed` without the original error. Returning failures as typed command replies keeps the original message available to the frontend while retaining structured logging.

## Validation

- `moon check desktop/internal/protocol --target native`
- `moon check desktop/internal/commands --target native`
- `moon check desktop/internal/host --target native`
- `moon check desktop/internal/extension --target native`
- `moon check desktop/frontend --target js`
- protocol tests: 15 passed
- endpoint tests: 5 passed
- engine tests: 125 passed
- host tests: 64 passed
- frontend tests: 441 passed
- `git diff --check`

`just check` was not run because `just` is not installed in the local environment.

### DIFF
--- a/desktop/frontend/archive.mbt
+++ b/desktop/frontend/archive.mbt
@@ -18,17 +18,18 @@ fn fetch_archived(
     @js.async_run(() => {
       // A host without the op (or without a session store) just reads as an
       // empty Archived group.
-      let reply = @interop.bridge_request(
+      let response = @interop.bridge_request(
         channel,
         @commands.session_list_archived,
         @protocol.EmptyPayload::NoPayload,
       ) catch {
         _ => return
       }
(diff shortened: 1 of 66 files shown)


tonyfettes:
Should be handled command by command

PR #851, opened 2026-08-13
test(desktop): skip flaky terminal tests

tonyfettes:
## Summary

- skip `close_all_terminals retires every live session` at runtime with `#skip`
- skip `terminal session runs in the requested directory` after both native CI channels moved the same deadlock there
- document why both temporary skips exist
- keep both test bodies type-checked so they do not silently rot

## Why

Native Linux CI can intermittently abort in the async deadlock detector while the terminal pump and session tasks are still alive. After the close-all regression was skipped, both stable-native and nightly-native failed in the requested-directory test through the same pump path. These skips temporarily remove the two observed trigger points while retaining the tests for a later teardown fix.

## Impact

Two terminal integration regressions are temporarily not executed. All other terminal package tests continue to run, and both skipped tests remain type-checked.

## Validation

- `moon -C desktop test --target native internal/terminal` — 9 passed, 0 failed
- `moon -C desktop fmt internal/terminal/terminal_test.mbt`
- `moon -C desktop info internal/terminal`
- `git diff --check`

### DIFF
--- a/desktop/internal/terminal/terminal_test.mbt
+++ b/desktop/internal/terminal/terminal_test.mbt
@@ -163,6 +163,10 @@ async test "shutdown rejects pending and later terminal opens" {
 ///|
 /// `close_all_terminals` (the connect handshake's sweep) retires every live
 /// session; each one still reports its exit.
+///
+/// Temporarily skipped because native Linux CI can abort here after its async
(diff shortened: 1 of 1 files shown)


tonyfettes:
This should be fixed by upgrading pty to 0.4.0

PR #865, opened 2026-08-13
refactor(desktop): key transcript diagram lifetimes

tonyfettes:
## Summary

- project transcript content into channel/session-scoped keyed Rabbita blocks so unchanged historical Markdown is not reparsed on streaming updates
- have Markdown report exact generated diagram wrappers and give each wrapper its own after-render viewport subscription
- add a public exact-wrapper `DiagramViewport` lifetime while keeping the existing multi-diagram API for Editor hosts
- leave the transcript root observer responsible only for mount, geometry, and local-image DOM signals

## Stack

- base: #857 (`codex/shared-diagram-viewports`)
- rebase this PR onto `main` after #857 merges

## Validation

- `moon fmt --check`
- `moon test --target native` (3447 passed)
- `moon test --target js` (2905 passed)
- `moon cram test tests/cram` (27 passed)
- `moon test --target all` (editor: 1372 wasm + 295 wasm-gc passed)
- Playwright smoke + component suites (143 passed)
- `moon build --target native`
- `moon build --target js`

`just` is unavailable in this checkout environment, so the repository recipes were expanded and run directly.

### DIFF
--- a/desktop/frontend/markdown/markdown.mbt
+++ b/desktop/frontend/markdown/markdown.mbt
@@ -6,6 +6,45 @@
 // text instead of being injected into the DOM.
 
 ///|
+/// One generated diagram that an incremental host can mount after rendering.
+/// The DOM id names the exact wrapper; the revision changes only when the
+/// diagram source at that position changes.
+pub struct MarkdownDiagram {
+  dom_id : String
+  revision : String
+} derive(Eq)
+
+///|
+pub fn MarkdownDiagram::dom_id(self : MarkdownDiagram) -> String {
+  self.dom_id
+}
+
+///|
+pub fn MarkdownDiagram::revision(self : MarkdownDiagram) -> String {
+  self.revision
+}
+
+///|
+/// Markdown nodes plus the exact generated diagrams contained in those nodes.
+/// Hosts that do not own diagram lifetimes can continue using `markdown_nodes`.
+pub struct RenderedMarkdown {
+  nodes : Array[@html.Html]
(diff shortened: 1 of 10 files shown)


tonyfettes:
Closing this as superseded by #1260.

#1260 replaced the transcript component this PR was built on: the projector now emits one `TranscriptItem` per durable event (an Assistant item keeps its reasoning and tool calls), and the renderer is a per-item keyed loop with stable `row\0s<sequence>` keys. The block pipeline here (`blocks()`, `RenderedTranscriptBlock`, per-block `diagram_targets`) targeted the flattened model that #984 and #1260 have since rewritten twice, so rebasing would mean rewriting the transcript half of the diff.

What survives, split into two follow-ups:

1. **Per-item memoization** (#1261). The `assoc_by` idea from this PR, redone on the merged model: `input.map(blocks)`, then `assoc_by(by=scope + row key)` with a pure `map(render)` per branch, so streaming deltas only re-render the live row. It keeps today's root `DiagramViewports` observer; its `refresh()` already retains controllers whose SVG is still in the DOM, so diagram interaction state is unaffected.
2. **Per-wrapper after-render diagram subscriptions** (#1262). The `markdown` change (report the exact generated wrappers) and the `diagram_viewport` change (single-wrapper `DiagramViewport` lifetime) from this PR apply as-is and can be cherry-picked from 0e0885636 when we do this. Deferred until memoization has landed and we have measured whether the root-observer scan is worth removing.

Thanks for the design work here. The memoization shape is the one we are keeping.

PR #881, opened 2026-08-15
chore: upgrade rabbita to 0.15.1

tonyfettes:
Upgrades all four modules (root, `cmd/viz_app`, `desktop`, `editor`) from rabbita 0.14.3 to 0.15.1. Rebased onto `main` after #876/#877/#879/#880.

## Why

0.15.1 fixes the positional-slot aliasing at its root — event slots moved off the shared `Props` onto per-element `IProps` — so a cached component's `Html` shifting position no longer re-targets its click commands at another row's element. That is the bug #876's keyed-`Map` container pattern works around.

## The catch

0.15.1 removes the js-target `@rabbita.render_to_string`. Server-side rendering became the native-only async `App::render`, and `internal/vdom.server_side_render` is not importable across modules. `moon.work` also resolves one version for the whole workspace, so the four modules move together.

**`viz` keeps its rendering tests** by rendering on native instead. That package was js-only by convention, not by need — it imports nothing but `@rabbita/html` and this module's own packages — so declaring `js+native` makes `App::render` reachable and all 45 assertions come back verbatim (each fragment is wrapped in a document; the assertions still only look inside the body). They live in `viz/render_test`, a native-only black-box package, because the rabbita and async imports they need would otherwise be unused in the js build and `moon check --target js --deny-warn` rejects that. Shared fixtures moved to `viz/testkit`, following `testkit/filesystem`. Parsing and the projection counts stay in `viz` and still run on js — that is the backend the viewer ships to, and String semantics differ there.

**The desktop packages cannot do this**: they reach `interop`, `@rabbita/dom`, and `@rabbita/js`, which are genuinely browser-bound. There, each markup assertion that was *deciding* something now asserts the pure helper the view reads:

| package | helper |
|---|---|
| `plan` | `fill_width`, `summary_line`, `completed_count`, `active_step`, `status_class` |
| `markdown` | `diagram_svg`, `local_file_path`, `local_image_path` |
| `composer` | `GoalInput::row`, `GitPresentation::button_class` |
| `frontend` | `@transcript.session_tree`, `is_archived_read_only` |

Those are new private extractions of logic that was inline in a view — no behavior change, no `.mbti` change.

## What was dropped

Desktop assertions with no pure counterpart, each noted in a comment where it stood:

- the `visually-hidden` speaker labels reaching assistive technology
- the `data-focus-owner` / `data-focus-target` contract on every rendered form element (Settings, Skills, Quick Open, browser address bar, composer)
- tool call ↔ result anchor ids and cross-links
- dock launcher order, fileeditor breadcrumb picker and review surfaces, and the Git popover's rows

`goal_banner` was deleted too: the composer has owned that row since #863, so the removed test was its only caller (and `icon_target` / `icon_close` went with it).

The sidebar's own markup tests needed nothing here — #876 had already replaced them with component-`Input` assertions.

## An async 0.20.5 casualty

rabbita 0.15.1 depends on `moonbitlang/async@0.20.5` (0.14.3 depended on async not at all), so MVS lifts the workspace from 0.20.4. async 0.20.5's wasm backend declares a host import — `thread_pool/make_open_stat_job` — that no moonrun provides, including the `latest` stable CI installs, so every wasm program linking `async/fs` fails at `WebAssembly.Instance()`. The extern carries `#unsafe_skip_stub_check`, which is why a missing host function surfaces as a run-time LinkError instead of a build error. Reported upstream.

That took out `moon test --target all` (the editor-server host blackbox test) and the browser smoke suite, whose Playwright webServer *is* that host. The third commit runs the reference server on native instead — nothing runs it on wasm in production, it already declared `native+wasm`, and the editor library keeps its 998 wasm tests. Restore `native+wasm` in `editor/server/moon.mod` once async ships a fix; the note is in the file.

## Verification

- `desktop`: `moon check --deny-warn` and `moon test` on js (2851 passed) and native (3511 passed)
- `editor`: `moon check --target js --deny-warn`
- root: `moon check --deny-warn` on js and native; `moon test --target js viz` (10 passed), `moon test --target native viz viz/render_test` (56 passed)
- `moon fmt` + `moon info` clean — the only `.mbti` change is `viz/pkg.generated.mbti`, which `moon info` had been skipping while the package was js-only (`subrun_child_ids`, `SubrunView`, `render_session`'s `subrun_children?`)

Not verified here: the slot-aliasing fix itself is DOM behavior and needs a run of the real app.


tonyfettes:
We will wait for upstream to release another 0.15.x with JS-backend HTML to string snapshotting features.

PR #889, opened 2026-08-15
feat(desktop): integrate DeepSeek Harness

tonyfettes:
## Summary

Run a local DeepSeek Harness (`dsh web`) process beside the existing OpenSeek engine and Codex app-server, and present dsh as **one more agent rather than one more place**: its conversations share the sidebar's Projects section with OpenSeek's and Codex's, and one is started by picking a dsh model in the composer's unified model chip.

- transcript, composer, Files, Editor, and Terminal are the existing shared surfaces — dsh supplies history, live assistant deltas, tool calls, approvals, and questions through them
- queue a prompt, steer a running turn, cancel it, and answer single-select / multi-select / free-text questions
- **Worktree** mode: Desktop cuts its normal Git worktree, registers it as a dsh Workspace, starts a conversation inside it, and sends the first prompt there
- archive a conversation to hide it from every grouping surface
- configure the DeepSeek API key from the app, in a dsh home this process owns
- startup failures surface with dsh's own error rather than a generic "not on PATH"

### The command catalog is the capability list

The renderer never names a dsh method. Each operation it may perform is its own Proton command with its own typed request — **17 commands, and no generic proxy**:

`dsh.status` · `dsh.pending` · `dsh.workspace.list` · `dsh.workspace.create` · `dsh.workspace.archive_session` · `dsh.session.list` · `dsh.session.create` · `dsh.session.history` · `dsh.session.models` · `dsh.session.select_model` · `dsh.session.prompt` · `dsh.session.cancel` · `dsh.llm.models` · `dsh.credentials.describe` · `dsh.credentials.set` · `dsh.credentials.unset` · `dsh.respond`

dsh ships more than fifty RPC methods. The rest — `settings.*`, `host.*`, `command.execute`, `agentPreset.*`, `goal.*`, `subagent.*`, `session.fork`, `workspace.delete`, `workspace.insertSessionBefore`, … — are unreachable because nothing names them, and a method a later dsh release adds stays unreachable until a command is written for it. The capability surface belongs to this repository rather than to whichever dsh happens to be installed.

Two request shapes are narrowed here rather than relayed: `session.history` takes its window size from the host, so the page cannot ask for a whole log, and `session.prompt` carries one text string that the endpoint expands into dsh's typed content array, so a part the page never authored cannot ride the call.

Two replies carry a native fact beside dsh's envelope, and both are *required* fields of dedicated reply types rather than optional ones every caller re-checks — `dsh.workspace.list` carries the authorization fence's verdict, and `dsh.session.list` carries the actor revision captured when its HTTP call actually started.

### Transport

dsh's existing carriers, unchanged: `POST /api/<method>` with complete ApiProxy `client-request` / `server-response` envelopes, `/api/events.mux` for all-session events, `/api/events.host` for host lifecycle, and `POST /api/respond` for approval and question answers. The native process is the only component that knows the loopback port, which keeps dsh's Host/Origin checks intact instead of adding CORS or pointing the renderer at localhost.

### Filesystem authority

dsh's Workspace registry decides which directory a conversation works in — never the session's display-only `cwd`. Selecting an accounted session opens Files, Editor, and Terminal at its Workspace root; an unaccounted one stays usable as a conversation but receives no filesystem authority. Grants are revoked as soon as either WebSocket drops and restored only by a fresh authoritative `workspace.list`.

### Cleanup, and what dsh does not offer

Archiving is the only cleanup verb, because dsh has **no session deletion at all**. It is also irreversible: dsh's archive set is append-only, and its own API documents unarchive as a future capability. The sidebar therefore confirms first and offers no Restore on dsh rows, unlike OpenSeek's and Codex's.

A Desktop-created dsh worktree records **no owner** in the worktree registry. dsh's Workspace registry decides where a conversation runs, that mapping changes without Desktop's involvement, and dsh has no operation that durably means "this conversation ended" — so a recorded owner could only go stale, and a stale owner is what makes a checkout impossible to clean up. `worktree.remove` therefore *asks* dsh's current `workspace.list` and refuses while an unarchived conversation lives anywhere inside the checkout. A dsh that has answered before but is unreachable now refuses too, since its registry is durable; only a dsh that has never answered in this process's life is skipped.

### Isolated home, and the DeepSeek key

`dsh web` gets a dedicated `DSH_HOME` under the per-user runtime directory — `<runtime_dir>/dsh`, created `0700` on POSIX — the same shape the Codex integration already gives its app-server. dsh keeps `settings.yaml`, the `.credentials.yaml` store, session logs, profiles, and storages there. The user's own `dsh` CLI home is neither read nor written, and clearing the app's runtime state removes every dsh artifact Desktop created. The costs mirror Codex's and are accepted: CLI conversations are absent from the sidebar and provider configuration starts empty.

The isolated home is also what makes dsh's credential seam reachable. dsh resolves a reference from four layers, and the inherited process environment wins and is **read-only**:

```text
inherited process environment      (read-only, wins)
> $DSH_HOME/.credentials.yaml      (provider-managed, writable)
> <invocation cwd>/.env            (read-only fallback)
> $DSH_HOME/.env                   (read-only fallback)
```

While the environment supplies the reference, `credentials.set` *refuses* rather than writing a value resolution would keep ignoring. The spawn therefore sets `DEEPSEEK_API_KEY` to the empty string, which that layer reads as absent — `extra_env` can add entries but not remove them — leaving the store in the isolated home as the writable source. Both overrides exist only when a home was prepared: without one there is no store of ours to reach, so the environment passes through untouched, and the page reports that honestly because `describe` returns `writable: false` and names the layer.

The three credential commands are the narrowest exception to the catalog rule rather than a widening of it. dsh's own methods take a reference name; these take none — the reference is a host constant, so the page configures the shipped DeepSeek route and cannot address any other credential dsh holds. `describe` carries no value in either direction, and `set` is the only direction a secret travels.

A missing key is not a startup failure: dsh starts, lists models, and opens conversations without one, and only a turn fails with `MISSING_CREDENTIAL`. The key card therefore appears above the composer — where that is about to be discovered — and can be dismissed with **Not now**, since reading an old conversation or configuring the key elsewhere are both reasonable things to be doing instead. The dsh topbar control is the way back.

Settings carries the same key under a **DeepSeek Harness** group beside Codex's, and both surfaces drive one page state: a key typed in either is the same draft, and neither can disagree with the other about what is stored. They differ in when they appear, which is the point — the card catches the user about to send a prompt that cannot run and hides once a key works, while the Settings row is drawn whether or not anything is wrong. Neither can prefill the field: the seam never returns a value, so it always starts empty, and "Clear" is a separate operation rather than saving a blank.

### Limits

dsh runs inside this native process, so its rows appear only under the local device's projects and the chip offers dsh only for a conversation that runs there. While a remote device is focused, dsh is absent from both surfaces. The commands are registered on the embedded Desktop extension alone — they are absent from the remote WebSocket catalog, `dsh.status_changed` / `dsh.event` are filtered before relay delivery, and the rows and model group are absent from browser-console builds.

Desktop does not bundle DeepSeek Harness: `dsh` and the Node environment it needs must already be on the login-shell `PATH`. What it configures is the home it points dsh at and the one credential inside it — everything else in that home stays dsh's.

## Validation

- `moon test --target js` — 3082 passed
- `moon test --target native` — 3464 passed
- `moon fmt` / `moon info` clean

E2E of the full flow is pending.


tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
@codex review

tonyfettes:
We don't have a good way to package dsh into SeekMoon without blow off the final package size. To package a dsh we need node.js + many HUGE node dependencies, like the claude agent sdk and node-pty, leaking to a few hundred of MiBs increase in final bundle.

PR #890, opened 2026-08-15
feat: add per-conversation sandbox opt-out

tonyfettes:
## Summary

- add a conversation-scoped `Sandbox` / `No sandbox` pill with shield icons immediately after the `Local` / `Worktree` control
- propagate the choice through start, compact, and goal requests; include it in the per-conversation engine fingerprint so a changed choice gets a matching serve process
- add `--no-sandbox-exec` / `OPENSEEK_NO_SANDBOX_EXEC=1`, inherited by sub-runs, while keeping the sandboxed default for older clients

## Validation

- `moon info`
- `moon fmt`
- `moon test desktop/frontend --target js` — 447 passed
- `moon test desktop/internal/protocol --target native` — 28 passed
- `moon test desktop/internal/engine --target native` — 138 passed
- `moon test cmd/openseek --target native` — 70 passed
- `moon test agent_tool/internal/sandbox --target native` — 18 passed
- explicit `OPENSEEK_NO_SANDBOX_EXEC=1` opt-out test — 1 passed

## Environment notes

- `just check` could not run because `just` is not installed in this environment.
- the standalone `desktop/internal/api` native test target did not link four Proton host-loop symbols; the workspace native check and protocol/engine suites passed.


tonyfettes:
Not planned. Should be replaced by myshell + fine-grained sandboxing.

PR #919, opened 2026-08-18
fix(windows): harden shell encoding, fix binary output handling, and align shell description

Betterlol:
Windows shell toolchain fixes:

### 1. Use powershell.exe (always available inbox)
`pwsh` (PowerShell 7) is an optional install — many systems lack it.
`@process.spawn(pwsh)` returns Access denied when the executable
is not found. Use `powershell.exe` (Windows built-in), always
available, with identical `-NoProfile -Command` semantics.

### 2. Set UTF-8 output encoding in PowerShell args
Prepends `$OutputEncoding=[Console]::OutputEncoding=[Text.Encoding]::UTF8;
chcp 65001>$null;` to every Windows shell command so that non-ASCII output
(e.g. Chinese characters) arrives as valid UTF-8 bytes instead of the
systems active code page (GBK/CP936).

### 3. Return binary output as lossy text instead of a hard error
Removes the `ShellBinaryOutput` variant. When the foreground execution
model detects invalid UTF-8 bytes, it now returns the lossy-decoded text
with an appended `<system>encoding=lossy</system>` warning.

### 4. Suppress console window pop-ups on Windows
Adds `no_console_window=true` to three `@process.spawn`/`@process.run`
call sites that were missing it.

### 5. Align shell_description with the actual shell program
`shell_description()` now tells the model to use PowerShell 5.1 syntax
and explicitly notes that chain operators (`&&`, `||`) are NOT supported,
preventing parser errors from pwsh-only syntax.

### 6. Let lossy-output commands finish normally
Previously `exec.stop()` was called immediately on invalid UTF-8,
killing the child and making `exit_code=None` rendered as `exit=0` —
so `printf '\377'; sleep 1; touch marker` was reported as success
even though `touch marker` never ran. Now `exec.wait()` lets the child
finish normally, preserving the real exit code.

### 7. Decode lossily at EOF consistently
The `None`/EOF branch of `read_output_prefix()` used strict
`@utf8.decode()` which raised `Malformed` on invalid bytes, while the
intermediate chunk branch used `decode_valid_utf8_prefix()` with lossy
fallback. Changed to `decode_valid_utf8_prefix()` for consistency.


agent_tool/internal/platform_shell/platform_shell.mbt
```diff
@@ -3,24 +3,38 @@
...
 
 ///|
-/// `pwsh` on Windows — see the note above.
+/// `pwsh` on Windows — always preferred for its modern syntax (`&&` and other
+/// PowerShell 7+ features). Falls back to the inbox `powershell.exe` (Windows
+/// PowerShell 5.1) only when `pwsh` is unavailable, so PowerShell 7-only
+/// syntax never produces a parser error. The availability probe is cached
+/// process-wide.
 #cfg(platform="windows")
-pub let program : String = "pwsh"
```
tonyfettes, line 18: Sorry on second thought I think now the probing logic is just too complex. Can we switch back to hardcoded powershell?

Betterlol: @tonyfettes Done, `default: powershell.exe`.

PR #920, opened 2026-08-18
fix(desktop): let the folder picker reach other Windows drives

### LINKED ISSUE #899: windows file explorer cannot choose other drive letter

tonyfettes:
Fixes #899.

The in-app "Add a project" picker had no way to another drive letter on Windows: the breadcrumb stopped at the drive root (`C:`) and the host reported no parent above it, so `D:\` was unreachable except by pasting a path. This adds a "This PC" level above the drive roots, the same shape Windows Explorer uses.

## Host (`internal/host/browse_ops.mbt`)

- Browsing the resource-path root `/` on a Windows host answers a new `Drives` reply listing the mounted drives (`C:`, `D:`, ...), found by probing every letter with `@fs.kind` — async, cheap next to the listing's readdir, and a letter without media simply fails the probe.
- POSIX hosts keep serving `/` as an ordinary root listing (pinned by a new test).

## Protocol (`internal/protocol/desktop_replies.mbt`)

- `BrowseDirectoryReply` gains the `Drives(entries~)` variant (`{"drives": [...]}` on the wire), with a JSON round-trip test. This is the only `pkg.generated.mbti` change.

## Frontend

- The picker's state holds an explicit `PickerLocation` — `Directory(uri)` or `DriveList` — instead of a bare path.
- The breadcrumb always roots the trail: `/` on POSIX, "This PC" on Windows paths; that crumb navigates to the drive list, and drive rows enter `/C:/`-style roots.
- Confirm is refused at the drive list (update guard + disabled Add button): it is not a directory anyone can register.

## Verification

- `just check` (deny-warn on both targets + `moon fmt --check`) clean.
- Full `moon test --target native` (3419) and `--target js` (2995) pass, including the new host, protocol, and picker-state tests.
- Windows cannot be exercised from this machine — that is what this draft is for. Manual check: open Add a project → trail shows `This PC › C: › ...` → click "This PC" → drive rows appear → click `D:` → it lists `D:\`; "Add" is disabled only while on the drive list itself.

### DIFF
--- a/desktop/frontend/model.mbt
+++ b/desktop/frontend/model.mbt
@@ -1355,8 +1355,9 @@ priv struct GoalEdit {
 } derive(Eq)
 
 ///|
-/// The in-app folder picker's state: the host directory currently shown and
-/// its subdirectories. `loading` marks a browse in flight; a failed
+/// The in-app folder picker's state: the host location currently shown and
+/// the rows inside it — a directory's subdirectories, or the Windows drive
+/// list's drives. `loading` marks a browse in flight; a failed
 /// navigation pops a toast while the previous listing stays usable.
 /// `channel` owns every browse and the eventual project mutation;
 /// `generation` numbers that picker's latest request. Both ride on replies,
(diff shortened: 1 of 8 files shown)


tonyfettes:
@Betterlol Can you please test multi-drive situation for me? My Windows machine has only one single C drive.

PR #949, opened 2026-08-20
deprecate Debug::to_repr extend declarations

tonyfettes:
Add `#deprecated` to all 578 `pub extend T with Debug::{to_repr}` declarations in the workspace (573 newly marked; 5 already carried a deprecated attribute and are left as-is). 168 source files + 82 regenerated `pkg.generated.mbti` interfaces (533 `#deprecated` lines, matching the repo's existing interface style).

## Verified locally

- `moon clean` then `moon check --target native --deny-warn` — clean, 0 warnings
- `moon check --target js` — 0 new warnings; only 4 pre-existing `implicit_impl_as_method` warnings at `desktop/frontend/composer/contract.mbt:109,124` (`RunStage`/`RunHeartbeat` derive `Debug` with no extend declaration), untouched by this PR and visible only under the local nightly toolchain
- `moon -C editor check --target all --warn-list +73 --deny-warn` — clean
- `moon info` regenerated, diff contains only `#deprecated` additions
- `moon fmt --check` — clean
- `moon test` — 1893/1894; the one failure (`internal/engine/ops.mbt:1368`, a session-read timeout test) passes in isolation and is unrelated to this attribute-only change

## Could not verify

- CI stable-toolchain JS row (`moon check --target js --deny-warn`): only nightly is available locally, and nightly surfaces the 4 pre-existing `RunStage`/`RunHeartbeat` warnings above. Stable is expected clean — the same lines are green on main CI.

Generated with SeekMoon

### DIFF
--- a/agent_explore/pkg.generated.mbti
+++ b/agent_explore/pkg.generated.mbti
@@ -29,6 +29,7 @@ pub fn Citation::equal(Self, Self) -> Bool
 pub fn Citation::from_json(Json, @json.JsonPath) -> Self raise @json.JsonDecodeError
 pub fn Citation::not_equal(Self, Self) -> Bool
 pub fn Citation::to_json(Self) -> Json
+#deprecated
 pub fn Citation::to_repr(Self) -> @debug.Repr
 
 pub struct ExploreReport {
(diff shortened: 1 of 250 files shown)


tonyfettes:
Well actually I don't think we need to add `#deprecated` warnings here because openseek has no downstream users.

PR #960, opened 2026-08-21
fix(desktop): integrate macOS titlebar controls

tonyfettes:
## Summary

- use Proton's titlebar overlay on macOS so SeekMoon paints the native window-control area
- reserve and align the traffic lights in both packaged and development HTML entries with a 36px header and an 80px leading inset
- add a packaging regression test for the native platform marker while leaving the browser console unchanged

## Validation

- `moon -C desktop check . --target native` (passes with 54 existing warnings)
- `moon -C desktop check frontend --target js` (passes with 70 existing warnings)
- `moon -C desktop test package/internal/packaging --target native` (16/16 passed)
- `moon -C desktop info`
- `moon -C desktop fmt`
- `git diff --check`
- visually verified through user-provided macOS development screenshots

### DIFF
--- a/desktop-dev.html
+++ b/desktop-dev.html
@@ -7,6 +7,20 @@
       content="width=device-width, initial-scale=1.0, interactive-widget=resizes-content"
     />
     <title>SeekMoon Development</title>
+    <script>
+      // The native host overlays macOS window controls on this development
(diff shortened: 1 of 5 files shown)


tonyfettes:
This sometimes crashes the application on exit.

PR #978, opened 2026-08-22
feat(desktop): stream live reasoning

bobzhang:
## Summary

- emit provider reasoning deltas unconditionally as transient protocol progress, with no public API switch or hidden serve flag
- show live reasoning in Desktop as a plain-text activity card, then settle it from the durable Assistant session event as Markdown
- show live reasoning in the TUI on its single dimmed Thinking activity row; the first answer delta replaces it with the normal Streaming preview
- rebuild resumed TUI history as the same Thought + Response projection from durable Assistant events
- keep AgentSession as the settled source of truth: deltas are only bounded live presentation state and never add durable rows
- preserve causal ordering between Desktop progress and durable follower commits so delayed reloads cannot duplicate or regress reasoning

## Architecture

The agent always reports available reasoning progress. Consumers decide whether and how to render it. Completed reasoning remains part of the existing durable Assistant event, so no second transcript source or reasoning lifecycle protocol is introduced.

## Testing

- `moon info && moon fmt`
- `just check`
- `just test`: 3,228 native tests, 3,090 JS tests, and 27 cram cases passed
- `just build`
- targeted TUI and Desktop suites rerun after the review fixes

## Review

- Fresh Codex review found stale checked-in JS interfaces; regenerated and committed them.
- A second fresh review found that resumed TUI history omitted durable reasoning; fixed with a single Assistant-to-items projection.
- A third fresh review found a remote reconnect suffix could be sealed beside the canonical reasoning; empty remote settlement now discards the provisional suffix.
- All completed review findings have regression coverage and are addressed.
- A final fresh Codex rerun was attempted after the last fix, but the CLI stopped before analysis because the account usage limit was reached (retry date reported as 2026-08-27). No final Codex sign-off is claimed.

### DIFF
--- a/agent/agent.mbt
+++ b/agent/agent.mbt
@@ -114,9 +114,8 @@ fn PlanCadence::record_plan_success(
 ///
 /// This owns a fresh runtime, task group, and tool registry for the duration of
 /// one turn. Use `run_turn_in_scope` when the caller needs to preserve runtime
-/// state or tool state across turns. `emit_reasoning_deltas` is transient: it
-/// adds provider reasoning fragments to the event stream without appending
-/// extra durable session items.
+/// state or tool state across turns. Provider reasoning fragments are emitted
+/// as transient events without appending extra durable session items.
 pub async fn run_turn_with_append(
   api_key~ : String,
   model~ : @deepseek.Model,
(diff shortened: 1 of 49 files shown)


bobzhang:
the rationle is that Agent may think 5 minutes, this PR tries to brinng reasoning back but stick with single source of truth. @tonyfettes

bobzhang:
Review update after simplification:

- Removed `emit_reasoning_deltas` from the public agent API and removed the hidden serve/Desktop plumbing; reasoning progress is unconditional and consumer-controlled.
- Added the TUI live Thinking → Streaming preview and durable Thought + Response reload projection.
- Addressed all three findings returned by fresh Codex review passes: JS interface drift, resumed-TUI reasoning loss, and partial remote reasoning after reconnect.
- Final gates pass on the stacked result. The final fresh Codex rerun was requested, but the CLI hit its account usage limit before analysis, so I am not claiming final sign-off.

PR #980, opened 2026-08-23
refactor(desktop): share one JSON object reader across protocol codecs

bobzhang:
## What

`desktop/internal/protocol` had **four** private wrappers over `Map[String, Json]` + `@json.JsonPath`, one per boundary:

| type | file |
| --- | --- |
| `AgentPayloadJsonObject` | `agent_payload_codec.mbt` |
| `CodexJsonObject` | `codex_commands.mbt` |
| `GitJsonObject` | `git_protocol_codec.mbt` |
| `TextSearchJsonObject` | `text_search_codec.mbt` |

They were byte-identical down to the decode error strings. Those strings are part of the wire contract, so four copies meant four chances to drift.

This collapses them into a single `JsonObject` in a new `json_object.mbt`, carrying the union of the readers the four boundaries needed (`required_json`, `optional_json`, `required_string`, `optional_string`, `required_bool`, `optional_bool`, `required_int`, `optional_int`, `required_array`, `required_string_array`).

`optional_thinking` stays in `agent_payload_codec.mbt` — it decodes `OpenSeekThinking` and belongs to that boundary, not to the generic reader.

## Behavior

None intended. Every reader keeps its exact error message and its exact null-versus-absent handling. Two readers were spelled two ways across the copies (a direct `match fields.get(...)` vs. `optional_json` + `guard`); the surviving spelling produces the same result and the same message for every input.

`pkg.generated.mbti` is unchanged — the wrapper was `priv`.

## Review

Roughly `+140 / -453`. The new file is the only thing to read closely; the four codec files are deletions plus a mechanical rename of the type at each call site.

## Verification

- `just test-moon` — 3221 native / 3078 JS, all passing

### DIFF
--- a/desktop/internal/protocol/agent_payload_codec.mbt
+++ b/desktop/internal/protocol/agent_payload_codec.mbt
@@ -4,64 +4,8 @@
 // at their exact path instead of silently changing the engine configuration.
 
 ///|
-priv struct AgentPayloadJsonObject {
-  fields : Map[String, Json]
-  path : @json.JsonPath
-}
-
-///|
-fn AgentPayloadJsonObject::decode(
-  json : Json,
-  path : @json.JsonPath,
-  expected : String,
-) -> AgentPayloadJsonObject raise @json.JsonDecodeError {
-  guard json is Object(fields) else {
-    raise JsonDecodeError((path, "expected " + expected + " object"))
-  }
-  { fields, path }
-}
-
-///|
-fn AgentPayloadJsonObject::required_json(
-  self : AgentPayloadJsonObject,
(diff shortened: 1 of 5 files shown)


bobzhang:

> The shared `JsonObject` helper preserves the behavior of the removed per-codec implementations. Targeted JS and native checks and all 44 protocol tests pass.

Codex independently re-ran `moon check` + `moon test` for `internal/protocol` on both `native` and `js` (44/44 passing on each) and confirmed `desktop/internal/protocol/pkg.generated.mbti` is byte-identical to `main`. No findings raised.

PR #981, opened 2026-08-23
refactor(eval): read CLI options through the shared internal/cli helpers

bobzhang:
## What

`internal/cli` already exists to hold the argparse accessors the command mains share. The two eval harness mains never got the memo — each carried its own byte-identical copy of `value` and `parse_positive_int`:

- `eval/prompt_task/cmd/main/config.mbt`
- `eval/file_edit/cmd/main/config.mbt`

Both copies match `internal/cli` character for character, including the failure strings (`missing parsed argument: ...`, `... must be a positive integer, got: ...`). Those strings are part of each command's user-facing contract, so three copies is three chances for them to drift.

This points both mains at `@cli` and deletes the local copies.

## Notes for the reviewer

- `flag` stays local to `prompt_task` — it is the only caller in the repo, so promoting it would just move the duplication problem.
- Both `moon.pkg` files swap `moonbitlang/core/string` (only needed by the deleted `parse_positive_int`) for `bobzhang/openseek/internal/cli`.
- `internal/cli/exports.mbt` carried a comment claiming `value` had no in-repo production consumer and existed only as a test seam. That is no longer true after this change, so the comment is dropped and the `cli.mbt` header now names the eval mains alongside `cmd/openseek` and `cmd/tui`.
- The rest of the diff is `value(` → `@cli.value(` and `parse_positive_int(` → `@cli.parse_positive_int(` at the call sites, plus `moon fmt` rewrapping the longer lines.

## Verification

- `moon test --target native eval/prompt_task/cmd/main eval/file_edit/cmd/main internal/cli` — 11/11

### DIFF
--- a/eval/file_edit/cmd/main/config.mbt
+++ b/eval/file_edit/cmd/main/config.mbt
@@ -1,44 +1,30 @@
 ///|
 fn config_from_matches(matches : @argparse.Matches) -> @harness.Config raise {
-  let runs = parse_positive_int(value(matches, "runs"), "--runs")
-  let min_successes = parse_positive_int(
-    value(matches, "min-successes"),
+  let runs = @cli.parse_positive_int(@cli.value(matches, "runs"), "--runs")
+  let min_successes = @cli.parse_positive_int(
+    @cli.value(matches, "min-successes"),
     "--min-successes",
   )
   guard min_successes <= runs else { fail("--min-successes must be <= --runs") }
   Config(
-    api_key=value(matches, "api-key"),
-    model=@deepseek.Model::parse(value(matches, "model")),
+    api_key=@cli.value(matches, "api-key"),
+    model=@deepseek.Model::parse(@cli.value(matches, "model")),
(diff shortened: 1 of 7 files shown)


bobzhang:

> The duplicated helpers are replaced consistently with equivalent shared implementations, and the required package dependencies are added. The changed packages compile and their targeted tests pass.

Codex also ran the whole `just test-moon` suite inside its sandbox and saw 36 failures. All of them are sandbox-environmental, not from this change: `deepseek/client/client_realworld_test.mbt` (`ResolveHostnameError` — no network), `desktop/internal/auth` (`@socket.TcpServer::new(): Operation not permitted`), and `cmd/openseek/internal/subtask/lifecycle_test.mbt` (git's `confstr()`/`DARWIN_USER_TEMP_DIR` warning corrupting `rev-parse` output). None of those packages is touched here. The same suite is 3221/3221 native and 3078/3078 JS outside the sandbox.

PR #982, opened 2026-08-23
refactor(subrun): share the capture-tool parse-and-validate step

bobzhang:
## What

`@agent_subrun.capture_tool` takes a `parse : (Json) -> Result[T, Array[String]]` and rejects-with-retry on `Err`. Every caller assembles that `parse` the same way — decode the report, run its contract validator, turn either failure into the problem list — and each wrote it out by hand:

- `agent_explore/submit.mbt`
- `agent_review/submit.mbt`
- `agent_worker/submit.mbt`

All three are byte-identical modulo the report type. `agent_worker` carried a fourth copy in `WorkerReport::parse_validated`, differing only in joining the problems into one string for the parent-side trust boundary.

This adds `@agent_subrun.validated` — placed next to `capture_tool`, where the reject-with-retry contract is already defined — and routes all four through it.

## Notes for the reviewer

- `WorkerReport::parse_validated` now delegates to `parse_submission` rather than repeating the decode-then-validate sequence. The child-side check and the parent-side re-check were always meant to be the same check; now they cannot drift. Its error shape is unchanged (`problems.join("; ")`).
- The three `parse_submission` functions are kept as named functions rather than inlined at the `capture_tool` call, so the existing tests that exercise them directly need no changes.
- `agent_subrun/pkg.generated.mbti` gains exactly one line: the new `validated`.

## Behavior

None intended. The same decoder and the same validator run in the same order, and problems reach the model in the same shape.

## Verification

- `moon test --target native agent_explore agent_review agent_worker agent_subrun` — 60/60

### DIFF
--- a/agent_explore/submit.mbt
+++ b/agent_explore/submit.mbt
@@ -24,17 +24,7 @@ fn submit_answer_tool(
 
 ///|
 fn parse_submission(args : Json) -> Result[ExploreReport, Array[String]] {
-  match ExploreReport::parse(args) {
-    Err(msg) => Err([msg])
-    Ok(report) => {
-      let problems = report.validate()
-      if problems.length() > 0 {
-        Err(problems)
-      } else {
-        Ok(report)
-      }
-    }
-  }
+  @agent_subrun.validated(ExploreReport::parse(args), ExploreReport::validate)
 }
 
 ///|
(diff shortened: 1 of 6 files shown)


bobzhang:

> The refactor preserves existing parse-and-validate behavior while centralizing it in a generic helper. Native checks and targeted tests for all affected packages pass.

Codex re-ran the four affected packages independently: `agent_subrun` 28/28, `agent_worker` 8/8, `agent_explore` 12/12, `agent_review` 12/12. No findings raised.

PR #983, opened 2026-08-23
refactor(openseek): build the turn's extra tools in one place

bobzhang:
## What

`main`'s one-shot turn (`run_task_turn`) and `serve`'s long-lived loop each built the same extra-tool list by hand:

1. `resolve_mcp_tools(...)` — the configured MCP servers' tools
2. `@agent_explore.definition(...)`, `review_tool_definition(...)`, `subtask_tool_definition(...)`
3. `@web_search.definition(...)`, but only when `web_search_api_key` returns a key the official DeepSeek endpoint accepts

The two expressions were byte-identical across 28 lines — two chances for one entry point to grow a tool surface the other does not have, which is exactly the kind of drift that is invisible until a user reports "it works in `serve` but not headless".

This extracts `build_extra_tools` into a new `cmd/openseek/tools.mbt` and calls it from both sites.

## Notes for the reviewer

- The parameters are exactly the locals each caller already had in scope: `exe`, `model`, `api_key`, `workspace_root`, `subrun_budget`, `review_context`, `api_url`, `child_session`. Nothing was hoisted or re-derived.
- `review_context` stays a closure parameter rather than a value, so the `review` tool still reads the LIVE standing goal at call time. `serve` reads the loop-updated session and `main` reads its `latest_session` ref — both keep their own closure.
- `budget` is still the single shared `SubrunBudget` per caller; the goal-met gate continues to bypass it (unchanged, in `gate.mbt`).

## Behavior

None intended. The same tools are registered in the same order for both entry points.

## Verification

- `moon test --target native cmd/openseek` — 70/70
- `moon cram test tests/cram` — 27/27

### DIFF
--- a/cmd/openseek/main.mbt
+++ b/cmd/openseek/main.mbt
@@ -277,38 +277,18 @@ async fn run_task_turn(
         latest_session.val.current_goal_baseline(),
       )
     }
-    let extra_tools = resolve_mcp_tools(matches, scope, workspace_root~) +
-      [
-        @agent_explore.definition(
-          self_exe=exe,
-          model~,
-          workspace_root~,
-          budget=subrun_budget,
-          api_url?,
-          child_session?,
-        ),
-        review_tool_definition(
-          self_exe=exe,
-          model~,
-          workspace_root~,
-          budget=subrun_budget,
-          context=review_context,
-          api_url?,
-          child_session?,
-        ),
-        subtask_tool_definition(
-          self_exe=exe,
-          model~,
-          workspace_root~,
-          budget=subrun_budget,
-          api_url?,
-          child_session?,
-        ),
-      ]
-    if web_search_api_key(model, api_key, api_url, matches)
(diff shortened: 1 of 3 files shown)


bobzhang:

> The refactor preserves the original tool definitions, ordering, arguments, shared state, and conditional web-search registration at both call sites. Focused tests and check/build gates passed.

It called out the live review closure specifically:

> The extracted helper preserves the original tool order, credentials, optional arguments, shared budget, task scope, and live review closure at both call sites. Focused tests and check/build gates passed; the full workspace test failures were sandbox/network-related and outside the changed code.

The sandbox failures it refers to are the usual environmental ones (no network for `deepseek/client/client_realworld_test.mbt`, no `TcpServer` bind for `desktop/internal/auth`, and git's `DARWIN_USER_TEMP_DIR` warning corrupting `rev-parse` in the subtask lifecycle tests). None of those packages is touched here.

PR #994, opened 2026-08-23
feat(desktop): reach a subagent's own transcript, and keep it live

bobzhang:
A sub-run's tool card holds the subagent's **report**. The work behind that report is a whole transcript of its own, written to `<parent>-sr-N`. Until now the only way to reach it was: read the `sr-N` out of the brief, find the parent in the sidebar, expand its twisty, match the ordinal by eye, click — which also navigates away from the transcript you were reading.

The viewer already had this link (`viz/view.mbt:2151` `subrun_chip`). The desktop did not.

## What changed

**The chip** — `desktop/frontend/transcript/component/view.mbt`

The card leads with a button that opens the child session. It sits at the top of the card **body**, not in the summary: the common shape for a sub-run is a step whose single tool call is inlined (`inline_tool_call_view`), which has no summary of its own — and a button inside a `<summary>` would have to fight that element's toggle activation. Leading the body puts it in the same place in both shapes, with no event-propagation subtleties at all.

Gated to OpenSeek transcripts (`TranscriptRenderer::subrun_link`). Codex writes no `-sr-N` records, so a brief that coincidentally reads like a sub-run must not offer a link to a session that cannot exist. An unsaved conversation has no parent id to hang a child off, and `subrun_child_session` rejects an empty parent.

**The parser** — `desktop/frontend/transcript/sessions.mbt`

`subrun_child_session(tool_name, brief, parent~)`, placed beside its exact inverse `subrun_label`. The contract is the brief `agent_explore`/`agent_review` stamp: `"<tool> sr-N (…)"`. Briefs are display-only — the model never sees one — so reading it here cannot change what a run did, and a brief that doesn't match simply yields no link.

**Live refresh** — `desktop/frontend/update.mbt`

Opening that transcript mid-run used to strand it. The host manages no writer for a child (it runs in a process the *parent engine* spawned), so no follower broadcasts its commits — `follower.mbt:16` states the invariant — and `session.load` answered it with a one-shot `read_document`. The transcript sat frozen at whatever the child had written when it was opened, which is indistinguishable from a stalled subagent.

The `SubrunProgress` push that already drives the dock lane now also drives a catch-up read of the child, but only when that child is the session actually on screen. `active_session` alone is not "visible" — a reader on Skills or Settings has one too — so the guard is `screen is Chat && channel == focused && session == active_session`.

## Two details that are load-bearing

**`ensure_transcript_reload_after_current`, not `begin_transcript_reload`.** The latter is single-flight and answers a signal arriving mid-read with `None`. That's right for a user re-selecting a row and wrong here: the probe publishes *only when the child advanced*, so a dropped signal can be the last one ever sent — a child that then finishes would strand the transcript one step short with nothing left to correct it. The regression test for this was verified to fail against the naive call before being kept.

**A background refresh, never `activate_on_success`.** The foreground path (`session_snapshot_command(..., true)`) pulls the UI to Chat, calls `Model::activate` — pruning other conversations and clearing their unseen-finish marks — and remounts the transcript component. The remount is exactly what would discard the reader's scroll position and open disclosures (`component.mbt:156` preserves `pinned` only when source/channel/session are unchanged and the component did not remount). None of that belongs on a progress tick.

The lane update and the child refresh are now **independent** rather than one gating the other: previously the handler returned early if the parent conversation wasn't found, and returned `@cmd.none` in every branch.

## Notes for review

- **`viz` keeps its own copy of the brief parser.** Sharing it would mean giving `viz` a dependency on `openseek_protocol` it does not otherwise have — a worse trade than one duplicated twelve-line contract. Flagging it rather than hiding it.
- **New `Msg::OpenSessionHere(String)`** — a transcript link has only a session id, and the component is constructed once, before any model exists to read a channel from. It delegates straight to `OpenSession(model.focused, id)` rather than duplicating that path.
- **Not addressed here:** the refresh re-reads the whole child record per change (`read_document`), rather than incrementally as a real follower would. That's the deliberate trade — it reuses machinery that already exists and is bounded by "someone is actively watching this one child." A follower for sub-run children would be the incremental fix, at the cost of loosening the writer-ownership invariant.

## Verification

`moon check --target js` clean, `moon fmt --check` clean, full suite **3120/3120**. Generated `.mbti` synced for both changed packages.

Nine new tests: parser accept/reject cases, chip present for a sub-run card, absent for Codex and for an ordinary tool, and four handler cases — refresh when viewed, no refresh when not, coalescing during an in-flight read, and refresh surviving a missing parent.

Not verified against running pixels; the app was not launched for this change.

https://claude.ai/code/session_01LebjbsShsSNTQg25Vi4tC7

### DIFF
--- a/desktop/frontend/boot.mbt
+++ b/desktop/frontend/boot.mbt
@@ -506,6 +506,7 @@ pub fn boot() -> Unit {
           annotation=target.annotation,
         )
       }),
+      open_session: emit.map(session => OpenSessionHere(session)),
     })
     let composer_input = state.map(model => model.composer_input())
     let composer_html = @composer.component(
--- a/desktop/frontend/model.mbt
+++ b/desktop/frontend/model.mbt
@@ -1725,6 +1725,11 @@ priv enum Msg {
   // Open a conversation on one device's explicit channel, selecting that
   // channel first when it differs from the current one.
   OpenSession(@interop.ChannelId, String)
+  // Open a conversation on the channel already focused. A transcript link has
+  // only a session id to go on — the child session a sub-run tool card names
+  // belongs to the same device as the transcript that names it — and the
+  // component is built once, before any model exists to read a channel from.
+  OpenSessionHere(String)
   // Open a row from the archived index. Its workspace is part of the row's
   // durable placement and selects the archived twin without restoring it.
   OpenArchivedSession(
--- a/desktop/frontend/transcript/component/component.mbt
+++ b/desktop/frontend/transcript/component/component.mbt
@@ -102,6 +102,9 @@ pub fn Input::Input(
 pub(all) struct Actions {
   open : @cmd.Emit[String]
   jump : @cmd.Emit[@mention_context.Target]
+  /// Open another session's transcript. Used by a sub-run tool card to reach
+  /// the child transcript it names.
+  open_session : @cmd.Emit[String]
 }
 
 ///|
@@ -265,6 +268,7 @@ pub fn component(
       overview_emit=emit.map(msg => Overview(msg)),
       path => (actions.open)(path),
       target => (actions.jump)(target),
+      session => (actions.open_session)(session),
       on_jump_latest=emit(JumpToLatest),
     )
   })
--- a/desktop/frontend/transcript/component/pkg.generated.mbti
+++ b/desktop/frontend/transcript/component/pkg.generated.mbti
@@ -20,6 +20,7 @@ pub fn component(@rabbita.Val[Input], Actions) -> @rabbita.Val[@html.Html]
 pub(all) struct Actions {
   open : @cmd.Emit[String]
   jump : @cmd.Emit[@mention_context.Target]
+  open_session : @cmd.Emit[String]
 }
 
 type Input derive(Eq)
--- a/desktop/frontend/transcript/component/rendering.mbt
+++ b/desktop/frontend/transcript/component/rendering.mbt
@@ -104,6 +104,20 @@ priv struct TranscriptRenderer {
   input : Input
   on_open : (String) -> @cmd.Cmd
   on_jump : (@mention_context.Target) -> @cmd.Cmd
+  on_open_session : (String) -> @cmd.Cmd
+}
+
+///|
+/// The context a sub-run tool card needs to link to its child transcript, or
+/// `None` when this transcript can hold no sub-run.
+///
+/// Only OpenSeek writes `<parent>-sr-N` child records; a Codex or legacy
+/// transcript with a coincidentally-shaped brief must not be given a link to a
+/// session that does not exist. An unsaved conversation has no parent id to
+/// hang a child off, and `subrun_child_session` rejects the empty parent.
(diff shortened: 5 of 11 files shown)


bobzhang:
## Codex review applied — both High findings fixed

Ran `codex exec` against this branch. It found two real defects, both confirmed against the code and now fixed in 0db9efcd.

### 1. The child could still permanently miss its final updates (High)

The refresh only fired on a probe advance — but the host cancels the probe at the `subrun_finished` bracket with no final poll (`internal/engine/engine.mbt:1753`), and `follow_progress` sleeps *before* it reads. Everything the child wrote in its last ≤1.5s reached nobody, and that window is exactly where its report lands.

So the original PR left a reader watching the child transcript stop just short of the end — the specific failure the refresh existed to prevent, surviving over the most valuable content. The finish is the only terminal boundary a child transcript has, and it now drives the same catch-up. The shared guard moved into `refresh_visible_subrun_child`.

### 2. Ordinary `subtask` launches never got a chip (High)

`subtask_launch_many` discards the per-worker briefs and reports one brief leading with a slice **count**, each worker's `sr-N` nested inside (`cmd/openseek/subtask_tool.mbt:356`). Token 1 is a number, so even a single-slice launch linked nothing, and parallel launches lost several ids at once.

Rather than special-case that shape, the parser now scans every token and returns **all** children in first-appearance order; the card renders a chip each. That's a simpler rule than the one it replaces — "every `sr-N` the brief names" instead of "token 1 must be `sr-N`" — and it covers the parallel case for free.

Also folded in: ordinals bounded at nine digits to match `@protocol.split_child_session_id`, so the documented inverse relationship is actually true rather than true-until-a-billion-subruns.

### Test gap, also fair

The positive handler test asserted only that the conversation reached `Reloading` — which an implementation that parked there without ever issuing a read would satisfy too. `Cmd` is opaque, so it now asserts the minted generation and drives the round trip: the fenced `SessionRefreshed` is accepted and adopted. New coverage for the subtask launch shape and the terminal refresh.

Both new behaviors were verified to **fail** when reverted before the tests were kept:

```
("a sub-run's finish refreshes the child transcript one last time") FAILED: `0 != 1`
Total tests: 429, passed: 428, failed: 1
```

### Knowingly not fixed here

- **Archived transcripts show a chip that silently no-ops** (Medium). Archived records project as `Source::OpenSeek`, so they get a chip; the click resolves against the live index and stops at the placement guard. Needs an archived flag threaded into the component's `Input`.
- **Retry exhaustion drops the queued follow-up with no visible error** (Medium). A later push restarts from `ReloadFailed`, so it isn't a permanent freeze in the common case — and fixing #1 removes the scenario where the lost signal was the last one.
- **`OpenSessionHere` reads `model.focused` at delivery** (Low), and a queued follow-up doesn't re-check visibility (Low).

### What the review cleared

Generation fencing is sound on this path; no success-path spin, double-load, or stale-generation overwrite. `activate_on_success=false` doesn't weaken the checks. Nested parents work, `capitalize` doesn't affect parsing, and every render path preserves the link.

Full suite **3123/3123**, `moon check` and `moon fmt --check` clean.

PR #996, opened 2026-08-23
feat(desktop): mark the turns that hit errors on the overview rail

bobzhang:
The rail is the only view of a whole conversation, and the one thing a reader scrolls a long conversation looking for is where it went wrong. Every tick looked the same, so finding the failure meant scrolling past every turn that did not have one — while the answer was already in the rows the rail is built from, since each tool row carries `is_error`.

An `Entry` now reports whether any tool in its turn came back an error, and its tick draws in the danger colour. `viz`'s step scrubber has flagged failing steps this way for a while (`steps_with_errors`); this fits the idea to the structure desktop's rail already has — one stop per **user turn**, not per model step, so no new grouping is invented.

### Three decisions worth stating

- **The turn's failures, not the run's health.** A turn that hit an error and recovered still marks. "The agent hit something here and worked around it" is the same place a reader wants to land.
- **The scan covers the whole turn**, while the existing reply search stops at the first answer. A tool that failed *after* the answer still failed in this turn, so the two loops deliberately have different stopping rules — now one loop with that stated.
- **The mark stays short and only changes colour**, so a failed tick is never mistaken for the active one. The rule is declared before hover and active, which carry equal specificity: those answer "where am I" and outrank this rule's "where should I look".

Colour is not the only carrier — the tick's `aria-label` gains `(had errors)`, so someone arrowing the rail hears which turns hit errors.

### Verification

`moon check` clean on js and native with `--deny-warn`; `moon test` 3130 js / 3246 native, all passing; `moon fmt --check` clean. New tests cover the projection (recovered turn marks, clean turn does not, the scan stops at the next prompt) and the rendered markup.

https://claude.ai/code/session_01682JUCDbF278qVApyjcbBQ

### DIFF
--- a/desktop/frontend/transcript_overview/pkg.generated.mbti
+++ b/desktop/frontend/transcript_overview/pkg.generated.mbti
@@ -23,6 +23,7 @@ pub(all) struct Entry {
   text : String
   ts : Double?
   reply : String?
+  failed : Bool
 } derive(Eq, @debug.Debug)
 
 pub(all) struct Hover {
--- a/desktop/frontend/transcript_overview/transcript_overview_test.mbt
+++ b/desktop/frontend/transcript_overview/transcript_overview_test.mbt
@@ -45,6 +45,63 @@ test "entries pick user turns and pair each with its first answer" {
 }
 
 ///|
+test "a turn marks failed for any tool error it contains, answered or not" {
+  let tool = is_error => {
+    @transcript.ItemKind::Tool(
+      call_id="c1",
+      name="shell",
+      arguments=None,
+      has_result=true,
+      is_error~,
+      brief=None,
+    )
+  }
+  let rows = @transcript_overview.entries(
+    [
+      item(User, "first question", sequence=1),
+      item(tool(true), "boom", sequence=2),
+      item(Assistant(is_danger=false), "recovered", sequence=3),
+      // A failure after the answer is still this turn's failure: the reply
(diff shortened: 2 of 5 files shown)


bobzhang:
Review complete — no blocking findings.

I rebased the three PR commits onto current `main` (`7d03a873`). The rebase applied cleanly with no textual conflicts, and the updated head is `4cfe43d4`.

What it does:

- Projects a `failed` flag onto each overview entry when any tool call in that user turn has `is_error=true`, including turns that later recovered.
- Draws those turns in the danger colour while preserving the wider active/hovered shape.
- Adds `(had errors)` to the tick's accessible label and shows `had errors` in the hover preview.
- Raises the preview-space budget so the extra row does not clip near the bottom of the transcript.

Verification:

- Focused `transcript_overview` tests: 13/13 passed.
- Full JS tests: 3,141/3,141 passed.
- `moon info`, `moon fmt`, native+JS checks, and native+JS builds passed locally.
- Rebased GitHub CI: native and JS checks passed; GitGuardian passed.
- Local `just test`'s native phase had 27 unrelated `run_moonbit`/sandbox-policy failures because the installed `moonrun 0.1.20260819` expects `spawn` where this checkout emits `allow`; the PR does not touch those packages, and the native GitHub job passed.

Visual fixture using the production transcript tokens and rail/preview CSS:

Failed turn highlighted on the conversation overview rail

PR #1000, opened 2026-08-24
fix(desktop): handle file watcher failures

tonyfettes:
## Summary

- retry non-cancellation `EINTR` interruptions from the file watcher without declaring the watcher stopped
- surface accepted current-generation watcher failures through the existing error toast rail
- keep watcher lifecycle state for retry-loop prevention while removing the persistent file-panel error row
- fence stale and repeated watcher failures and cover the toast and rendering behavior with tests

## Root cause

On macOS the async filesystem watcher can expose an interrupted `kevent` probe as `OSError("@fs.Watcher::wait(): Interrupted system call")`. OpenSeek treated every non-cancellation wait error as a permanent watcher failure even though `EINTR` consumes no filesystem event and is safe to retry.

## Validation

- `moon check --target native desktop/internal/host`
- `moon test --target native desktop/internal/host` (98 passed)
- `moon check --target js desktop/frontend`
- `moon test --target js desktop/frontend` (425 passed)
- `moon info`
- `moon fmt`
- `git diff --check`

### DIFF
--- a/desktop/frontend/fileeditor/explorer_search_wbtest.mbt
+++ b/desktop/frontend/fileeditor/explorer_search_wbtest.mbt
@@ -51,6 +51,17 @@ test "Search inventory hides unrelated file watcher errors" {
 }
 
 ///|
+test "file watcher errors stay off the Files inventory" {
+  let html = render_explorer_for_test({
+    ..owned_state(),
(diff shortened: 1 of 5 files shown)


tonyfettes:
@Guest0x0 Might be better to retry EINTR in async.

PR #1033, opened 2026-08-24
fix(desktop): contain narrow composer content

bobzhang:
## Summary

- let the composer card and its grid children shrink with a narrow conversation pane
- keep placeholder text wrapped inside the textarea border
- constrain the Git status button to the width assigned by the toolbar flex layout

## Verification

- browser layout check at a 260px conversation-pane width: input, textarea, toolbar, status group, and Git button all remain within the card; no horizontal scroll overflow
- moon test desktop/frontend/composer --target js (21 passed)
- just check
- just test (3249 native, 3142 JS, and 27 cram tests passed)
- just build
- moon info && moon fmt

A fixed-state snapshot is attached in a follow-up comment.


bobzhang:
## Verification snapshot

Fixed at a 260px conversation-pane width. The textarea placeholder wraps inside the focus border, and the compact toolbar controls remain within the card.

Fixed narrow composer

PR #1041, opened 2026-08-25
fix(tests): isolate cram session environment

bobzhang:
## Summary

- clear inherited OPENSEEK_SESSION and OPENSEEK_SESSION_ROOT in session-writing CLI cram fixtures
- keep fixture prompts in temporary stores when tests run from a live OpenSeek agent
- document why the isolation is required

## Validation

- ran tests/cram/cli.md under a fake parent session: 20/20 passed and the parent store remained absent
- moon info
- moon fmt


bobzhang:
Superseded by #1046, #1047, and #1050. Session identity and placement no longer come from these inherited environment variables: the CLI no longer binds the session selector, Desktop passes the root explicitly, and #1050 removes the legacy root identifier repository-wide. Keeping these fixture unsets would now be a no-op and would reintroduce the retired names into git grep.

PR #1046, opened 2026-08-25
fix(session): pass session id explicitly

bobzhang:
## Summary

- remove the `OPENSEEK_SESSION` CLI environment binding and all remaining references to that interface
- pass the session id to TUI and desktop engine children with `--session`
- document the explicit custom-engine argument contract

This is the first of two stacked changes. `OPENSEEK_SESSION_ROOT` remains unchanged here and moves to argv in #1047.

## Security rationale

A session id selects a writable durable conversation. Making it argv-only keeps that capability attached to the intended engine launch instead of ambient process state. There is intentionally no compatibility shim: `git grep -w OPENSEEK_SESSION` returns no matches on this branch.

## Validation

- cmd/openseek 80/80
- cmd/tui 86/86
- desktop/internal/engine 107/107
- `git grep -w OPENSEEK_SESSION` returns no matches

### DIFF
--- a/cmd/openseek/main.mbt
+++ b/cmd/openseek/main.mbt
@@ -1346,7 +1346,6 @@ test "cli parses env backed options and task" {
     "OPENSEEK_MAX_STEPS": "120",
     "OPENSEEK_SYSTEM_PROMPT_FILE": "/tmp/prompt-a.md",
     "OPENSEEK_SYSTEM_PROMPT_ADDENDUM_FILE": "/tmp/prompt-b.md",
-    "OPENSEEK_SESSION": "demo",
     "OPENSEEK_SESSION_ROOT": "/tmp/openseek-sessions",
   })
   let (model, _, api_url) = agent_settings(parsed)
@@ -1361,7 +1360,7 @@ test "cli parses env backed options and task" {
     @cli.value(parsed, "system-prompt-addendum-file"),
     "/tmp/prompt-b.md",
   )
-  assert_true(parsed is { values: { "session": ["demo", ..], .. }, .. })
+  assert_false(parsed is { values: { "session": [_, ..], .. }, .. })
   assert_eq(session_root(parsed), "/tmp/openseek-sessions")
   assert_eq(task_text(parsed), "write MoonBit")
 }
--- a/cmd/tui/loop.mbt
+++ b/cmd/tui/loop.mbt
@@ -149,7 +149,6 @@ fn engine_extra_env(config : AppConfig) -> Map[String, String] {
     "OPENSEEK_API_URL": config.api_url.unwrap_or(""),
     "OPENSEEK_MODEL": config.model.to_string(),
     "OPENSEEK_THINKING": config.thinking.to_string(),
-    "OPENSEEK_SESSION": config.session_id.value(),
     "OPENSEEK_SESSION_ROOT": config.session_root,
     "OPENSEEK_DIR": config.cwd,
   }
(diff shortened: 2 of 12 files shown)


cmd/openseek/main.mbt
```diff
@@ -25,6 +25,10 @@ async fn dispatch() -> Unit {
   let argv = @env.args()[1:]
   let env = @env.get_env_vars()
   let matches = root_command().parse(argv~, env~)
+  // A durable session is an explicit argv capability, never ambient process
+  // state. Drop a legacy exported value before the UI or engine can launch any
+  // child tools; the parsed `--session` value already lives in `matches`.
+  @env.unset_env_var("OPENSEEK_SESSION")
```
bobzhang: is this still needed given it is removed

cmd/openseek/main.mbt
```diff
@@ -1361,11 +1364,20 @@ test "cli parses env backed options and task" {
...
+  assert_false(parsed is { values: { "session": [_, ..], .. }, .. })
   assert_eq(session_root(parsed), "/tmp/openseek-sessions")
   assert_eq(task_text(parsed), "write MoonBit")
 }
 
+///|
+test "cli ignores the legacy session environment variable" {
+  let parsed = run_matches(argv=["write", "MoonBit"], env={
+    "DEEPSEEK": "test-key",
+    "OPENSEEK_SESSION": "parent-session",
```
bobzhang: this test needed?

cmd/tui/loop.mbt
```diff
@@ -171,9 +170,10 @@ fn engine_env(
   config : AppConfig,
   inherited : Map[String, String],
 ) -> Map[String, String] {
-  // The merge overwrites every OpenSeek setting — including the always-present
-  // session id — so inherited session vars can never shadow this TUI's config.
+  // Session identity travels in argv. Remove any legacy inherited value before
+  // spawning either the bundled engine or a custom one.
   let env = inherited.copy()
+  env.remove("OPENSEEK_SESSION")
```
bobzhang: same

cmd/tui/loop.mbt
```diff
@@ -1022,19 +1020,16 @@ fn compact_session_count(cmds : Array[Cmd]) -> Int {
...
-    fail("missing session id env")
-  }
   guard env.get("OPENSEEK_SESSION_ROOT") is Some(session_root) else {
     fail("missing session root env")
   }
   guard env.get("OPENSEEK_DIR") is Some(dir) else {
     fail("missing workspace dir env")
   }
-  assert_eq(session_id, "test")
+  assert_true(env.get("OPENSEEK_SESSION") is None)
```
bobzhang: same

cmd/openseek/main.mbt
bobzhang: Agreed; removed the compatibility unset. The identifier now disappears entirely with `git grep -w OPENSEEK_SESSION`.

cmd/openseek/main.mbt
bobzhang: Removed. There is no legacy-environment behavior left to test.

cmd/tui/loop.mbt
bobzhang: Removed the inherited-value cleanup as well; the old interface no longer appears in the codebase.

cmd/tui/loop.mbt
bobzhang: Removed. The remaining test covers the explicit argv contract without referring to the deleted environment variable.

bobzhang:
✅ Fresh Codex CLI sign-off on final commit `a47a1db8` (Codex CLI 0.147.0, new ephemeral read-only session `01a03729-5a07-7433-9a3c-2c5b0b5b90ae`): no actionable findings.

The earlier independent review found that split `--session <id>` argv broke valid IDs beginning with `-`. This was addressed by emitting `--session=<id>` in both TUI and Desktop launchers, updating the engine fixture, and adding leading-hyphen regression coverage.

Final validation for the addressed feedback: `just check`; TUI 86/86; Desktop engine 107/107; exact `git grep -w OPENSEEK_SESSION` remains empty.

PR #1047, opened 2026-08-25
fix(session): pass session root explicitly

bobzhang:
## Summary

- remove the `OPENSEEK_SESSION_ROOT` binding from the OpenSeek CLI
- pass the selected root to TUI and desktop engine children with `--session-root`
- scrub inherited session-root values before spawning bundled or custom engines
- update custom-engine docs, help snapshots, and engine fixture coverage

This is stacked on #1046; review only the second commit here.

## Scope

This removes `OPENSEEK_SESSION_ROOT` from the agent launch and descendant-process path. The trusted Desktop host still supports the same variable as an internal global-store configuration fallback, and the read-only visualizer keeps its compatibility input; neither value is forwarded to agent engines or tools.

## Security rationale

A session root determines where durable conversations can be written. Passing it as an engine argument keeps that write capability attached to the intended launch instead of exposing it as ambient state to tools and nested OpenSeek commands.

## Validation

- `just check`
- `just test` (native 3303/3303, JS 3209/3209, cram 28/28)
- `just build`
- targeted cmd/openseek 81/81, cmd/tui 86/86, desktop/internal/engine 107/107
- CLI/TUI cram with poisoned legacy session variables: 25/25; poison root remained empty

### DIFF
--- a/cmd/openseek/README.md
+++ b/cmd/openseek/README.md
@@ -53,8 +53,8 @@ exists.
 `--system-prompt-file` and `--system-prompt-addendum-file` can also be supplied
 with `OPENSEEK_SYSTEM_PROMPT_FILE` and
 `OPENSEEK_SYSTEM_PROMPT_ADDENDUM_FILE`. `--session` explicitly creates or
-resumes that session under `--session-root` / `OPENSEEK_SESSION_ROOT` (default
-`.openseek`). Relative session roots are resolved under `--dir`.
+resumes that session under `--session-root` (default `.openseek`). Relative
+session roots are resolved under `--dir`.
 
 Every run records a durable session: without `--session`, a generated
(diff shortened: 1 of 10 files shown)


bobzhang:
✅ Fresh Codex CLI sign-off on final commit `7c2990f7` (Codex CLI 0.147.0, new ephemeral read-only session `01a03729-599c-7e53-a200-0a12a7d632aa`): no actionable findings.

The review traced the changed behavior through CLI/TUI spawning, Desktop workspace-derived engine configuration, session queries, and subrun propagation. Session roots are removed from agent child environments and forwarded explicitly as `--session-root=<root>`.

Final validation: `just check`; `just test` — Native 3303/3303, JS 3209/3209, cram 28/28; CLI 81/81; TUI 86/86; Desktop engine 107/107; attached-value parsing also verified with values beginning with `-`.

PR #1069, opened 2026-08-26
fix(deepseek): preserve tool-call reasoning

bobzhang:
Summary:
- replay DeepSeek reasoning_content on historical assistant tool-call messages
- keep ordinary DeepSeek replies and the existing Kimi policy unchanged
- cover both message serialization and full request encoding

Validation:
- moon test deepseek/json_encode.mbt deepseek/encode_chat_request_test.mbt --target native
- just check
- just test
- just build


tonyfettes:
CI failed because pre-release toolchain bans the `#valtype` usage in `moondiff` package. Verified locally with stable toolchain.

PR #1085, opened 2026-08-27
refactor(deepseek): always replay assistant reasoning_content

bobzhang:
## Summary

Simplifies message encoding by always including assistant `reasoning_content` when replaying conversation history, instead of gating it on model and request-level tools. `ChatMessage::to_json_for_model(model, tools~)` collapses into a plain `impl ToJson for ChatMessage`, and `encode_chat_request` no longer threads `tools` into per-message encoding (net −68 lines).

## What the provider docs say

Checked each provider's docs on replaying `reasoning_content` in input history:

| Provider | Replaying assistant `reasoning_content` in history |
|---|---|
| DeepSeek | **Required when tools are enabled** — "the intermediate assistant's reasoning_content must … be passed back to the API in all subsequent requests — even for turns where the model did not perform a tool call." Without tools: "If passed to the API in subsequent turns, it will be ignored" (no 400, unlike the old R1 endpoint). [Thinking mode guide](https://api-docs.deepseek.com/guides/thinking_mode#thinking-mode-with-tool-calls) |
| Kimi K2.7 Code | **Always required** for every historical assistant message. [Kimi guide](https://platform.kimi.ai/docs/guide/use-kimi-k2-thinking-model#using-the-kimi-k2-7-code-model) |
| GLM (Z.ai) | **Always required**: "Remember to return the historical reasoning_content to keep the reasoning coherent" / "you must return the complete, unmodified reasoning_content back to the API." [Thinking mode guide](https://docs.z.ai/guides/capabilities/thinking-mode.md) |

Since every provider either requires or ignores replayed reasoning, per-message encoding is model-independent — the strictest behavior (always send) is safe everywhere. This also means a future `Glm(...)` model variant would not touch the message encoder; GLM-specific handling (`thinking: {type: enabled}` without `reasoning_effort`, plus its `clear_thinking` knob) belongs in the request-level `encode_chat_request` match where Kimi is already special-cased.

## Behavior change

Non-tool DeepSeek requests now include `reasoning_content` on historical assistant messages. Per the DeepSeek docs this is ignored server-side; the only cost is the extra bytes in the payload.

## Testing

- `moon check` clean, `moon fmt` applied, `moon info` produced no interface change (trait impls aren't listed in `.mbti`).
- All 2060 tests pass. The now-redundant per-model tests merged: DeepSeek/Kimi tool-call reasoning tests became one "assistant tool calls preserve reasoning content"; "deepseek reasoning follows tool availability" became "assistant reasoning content is always replayed".

https://claude.ai/code/session_01PZ9BYUZxaL9usinbe4Vw9a

### DIFF
--- a/deepseek/json_encode.mbt
+++ b/deepseek/json_encode.mbt
@@ -1,36 +1,16 @@
 ///|
-/// Keep assistant reasoning_content when the provider requires it.
-///
-/// DeepSeek requires every historical assistant reasoning trace to be sent
-/// back when tools are enabled. Kimi K2.7 Code always requires the reasoning
-/// from every historical assistant message to be preserved.
-///
-/// See: https://api-docs.deepseek.com/guides/thinking_mode#thinking-mode-with-tool-calls
-/// See: https://platform.kimi.ai/docs/guide/use-kimi-k2-thinking-model#using-the-kimi-k2-7-code-model
-fn ChatMessage::reasoning_json_fields(
-  message : ChatMessage,
-  model : Model,
-  preserve_deepseek_reasoning : Bool,
-) -> Array[(String, Json)] {
-  match (model, message.role, message.reasoning_content) {
-    (Deepseek(_), Assistant, Some(reasoning)) if preserve_deepseek_reasoning =>
-      [("reasoning_content", Json::string(reasoning))]
-    (Kimi(_), Assistant, Some(reasoning)) =>
-      [("reasoning_content", Json::string(reasoning))]
-    _ => []
-  }
-}
-
-///|
-/// Encodes a `ChatMessage` for a selected chat-completions model.
+/// Encodes a `ChatMessage` as a chat-completions `messages[]` entry.
 ///
 /// Assistant messages whose `content` is empty but carry tool calls are
 /// encoded with JSON `content: null`, as the API requires.
-fn ChatMessage::to_json_for_model(
-  message : ChatMessage,
-  model : Model,
-  preserve_deepseek_reasoning? : Bool = false,
-) -> Json {
+///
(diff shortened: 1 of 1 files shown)


bobzhang:
cc @tonyfettes

bobzhang:
## Self-review (multi-agent, adversarially verified)

Ran /code-review on this PR; 8 findings survived verification. Triage:

**Fixed in 0129d828:**
- Stale comments still framing reasoning replay as Kimi-only (`agent/auto_compact_test.mbt`, `agent/turn_loop.mbt`) and the `deepseek/README.mbt.md` wording that implied DeepSeek requests drop the field.
- The three manual array comprehensions (`messages`, `tool_calls`, `tools`) now use the `ToJson` impls directly (`messages.to_json()` etc.) instead of re-implementing `Array[X : ToJson]`.
- Missing coverage: added a test pinning that non-assistant messages drop `reasoning_content` (the `role is Assistant` guard was otherwise unpinned), and a request-level test pinning the changed contract itself — no-tools DeepSeek requests replay assistant reasoning.

**Accepted as deliberate trade-offs (not fixed):**
- The compaction/checkpoint path is now the one production no-tools DeepSeek call that sends reasoning + `thinking: disabled`; the docs say the field is ignored, and the new request-level test pins the shape. If DeepSeek ever deviates, compaction fails soft (session returned uncompacted).
- Checkpoint requests carry the pressured history's reasoning traces, so compaction prompts got heavier. Known cost of the simplification; tools-enabled requests (the dominant path) already paid it.
- `Some("")` serializes as an empty `reasoning_content` — pre-existing behavior on the tools-enabled path, unchanged here.
- Message encoding no longer has a per-model seam; reintroducing one is mechanical if a future provider rejects the field (GLM, per its docs, requires it instead).

2062/2062 tests pass.

PR #1087, opened 2026-08-27
perf(desktop): avoid transcript resize VDOM diffs

tonyfettes:
## Summary

- split transcript content changes from geometry-only resize handling
- keep pinned transcripts at the tail without scanning active turns on each resize
- avoid same-position scrollTop writes and skip duplicate scans from programmatic pinned scroll events
- notify Rabbita only when an unpinned resize changes the visible overview turn

## Why

The transcript ResizeObserver previously emitted ContentRendered for every size notification. Dragging the transcript/editor splitter therefore scheduled Rabbita reconciliation even when transcript content had not changed.

Pinned transcripts also had a bottom-only hot path: every resize wrote scrollTop and the resulting captured scroll event scanned the long transcript again. The pinned path now writes only after a real gap exceeds the existing 4px tolerance and otherwise returns without an active-turn scan.

## Validation

- moon test desktop/frontend/transcript/component --target js (50 passed)
- just check
- just build
- just test on the initial split (native 3323 passed, JS 3246 passed, cram 28 passed)

## Remaining cost

Changing pane width still requires the browser to reflow wrapped transcript content. Unpinned resize callbacks also read turn geometry so the overview highlight can follow the reading position.

### DIFF
--- a/desktop/frontend/transcript/component/component.mbt
+++ b/desktop/frontend/transcript/component/component.mbt
@@ -439,14 +439,21 @@ fn transcript_sub_loader(
           element.get_property("id").to_option() == Some("transcript") else {
           return
         }
+        let was_pinned = transcript_pinned_watch.val
         let client_height = element.get_client_height()
         let pinned = element.get_scroll_height() -
           element.get_scroll_top() -
           client_height <=
           4.0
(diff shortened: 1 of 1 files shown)


tonyfettes:
Actually is slower.

PR #1089, opened 2026-08-27
chore: format with moon 0.1.20260827

tonyfettes:
## Summary

- The stable channel moved to `moon 0.1.20260827` / `moonc v0.10.11` today; its formatter writes a trailing comma inside multi-line struct literals and spreads.
- CI installs `moonbit-version: latest`, so every branch pushed after the bump fails `moon fmt --check` on files it never touched — #1083 hit it first, and `main` will on its next run.
- This is `moon fmt` over the whole workspace with that toolchain and nothing else: 581 files, formatting only. `moon info` under the same toolchain changes no `.mbti`.

## Test plan

- [x] `moon fmt --check` clean under `moon 0.1.20260827`
- [x] `moon info` leaves every `.mbti` unchanged
- [ ] CI (stable, native) and (stable, js) green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1083, opened 2026-08-27
feat(prompt): move mbtx onto moonbitlang/async/shell

tonyfettes:
## Summary

- The default prompt now owns the `mbtx` tutorial and teaches the shell-free `moonbitlang/async/shell` API (`@shell.glob`, `.output()` read through `Output`'s accessor methods, `each_line` for streaming `moon check --output-json`); the tool description keeps only the enforced contract.
- `moonbitlang/async` 0.21.0 → 0.21.1.
- Mirror tests in `agent_tool/run_moonbit` run the prompt's example shapes through the real `mbtx` runner (`Output` accessors, streamed diagnostics, `@shell.glob`).
- The explore, review, audit, and worker prompts carry the same Running Commands block (checklist, `.output()` example with `Output` accessors, literal argument vectors, `@fs.tmpdir(prefix=)`, coreutils table) ahead of each role's own narrowing, since the tool description no longer teaches the snippet shape; READMEs that still said `bobzhang/myshell` now say `moonbitlang/async/shell`.
- The system prompt now also states the mbtx checklist (separate imports, `async fn main`, async helpers, no `await`/`print`), the program allowlist, and the coreutils-to-MoonBit table right after it; the tool description keeps the allowlist plus refusal/escalation rules and no longer carries the table. Arrow-lambda rule with a checked example; `pkgtype(kind: "executable")` in the moon.pkg example.

## Eval (deepseek-v4-flash, 10 trials per arm)

Arms are separately built binaries: **baseline** = `main` (`bobzhang/myshell` examples in the tool description), **draft** = first version of this prompt, **shipped** = this PR.

| task | metric | baseline | draft | shipped |
|---|---|---:|---:|---:|
| corpus audit (rg through `@shell.Cmd`, 6 graded sections) | audits fully correct | 10/10 | 9/10 | 10/10 |
| | rg calls whose first diagnostic sits on the `Cmd` line | 4/82 | 7/81 | **0/57** |
| | mean steps | 12.1 | 10.5 | 9.9 |
| TOML parser CLI (harness-validated) | validated trials | 10/10 | 9/10\* | 10/10 |
| | mean steps (finished) | 47.9 | 39.8 | 46.0 |

\* the one miss died at step 2 on `TlsError("bad decrypt")` mid-stream; the other nine validated.

[...]

PR #1090, opened 2026-08-27
ci: pin the stable MoonBit toolchain to 0.10.10+f8a486b6f

tonyfettes:
## Summary

- Stable moved to `moon 0.1.20260827` / `moonc v0.10.11` today. Its native backend crashes on `moonbit-community/graphviz@0.1.3` (pulled in by the editor via `kokic/uml@0.1.5`): `Machine_of_clam_lower.lower_break: break value is required for scalar loop`. Every `moon test --target native` in the workspace links those packages, so CI and the editor workflow fail on every branch, `main` included (#1083, #1089); nightly crashes the same way.
- `ci.yml` and `editor.yml` install `0.10.10+f8a486b6f` instead of `latest`: the version `main`'s last green run used, the one `desktop/.moonbit-version` already pins, and under which the crashing package passes (`editor/internal/viewer/contrib/hover`: 52/52).
- The two formatters disagree (0.10.11 writes trailing commas in multi-line struct literals; 0.10.10 rejects them), so this pin and the 0.10.11 reformat in #1089 are mutually exclusive — #1089 is closed until the ICE is fixed upstream, and unpinning must reformat the workspace in the same change.

Not touched: `desktop-release.yml` and `copilot-setup-steps.yml` still install `latest`; the release build may hit the same crash and should be pinned the same way if it does.

Repro of the crash with 0.10.11: `cd editor && moon test --target native internal/viewer/contrib/hover`.

## Test plan

- [x] `moon test --target native internal/viewer/contrib/hover` passes under `0.10.10+f8a486b6f` (isolated install)
- [x] YAML parses
- [ ] CI (stable, native/js) and Editor green on this PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,14 @@ jobs:
         target: [native, js]
         include:
           - moonbit-channel: stable
-            moonbit-version: latest
+            # Pinned, not `latest`: stable 0.10.11 (moon 0.1.20260827) crashes
+            # the native backend on moonbit-community/graphviz@0.1.3
+            # ("Machine_of_clam_lower.lower_break: break value is required for
+            # scalar loop"), which the editor pulls in through kokic/uml, so
+            # every `moon test --target native` dies before running. Its
+            # formatter also changed (trailing commas in struct literals), so
+            # unpinning means reformatting the workspace in the same change.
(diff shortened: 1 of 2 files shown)


tonyfettes:
Closed, b/c release does not require moon test.

tonyfettes:
Closed, b/c release does not require moon test.

PR #1103, opened 2026-08-28
Use Json(...) constructor shorthand for ToJson-typed values

bobzhang:
## Summary
Replace single-argument `Json::string(...)`, `Json::number(...)`, `Json::boolean(...)`, `Json::array(...)`, and `Json::object(...)` calls with the `Json(value)` constructor, which accepts any `T : ToJson` (`Json::Json(value : T) -> Json`). The old helpers still exist but are redundant at these call sites.

Found with a structural scan (`Json::$_($_)` via moongrep; rg cross-check) and verified call-site by call-site: every replaced argument type implements `ToJson` (`String`, `Bool`, `Double`, `Int`, `Array[Json]`, `Map[String, Json]`, ...), so `Json(x)` produces the identical value.

Empty map/array literals that lost their type context are explicitly annotated — `Json((Map([]) : Map[String, Json]))` and `Json(([] : Array[Json]))` — to keep type inference unambiguous (no `unresolved_type_variable` warnings).

## Deliberately left unchanged
- `Json::null()`, `Json::empty_object()` — zero-argument, no `ToJson` argument.
- `Json::from_json` / `Json::stringify` matches are actually `FromJson::from_json` trait calls and `Json::stringify(x)` method calls — not constructors.
- Variant patterns (`Json::Array(...)` etc. in `match`) and doc-comment lines.

## Verification
- `moon check`: 0 errors, 0 warnings; `moon fmt --check`: clean.
- 410 call sites across 70 files; `ToJson` semantics probed for every argument category (int/double literals, empty literals, maps, arrays).
- `moon test`: 2059/2065 passed. The 6 failures are all in `agent_tool/shell` (subprocess output-cap/timing tests) and are environmental to this machine's sandbox, not this change: the same suite passes 138/138 on a clean worktree of this exact commit, and this branch does not touch `agent_tool/shell`. A raw `@process.spawn` probe in the workspace shows the sandbox mangles child stdout (UTF-16 bytes), which hangs the shell tool's output collection — CI runs in a clean environment.

One edge note: `Json::number(x)` → `Json(x)` for `Double` args changes NaN/±Infinity handling to `Double.to_json`'s string encoding ("NaN"/"Infinity") instead of a `Number` — none of the converted call sites can produce those values.

Generated with SeekMoon


bobzhang:
## CI status

All **stable** checks pass: `check (stable, js)`, `check (stable, native)`, `moon (stable, macos-latest)`, `moon (stable, ubuntu-latest)`, `browser-correctness`, GitGuardian.

The two failing jobs are `moon (nightly, ...)`, and they fail on the **nightly formatter drift**, not on this change: the nightly `moon fmt` now wants trailing commas / different record-literal spacing in 10 files that this PR does not touch (all under `editor/base/*` and `editor/internal/shell/examples/` — zero overlap with the 70 changed files). The stable and nightly formatters disagree in the opposite direction too (see the recent `codex/desktop-ux-remediation` CI failure, where the stable check rejected committed trailing commas), so there is no formatting that satisfies both gates; the nightly check is currently red for any PR against this repo. Resolving it needs a toolchain/CI decision outside this change (e.g. pinning the formatter or accepting nightly drift).

Everything this PR changes is green under the stable toolchain.

Generated with SeekMoon

PR #1106, opened 2026-08-28
feat(desktop): show every tool call as its own transcript row

bobzhang:
## Why

The desktop transcript folded a whole model step behind one summary line, so a step with many tool calls read as a single collapsed sentence (`#2 · Read moon.mod · moon test · +5 more`) and every call hid behind two levels of disclosure. The viz HTML render already does better: each call is an always-visible row that folds only its own arguments and output.

## What

- `activity_view` no longer wraps a step in one `<details>`. Each tool call renders as its own always-visible row (`tool_call_view`), grouped behind the same guide line an expanded fold used to draw. Only reasoning stays behind a fold.
- A step that opens with a thought carries the ordinal and commit instant on that fold's summary (`#2 14:32:10 · Thought`); a step that goes straight to tools gets a plain heading line with the same columns.
- Failed calls still start open and read red, and now carry a literal `failed` marker on their own row, replacing the step-level badge (state never rests on color alone).
- The plan progress meter moves from the step summary to the `plan` call's own row, next to the checklist it summarizes.
- The category/brief summary machinery (`tool_group_summary`, `tool_activity_summary`), the burst fold (`tool_line_view`), and the singleton-inlining special cases (`inline_singleton_tool`, `flatten_tool_activity`) are no longer needed and are removed, along with the `.tool-line` CSS.

Net −236 lines.

## Verification

- `moon check --target js --deny-warn` and `moon check --target native --deny-warn` from the workspace root: clean.
- `moon test --target js` in `desktop/`: 3264/3264 pass (folding tests rewritten to assert the per-call rows, heading/thought-fold header placement, failed marker, and per-row plan meter).
- `moon fmt --check` verified with CI's pinned toolchain (`0.10.10+f8a486b6f`, fetched into a scratch dir): clean — the workspace was deliberately *not* reformatted with the newer trailing-comma formatter.

https://claude.ai/code/session_0153tDruHBecLpa661xnh2Sb

### DIFF
--- a/desktop/frontend/plan/plan.mbt
+++ b/desktop/frontend/plan/plan.mbt
@@ -165,7 +165,7 @@ pub fn classify(item : @transcript.TranscriptItem) -> Call? {
 /// unresolved for the moment before its rejection lands, and blanking the
 /// meter there would flicker on every one of them.
 ///
-/// Over one folded run of activity this is the plan as that run left it.
+/// Over one run of activity this is the plan as that run left it.
 pub fn latest(items : Array[@transcript.TranscriptItem]) -> Array[Step]? {
   for item in items.rev_iter() {
     match classify(item) {
--- a/desktop/frontend/plan/view.mbt
+++ b/desktop/frontend/plan/view.mbt
@@ -1,6 +1,7 @@
-// Two renderings of the same checklist, at two altitudes: a meter on a folded
-// run, and the full card inside an expanded `plan` call. They share the row
-// and meter builders below so the two cannot drift apart.
+// Two renderings of the same checklist, at two altitudes: a meter riding
+// the `plan` call's own transcript row, and the full card inside that row
+// once expanded. They share the row and meter builders below so the two
+// cannot drift apart.
 //
 // There is deliberately no third, standing rendering. A plan is a durable
 // tool call and reads as one, in the turn that made it; hoisting the latest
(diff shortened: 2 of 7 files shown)


bobzhang:
**Codex (subal, xhigh) review: signed off** after 6 rounds against `main`, each on a fresh instance.

Final verdict (round 7, no findings):

> The patch consistently exposes individual tool-call rows while preserving reasoning disclosures, failure indicators, plan progress, and transcript grouping. The changed workspace passes `moon check --target all` and the targeted transcript tests.

Findings raised and addressed along the way (all P2, all fixed in follow-up commits on this branch):

1. Tool-only step headings misaligned the `#N`/instant columns vs reasoning-led steps (flex gap, no `▸`-width lead) → hidden-marker lead + markup-space gap (`a75339608`).
2. Long briefs wrapped the call row / forced the pane wider → one-line ellipsis on summary briefs, fold marker moved out of the ellipsized span (`ecc9af839`).
3. That ellipsis clipped bodyless `.tool-call-label` rows whose label is the text's sole copy (goal notices) → ellipsis scoped to summaries; labels wrap (`54f65947e`).
4. A clipped summary brief had no path to its full text → full label as `title` tooltip + un-clamp when the row is open (`0d8a1b3b3`).
5. The open-state un-clamp let unbroken tokens (Codex command strings) paint past the pane → `overflow-wrap: anywhere` (`18b311804`).
6. Same overflow for unbroken tokens in bodyless labels → `min-width: 0` + `anywhere` on `.tool-call-label` text (`c2a51ca85`).

Also verified locally: `moon check` js+native `--deny-warn` clean, 3264/3264 desktop js tests, and `moon fmt --check` clean under CI's pinned toolchain `0.10.10+f8a486b6f`.

PR #1109, opened 2026-08-28
feat(desktop): separate tool calls from results

bobzhang:
## Summary

- render tool requests and results as distinct transcript rows, matching the viz HTML renderer
- group requests before results only when multiple calls share the same durable Assistant event sequence; keep sequential calls paired request/result
- keep request argument tabs and result output tabs independent, including stable radio groups for multi-call events
- key request/result DOM rows so late results preserve disclosure and tab state
- keep streamed output visible for pending Codex commands as a distinct `Tool output` row without claiming completion

## Independent Codex review

A first fresh Codex CLI review identified three P2 findings; this PR addresses all three with keyed late-result reconciliation, preserved synthetic runtime-notice labels, and single-pass result view construction. A second fresh review found one P1 lifecycle regression where pending command output disappeared; that is fixed and covered by a regression test. A final fresh Codex CLI review against commit `ecc2d65bb` found no actionable issues.

## Verification

- `moon info && moon fmt`
- focused transcript component tests: 56/56
- `just check`
- `just test` (native 3327/3327, JS 3273/3273, cram 28/28)
- `just build`
- native desktop visual QA against a recorded session: confirmed two- and three-call bursts render all request rows before result rows; expanded requests show Arguments / Original JSON while expanded results show Output separately

### DIFF
--- a/desktop/frontend/transcript/component/edit_diff_tests.mbt
+++ b/desktop/frontend/transcript/component/edit_diff_tests.mbt
@@ -83,7 +83,7 @@ test "a payload with no readable change decodes to nothing" {
 
 ///|
 #warnings("-alert_unstable")
-test "an edit row tabs its change, payload, and result" {
+test "an edit row tabs its change and payload apart from its result" {
   let arguments =
     #|{"path":"lib/parser.mbt","start_line":12,"old_string":"args.length()","new_string":"args.len()"}
   let rendered = tool_call_view({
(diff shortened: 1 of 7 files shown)


bobzhang:
Independent Codex CLI review completed on the final head (`ecc2d65bb`).

- First fresh reviewer found three P2 issues; fixed in `fbe4826eb`.
- Second fresh reviewer found the pending streamed-output regression; fixed in `ecc2d65bb` with a regression test.
- Final fresh reviewer (session `01a047bb-7c57-7e83-9ef5-f0eb35e3fbd1`) found no actionable issues.

Local final-head validation is green: `just check`, `just test` (native 3327/3327, JS 3273/3273, cram 28/28), and `just build`.

PR #1126, opened 2026-08-29
feat(mbtx): distinguish compile and runtime errors

bobzhang:
## Summary

- classify foreground MBTX failures from structured compiler diagnostics and built artifacts while keeping a single moon run invocation
- retain the build-only compile preflight for explicit background runs so jobs are scheduled only after compilation succeeds
- present compile-time and runtime failures distinctly in both the desktop transcript and standalone HTML/viz renderer
- cover compiler diagnostics, runtime failures, compiler-like program output, exit 255, warnings, and legacy visualization behavior

## Testing

- moon test viz --target js
- just check
- just test
- just build
- just editor-test

The optional just editor-test-browser smoke could not start in this worktree because editor/node_modules/.bin/playwright is absent.

### DIFF
--- a/agent/tool_definition.mbt
+++ b/agent/tool_definition.mbt
@@ -193,7 +193,9 @@ async test "a background run that fails to compile reports the error inline" {
     guard action is Respond(output) else { fail("expected Respond") }
     assert_true(output.is_error)
     assert_false(output.content.contains("started background job"))
-    assert_true(output.content.contains("did not compile"))
+    assert_true(output.content.contains("compile-time error"))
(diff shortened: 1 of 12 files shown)


bobzhang:
Superseded by the staged-execution line that just landed: #1167 restructured mbtx into a synchronous build phase + auto-backgrounded run (classification by construction, no diagnostics parsing or artifact probing), and #1168 / #1170 shipped the desktop and viz/HTML stage rendering plus the build-errors-only filter. Everything this PR set out to do is covered there.

PR #1142, opened 2026-08-30
refactor(mbtx): separate handoff and foreground timeouts

bobzhang:
Stack 1/3.

## Summary
- split the runtime-backed auto-background threshold from the non-detachable foreground hard timeout
- preserve existing behavior: both defaults remain 300 seconds
- retain `run_timeout_ms` as a compatibility fallback while allowing either new bound to be overridden independently
- add a routing test that proves runtime-less subruns ignore the handoff threshold

## Review
Reviewed with a fresh Codex CLI using `model_reasoning_effort="ultra"`; all findings were addressed.

## Validation
- `moon test agent_tool/mbtx` (51 passed)
- `moon test agent` (92 passed)
- `moon check --warn-list +73` (0 errors; 36 pre-existing warnings)
- `moon info && moon fmt`
- `git diff --check`

### DIFF
--- a/agent/tool_definition.mbt
+++ b/agent/tool_definition.mbt
@@ -265,7 +265,7 @@ async test "a foreground run past its deadline becomes a job instead of dying" {
       workspace_root=".",
       bg_runtime~,
       job_dir=spill_dir,
-      run_timeout_ms=1500,
+      auto_background_after_ms=1500,
     )
     let source =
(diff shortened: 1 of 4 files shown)


bobzhang:
Superseded by #1148, which contains this commit and the complete automatic-handoff series rebased onto the latest main. Closing this stacked entry to keep one review and merge path.

PR #1143, opened 2026-08-30
refactor(mbtx): hand off long runs automatically

bobzhang:
Stack 2/3; depends on #1142.

## Summary
- run every runtime-backed `mbtx` call through one retained foreground execution
- return inline for up to 5s, then automatically adopt the same process as a background job
- remove the model-facing `run_in_background` choice and reject stale calls with migration guidance
- preserve source-path rewrites and actual sandbox-policy state across handoff
- reclaim handed-off snippet build directories at terminal status
- cache terminal denial classification and isolate rewrite metadata from mutable callers

## Review
A fresh Codex CLI review ran with `model_reasoning_effort="ultra"`. Its correctness findings around terminal cleanup, non-wasm policy state, repeated full-log scans, stale-argument handling, `job_stop` guidance, and mutable rewrite aliases were addressed with regression tests.

Two findings preserve the explicitly chosen scope of this stack: output that crosses the existing 48k inline cap before handoff is still terminated with actionable feedback, and the public compatibility timeout alias retained by stack 1 is intentionally removed here as the staged API migration to the two named policies.

## Validation
- `moon test -p bobzhang/openseek/agent_tool/bgjobs` (15 passed)
- `moon test -p bobzhang/openseek/agent_tool/job_output` (13 passed)
- `moon test -p bobzhang/openseek/agent_tool/mbtx` (51 passed)
- `moon test -p bobzhang/openseek/agent` (95 passed)
- `moon check --warn-list +73` (0 errors; 36 pre-existing warnings)
- `moon info && moon fmt`
- `git diff --check`

### DIFF
--- a/agent/tool_definition.mbt
+++ b/agent/tool_definition.mbt
@@ -160,94 +160,236 @@ async test "the registry has no shell tool family" {
     assert_false(names.contains("shell_output"))
     assert_false(names.contains("shell_stop"))
     guard tools.find("mbtx") is Some(mbtx) else { fail("mbtx tool missing") }
-    // A job runtime is wired here, so the background argument must be OFFERED:
-    // the schema is what decides whether the model can make the call at all.
+    // Background handoff is automatic, so the schema carries no duration-mode
+    // choice for the model to make.
     let JsonSchema(schema) = mbtx.schema
     guard schema is { "properties": Object(properties), .. } else {
       fail("expected an object schema")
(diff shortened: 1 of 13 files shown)


bobzhang:
Superseded by #1148, which contains this commit and the complete automatic-handoff series rebased onto the latest main. Closing this stacked entry to keep one review and merge path.

PR #1144, opened 2026-08-30
chore: migrate deprecated lexscan to lexmatch

bobzhang:
## What

Every `lexscan` in the repo targets a `String` or `StringView` — the non-streaming form the compiler now deprecates:

> Warning (deprecated_syntax): The syntax `lexscan` with a String target for non-streaming lexical matching is deprecated. Use `lexmatch` instead.

25 sites across both workspaces (9 in `editor/`, 16 in the root). These warnings are currently the **only** thing keeping `just check` and the editor's `moon check --target all --warn-list +73 --deny-warn` from passing clean, so they are on track to break the stable CI gate as soon as the toolchain there catches up.

## Change

Pure keyword rename at 25 sites. Every changed line is the keyword and nothing else — verified by filtering the diff for any line that isn't `-lexscan` / `+lexmatch` (empty).

**On safety of the sweep:** the sites were taken from the *compiler's own warning locations*, not a textual `grep`. Streaming `lexscan` over a `@lexbuf.Lexbuf` / `AsyncLexbuf` / `StringScanner` is **not** deprecated and must keep the old keyword. I diffed the warned set against every `lexscan` in the tree: the two sets are identical, so there are no streaming uses here — but that is a fact I checked rather than assumed.

Two doc touch-ups fall out of this:
- `editor/syntax/lang_moonbit/lexer.mbt` already described its rule sets as "`lexmatch` compiled to a tagged DFA", so the code now agrees with its own module doc.
- One comment in `cmd/openseek/mcp.mbt` naming the old keyword is updated.

## Testing

Fully verified:
- `just check` — `moon check --target native --deny-warn`, `--target js --deny-warn`, `moon fmt --check` all **clean**. Previously 25 warnings.
- `moon check --target all --warn-list +73 --deny-warn` in `editor/` (the exact stable-CI command) — **clean**. Previously 9 warnings.
- `moon test --target all` in `editor/` — native 1161, js 1738, wasm 1033, all passing, counts identical to pre-change.
- Root packages I could build: `shell_parse` 24, `command_policy` 9, `internal/error` 5 — all passing.

**Could not run locally, flagging honestly:** the root `moon test` does not build on my machine, on this branch *and on clean main alike* (20 identical `Error: [4222]`). The cause is the vendored `moonbitlang/x` 0.4.49 dep — `.mooncakes/moonbitlang/x/path/win32/path.mbt:278` uses `lexscan` over a `String`, which my local toolchain (moonc v0.10.7, 2026-08-07) has escalated from deprecation to a hard error. Any package depending on `@path` therefore can't build locally, which covers `agent_tool/shell`, `agent_tool/write`, `agent`, `cmd/openseek`, and `cmd/viz_app`. This is pre-existing and unrelated to this PR, so **CI is the gate for those packages.**

That dependency is worth noting on its own: it is the same deprecation, one level down, and it will break the build outright the moment CI's toolchain advances to where mine already is.

## Risk

`agent_tool/shell/sandbox.mbt` and `command_policy.mbt` parse shell commands for the sandbox, so they are security-relevant. A rename the compiler accepts and whose tests pass is low risk, but reviewers should look there first.

### DIFF
--- a/agent/goal.mbt
+++ b/agent/goal.mbt
@@ -340,7 +340,7 @@ fn decode_goal_status(arguments : Json) -> Result[@goal.GoalStatus, String] {
 /// Tool packages normally share `agent_tool/internal/error`, but the agent is
 /// outside that internal visibility boundary.
 fn goal_decode_failure_message(error_text : String) -> String {
-  lexscan error_text {
+  lexmatch error_text {
     // Greedy across lines to the LAST ` FAILED: ` marker — the same piece
     // rev_find chose.
     (re"^(.|\n)* FAILED: ", after=message) =>
--- a/agent_tool/internal/auto_check/parse_gate.mbt
+++ b/agent_tool/internal/auto_check/parse_gate.mbt
@@ -317,7 +317,7 @@ fn loc_line_span(loc : String) -> (Int, Int)? {
   // An end that does not open with digits (or is missing) falls back to the
   // start line, matching the compiler's lenient loc forms; parse failures on
   // validated digit runs can only be overflow, which is malformed too.
-  lexscan loc {
+  lexmatch loc {
     (re"^[0-9]+" as start) +
     re"(:[^\-]*)?-" +
     (re"[0-9]+" as end_) +
--- a/agent_tool/internal/error/error.mbt
+++ b/agent_tool/internal/error/error.mbt
@@ -3,7 +3,7 @@
 /// rendering. This keeps tool argument errors stable when validation lives in
 /// an internal package and the raw error string includes source locations.
 pub fn failure_message(error_text : String) -> String {
-  lexscan error_text {
+  lexmatch error_text {
     // Greedy across lines to the LAST ` FAILED: ` marker — the same piece
     // rev_find chose.
     (re"^(.|\n)* FAILED: ", after=message) =>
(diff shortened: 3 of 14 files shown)


bobzhang:
## Codex review sign-off

**Verdict: no findings.**

> The patch consistently replaces deprecated non-streaming `lexscan` uses with the equivalent `lexmatch` syntax without changing matching logic. No remaining repository-owned `lexscan` occurrences or formatting issues were found.

Codex verified all 14 changed files against the base, confirmed no repository-owned `lexscan` remains, and re-ran `moon fmt --check` clean.

Note: the untracked `desktop/openseek.stdin` that shows up in codex's `git status` output is test residue from a local run, not part of this PR — nothing untracked is committed here.

bobzhang:
Closing as **obsolete** — superseded by work already on `main`.

This branch was cut from `f7d341a8` and `main` has since advanced **453 commits**. In that window the identical migration already landed: `main` now has **zero** repository-owned `lexscan` occurrences, every file this PR touched already uses `lexmatch`, and even the one comment this PR rewrote (`cmd/openseek/mcp.mbt`, "an unanchored `lexmatch` arm SEARCHES") is already updated there.

So there is nothing left to contribute, and merging would be actively harmful: the branch is 453 commits stale, so squashing it onto `main` would revert three weeks of unrelated work in `agent_tool/shell/sandbox.mbt`, `editor/syntax/lang_moonbit/lexer.mbt`, and others. Closing instead of rebasing to an empty diff.

The one item from this PR's description still worth acting on separately is the dependency note: `.mooncakes/moonbitlang/x/path/win32/path.mbt:278` uses `lexscan` over a `String`, which newer toolchains (moonc v0.10.7+) reject as a hard `Error: [4222]` rather than a deprecation. That breaks the root `moon test` locally for anything depending on `@path`. It needs a `moonbitlang/x` bump, not a change in this repo.

PR #1156, opened 2026-08-31
fix(desktop): refine Windows immersive titlebar

Milky2018:
## Summary

- give native Windows overlay windows a compact 40 px chrome row and reserve the real caption-control side of the header
- publish explicit CEF drag/no-drag regions, including a dedicated drag strip for pages without a topbar, while preserving native double-click maximize behavior
- synchronize native Windows caption glyphs with the effective OpenSeek theme and retain that choice across activation, DPI, and system-theme messages
- keep the Proton 0.1.18 prebuild path while passing direct MSVC linker flags, preferring the detected MSVC toolchain over generic `CC`/`CXX`, and linking the required User32 APIs
- cover platform classes, drag-strip visibility, native theme flags, endpoint isolation, and packaged CSS contracts

The native theme bridge is a temporary workaround for Proton's hard-coded dark
caption mode. Its removal condition is tracked inline against
https://github.com/moonbit-community/proton/issues/214.

## Testing

- `moon -C desktop fmt --check`
- `moon -C desktop test internal/platform --target native` — 1 passed
- `moon -C desktop test internal/extension --target native` — 7 passed
- `moon -C desktop test package/internal/packaging --target native` — 17 passed
- `git diff --check`

## Validation boundaries

- `moon -C desktop info` and `moon -C desktop test --target js` are blocked
  locally before reaching these changes by the checked-out `moondiff`
  dependency: `Value type is not allowed for type with type parameter as
  field` in `astdiff/dijkstra.mbt:12`. The repository stable JS/native CI
  passed before the Windows linker follow-up and is running again for the
  updated head.
- A Windows build exposed and motivated the direct-linker flag and User32
  fixes. The updated head still needs a Windows rebuild and GUI pass to confirm
  the final native link and titlebar behavior.


desktop/internal/platform/platform_native.c
tonyfettes, line 1: @Milky2018 Should we to upstream this part to proton?

Milky2018:
Before:

企业微信截图_a7662094-565b-4221-9546-9b3b3a82c5c3

After:

企业微信截图_fd72dd9c-88ec-4764-a843-99f7bbab84db

desktop/internal/platform/platform_native.c
Milky2018: https://github.com/moonbit-community/proton/issues/214

Maybe in the next release

tonyfettes (approved):

PR #1163, opened 2026-08-31
fix(desktop): render durable step prose before its activity

### LINKED ISSUE #1075: OpenSeek transcript renders same-step activity before assistant prose
## Summary

OpenSeek renders reasoning/tool activity from a durable assistant step before that step's user-facing prose.

For an assistant event containing prose followed by a tool call, the transcript currently appears as:

```text
activity / tool call
assistant prose
```

Expected:

```text
assistant prose
activity / tool call
```

## Minimal reproduction

Use one durable OpenSeek assistant sequence with these visible rows:

```text
Assistant("I will inspect the project.", sequence=2)
Tool(name="read", arguments="{}", sequence=2)
```

A temporary white-box test against `TranscriptRenderer::stream_children` consistently produced:

```text
actual:
["activity\0head", "row\0s2\0o0"]

expected:
["row\0s2\0o0", "activity\0after\0s2\0o0"]
```

The test failed identically on two consecutive runs. A DOM-level reproduction through the production `transcript_view` also placed the activity markup before the prose (`activity_at=511`, `assistant_at=926`) on two runs.

As a control, changing only the transcript source from `OpenSeek` to `Codex` restored the expected prose-before-activity order.

## Root cause

The durable session projector preserves the assistant payload structure in `desktop/frontend/transcript/sessions.mbt:296-339`. In particular, tool calls are appended after visible assistant prose (`:322-339`).

The OpenSeek-only durable-step folding branch in `desktop/frontend/transcript/component/rendering.mbt:197-225` then:

1. collects every `Reasoning` and `Tool` row with the same durable sequence;
2. inserts the combined activity node at `:208-212`;
3. inserts assistant prose afterward at `:217-225`.

The comment at `:214-216` explicitly describes this activity-first behavior. The assertion in `desktop/frontend/transcript/component/view_tests.mbt:258-262` currently locks it in as expected output.

This rules out CSS, Rabbita keyed reconciliation, session projection, and streaming timing as the cause: the keyed child map itself is already in the wrong display order for OpenSeek.

## Suggested fix

Within the same durable-step folding branch, render non-activity prose before inserting the combined activity node. Preserve the activity key derived from the pre-step `left_boundary` so that the live-to-durable handoff can continue reusing the browser-owned `<details>` node even when it moves after the newly committed prose.

Update the existing display-order assertion and add a regression test for an assistant event containing both non-empty `content` and `tool_calls`.

0717lee:
Closes #1075.

## What this does

In the OpenSeek durable-step folding branch of `TranscriptRenderer::stream_children` (`desktop/frontend/transcript/component/rendering.mbt`), the combined activity node was inserted before the step's assistant prose, so an event containing prose followed by a tool call displayed as activity first, prose second.

The branch now renders the non-activity prose rows first and inserts the combined activity after them. The activity keeps its key on the pre-step `left_boundary` (captured before the prose loop), so:

- the live-to-durable handoff continues reusing the browser-owned `<details>` element, and
- the provisional next-step card (keyed after the last rendered row) never collides with the settled card.

One note on the repro in #1075: its "expected" key list keys the activity after the prose row, while the issue's own suggested-fix paragraph asks to preserve the pre-step boundary key. The two disagree for a step at the head of the stream, and the pre-step key is what keeps the settled card's identity identical to the live card's (see `live step hands off to its durable assistant card key`, which this branch keeps passing), so this PR follows the suggested-fix paragraph. Display order matches the issue either way.

## Tests

- Updated the two existing display-order assertions in `view_tests.mbt` — keys are unchanged; only their insertion order moves.
- Added a regression test mirroring the issue's minimal repro: one durable assistant event with non-empty prose followed by a tool call in projector order asserts the prose row precedes the activity card, and the activity card contains the tool call but not the prose.

Verified locally: `moon check desktop/frontend --target js` clean; `moon test desktop/frontend/transcript/component --target js` 57/57 passing.

## Limitations

- Verified through the renderer's keyed child map (the same surface the issue's repro probes), not through a running desktop app.
- Codex and legacy transcripts are untouched: rows without OpenSeek step ordinals keep the adjacent-activity behavior.

### DIFF
--- a/desktop/frontend/transcript/component/moon.pkg
+++ b/desktop/frontend/transcript/component/moon.pkg
@@ -5,6 +5,7 @@ import {
   "moonbit-community/rabbita/clipboard",
   "moonbit-community/rabbita/cmd",
   "moonbit-community/rabbita/dom",
+  "moonbit-community/rabbita/internal/vdom",
   "moonbit-community/rabbita/html",
   "moonbit-community/rabbita/js",
   "moonbit-community/rabbita/sub",
--- a/desktop/frontend/transcript/component/moon.pkg
+++ b/desktop/frontend/transcript/component/moon.pkg
@@ -5,7 +5,6 @@ import {
   "moonbit-community/rabbita/clipboard",
   "moonbit-community/rabbita/cmd",
   "moonbit-community/rabbita/dom",
-  "moonbit-community/rabbita/internal/vdom",
   "moonbit-community/rabbita/html",
   "moonbit-community/rabbita/js",
   "moonbit-community/rabbita/sub",
--- a/desktop/frontend/transcript/component/rendering.mbt
+++ b/desktop/frontend/transcript/component/rendering.mbt
@@ -220,20 +220,12 @@ fn TranscriptRenderer::stream_children(
           }
           index += 1
         }
-        if !activity.is_empty() {
-          items.set(
-            TranscriptRenderKey::activity_after(left_boundary).wire(),
-            activity_view(
-              activity,
-              self.on_open,
-              step_label="#\{step_number}",
-              subrun?=self.subrun_link(),
-            ),
-          )
-        }
-        // Assistant narration and final answers remain visible. The activity
-        // is emitted first as the step's work disclosure; prose then becomes
-        // the stable left boundary for the next provisional/durable step.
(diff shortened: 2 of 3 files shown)


tonyfettes:
Superseded by #1260

PR #1164, opened 2026-08-31
feat(desktop): highlight active historical diff row

Yu-zh:
## Summary

- keep the matching commit-history file row highlighted while its historical diff is active
- match the complete immutable diff identity so identical paths in other commits are not selected
- reuse the existing accent tokens and expose the state with aria-current

## Verification

- moon test desktop/frontend/fileeditor --target js
- just check
- repeated both commands in an isolated worktree at the committed revision


Yu-zh:
The highlight on the file name is what I added

image

bzy-debug (approved):

PR #1165, opened 2026-09-01
feat(mbtx): classify compile-time vs runtime failures structurally

bobzhang:
## Problem

A failed `mbtx` call reported `[mbtx: exited 1]` whether the **build rejected the snippet** or the **program ran and trapped** — `moon run` exits 1 for both. The model couldn't tell "fix the source" from "debug the program", and nothing in the output can say either: a snippet legitimately runs `moon check`/`moon test` as child processes, so compiler-diagnostic text in the log is routinely the program's own *successful* output. Any output-based classifier would misfire exactly there.

## Design: a structural signal

After `moon run` exits, probe the throwaway `--target-dir` for the runnable artifact moon links **only after a successful build** — verified against moon's own build plan (`--dry-run`) for all four targets:

| target | artifact |
|---|---|
| wasm / wasm-gc | `<target>/debug/build/single/single.wasm` |
| js | `js/debug/build/single/single.js` |
| llvm | `llvm/debug/build/single/single.exe` |

The dir is fresh per call, so **presence proves the program was launched**, and **absence on a failed run proves the failure came from the build** — the framing deliberately claims no more than that (`BUILD failed — the program never ran`), since absence alone cannot distinguish compiler rejection from a toolchain/dependency fault; the diagnostics below the label say which. The probe is tri-state (`Bool?`): a probe fault (EIO/EACCES) claims **neither** stage and keeps the plain exit report, per the repo rule against folding unrelated errors into the "missing" case. The release profile is probed as a fallback so a future `moon run` default flip degrades to an extra stat.

## What changes

- **Foreground failures** are labeled: `[mbtx: BUILD failed — the program never ran; the output below is the build's]` vs `[mbtx: exited N at RUNTIME — the program compiled and ran]`, with briefs `mbtx (build failed)` / `mbtx (exit=N)`. Success output is untouched.
- **Foreground timeouts** say *which half* never finished: build vs program (probe-unknown keeps the original hedge).
- **Handed-off jobs**: `BgJobRuntime::adopt` gains an optional launcher-supplied `exit_note` classifier. The exit watcher computes it at terminal time **before** `on_terminal` reclaims the build dir (the classifier's evidence), caches it, and it rides both the completion notice and `job_output`.
- The exit note is **single-flight** (`NotComputed → Computing → Cached`): exactly one task runs the classifier; a `job_output` read racing the exit joins the watcher's flight instead of recomputing — a second computation could finish after cleanup emptied the evidence directory and overwrite the correct answer (found by codex xhigh review with a deterministic failing repro; that repro is now a regression test). A cancelled flight resets so the watcher still computes and cleanup is never wedged.
- A **raising classifier is reported, not silenced**: the failure is cached as a tagged note (`exit-stage classification failed (…)`) instead of being indistinguishable from "nothing to say".
- Kills carry no exit code (`None`) into the classifier; mbtx declines to classify them since the stop reason is already named by the reader.

## Review

codex CLI (subal) xhigh reviewed the initial commit and raised 2×P1 + 1×P2 (evidence-overclaiming wording; the recompute-after-cleanup cache race; silent classifier-failure collapse) — all three addressed in the second commit, then the full branch was re-reviewed.

## Tests (all real-`moon` e2e unless noted)

- build failure → labeled + brief; runtime trap → labeled runtime; **negative control**: a snippet that *prints* compiler-diagnostic-shaped text is classified by what it *did*, not what it printed (trap → runtime; clean exit → no label).
- timeout at 50ms (below any possible compile) → "BUILD never completed"; compiled/unknown timeout wording pinned by whitebox tests.
- the per-target artifact table pinned with fabricated artifacts (llvm cannot even build on this toolchain: missing core bundle) incl. the all_pkgs.json non-artifact and the release fallback.
- bgjobs: note computed **before** terminal cleanup; the reviewer's race repro as a regression test (racing reader joins the flight, classifier runs exactly once); a raising classifier yields the tagged note; `None` exit code for kills; no classifier → no note.
- `job_output` renders the note between output and footer; a handed-off build failure (handoff at 1ms) is classified in its completion notice.

`moon check --deny-warn` clean repo-wide; 191/191 tests green across the four touched packages; `moon fmt --check` clean; only the intended `bgjobs` mbti surface changed.

### DIFF
--- a/agent/tool_definition.mbt
+++ b/agent/tool_definition.mbt
@@ -699,11 +699,11 @@ async test "a finished background job queues a notice and wakes the controller" 
 }
 
 ///|
-/// A compile error that only surfaces AFTER the handoff is still classified:
+/// A build failure that only surfaces AFTER the handoff is still classified:
 /// the adopted job's exit note is computed at terminal time (before the build
 /// dir is reclaimed) and rides the completion notice, so the model learns the
 /// snippet never built without spending another tool call.
-async test "a handed-off compile error is classified in its completion notice" {
+async test "a handed-off build failure is classified in its completion notice" {
   @async.with_task_group(group => {
     let scope = @agent_runtime.AgentTaskScope(group)
     let notices : Array[String] = []
@@ -734,7 +734,7 @@ async test "a handed-off compile error is classified in its completion notice" {
       @async.sleep(25)
     }
     guard notices is [notice, ..] else { fail("no completion notice arrived") }
-    assert_true(notice.contains("COMPILE error"))
+    assert_true(notice.contains("BUILD failed"))
     assert_true(notice.contains("never ran"))
     @fs.rmdir(spill_dir, recursive=true) catch {
       _ => ()
--- a/agent/tool_definition.mbt
+++ b/agent/tool_definition.mbt
@@ -16,11 +16,14 @@ fn background_notice_text(snapshot : @bgjobs.BgJobSnapshot) -> String {
     Running => "running"
   }
   // The launcher's exit classification rides along (e.g. mbtx naming a failed
-  // job a compile error vs a runtime failure): the notice is where the model
+  // job a build failure vs a runtime one): the notice is where the model
   // first learns the outcome, and the snapshot's cached note is reliable here
-  // because on_job_exit fires only after the exit watcher computed it.
+  // because on_job_exit fires only after the exit watcher computed it. A
+  // failed classification is named as such rather than shown as a note.
   let note = match snapshot.exit_note {
-    Some(note) => " \{note}"
+    Some(Note(note)) => " \{note}"
+    Some(ClassificationFailed(reason)) =>
+      " (exit-stage classification failed: \{reason})"
     None => ""
   }
   "background job \{snapshot.id} finished (\{status}): `\{snapshot.command}` — read its output with job_output (job_id=\"\{snapshot.id}\").\{note}"
--- a/agent/tool_definition.mbt
+++ b/agent/tool_definition.mbt
@@ -15,7 +15,15 @@ fn background_notice_text(snapshot : @bgjobs.BgJobSnapshot) -> String {
       }
     Running => "running"
   }
-  "background job \{snapshot.id} finished (\{status}): `\{snapshot.command}` — read its output with job_output (job_id=\"\{snapshot.id}\")."
+  // The launcher's exit classification rides along (e.g. mbtx naming a failed
+  // job a compile error vs a runtime failure): the notice is where the model
+  // first learns the outcome, and the snapshot's cached note is reliable here
+  // because on_job_exit fires only after the exit watcher computed it.
+  let note = match snapshot.exit_note {
+    Some(note) => " \{note}"
+    None => ""
+  }
+  "background job \{snapshot.id} finished (\{status}): `\{snapshot.command}` — read its output with job_output (job_id=\"\{snapshot.id}\").\{note}"
 }
 
 ///|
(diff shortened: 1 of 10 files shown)


bobzhang:
## codex CLI (subal) xhigh review record — 6 rounds

This branch was reviewed by the in-house codex CLI at `model_reasoning_effort=xhigh` after every substantive push (`--base ae9dedfcb`, the merge-base). Findings converged monotonically: **P1 correctness bugs → honesty-of-claim wording → fault-injection corners.**

|---|---|---|
| 1 | **P1** wording overclaimed evidence ("COMPILE error" when absence only proves the build didn't finish); **P1** exit-note cache race — a `job_output` read racing the exit watcher could recompute after cleanup emptied the evidence dir and overwrite the correct note (reviewer built a failing deterministic repro); **P2** classifier failures silently cached as "no note" | All fixed in `bc530fc14`: evidence-honest "BUILD failed" framing, tri-state probe, single-flight exit note (repro kept as a regression test), tagged failures |
| 2 | **P2** artifact presence proves the *build* completed, not that the program launched (missing `node` case); **P2** timeout test leaned on compiler latency (50ms) | Fixed in `82eb7dcd6`: "from the run, not the build" framing; 1ms physical bound |
| 3 | **P2** build failures are not all exit 1 (reviewer reproduced **exit 255** from a dep-cache fault) and the code was dropped; **P2** watchdog-killed jobs got no stage classification; **P2** classifier failures stringified into ordinary notes | Fixed in `97596c8d1`: exit codes preserved (`BUILD failed (exit 255)`); `BgJobExit` carries kill flags so reaped/flooding jobs are classified (requested stops stay silent); typed `BgJobExitNote::ClassificationFailed` |
| 4 | **P2** a snippet can wipe its own build tree via an allowed `moon clean --target-dir` child (children are outside the wasm policy's fs rules) and relabel a runtime failure as a build one (demonstrated live) | Fixed in `ed6a27da2`: the build-failed label now requires moon's own `all_pkgs.json` residue; a bare tree claims **nothing** — sabotage can only unclassify, never relabel. Exploit kept as a regression e2e. Also makes the honest bare-tree case (dep-resolution faults) stop overclaiming |
| 5 | **P2** output-cap kills reported a fabricated `exit -1` inside stage labels while also saying STOPPED; **P2** probe faults bypassed the typed failure path in the background | Fixed in `14af29930`: killed runs keep their `None` exit and get kill framing (`mbtx (killed)`); probe faults now raise into the background's typed `ClassificationFailed`, while foreground sites choose "unclassified" explicitly |
| 6 | **P2** foreground probe-fault rendering: an EACCES/EIO stat failure on the tool's own throwaway dir folds into the unclassified state instead of a fourth tagged state | **Declined, on the record** — see below |

### Round 6: considered and declined

The trigger is an OS-level stat fault on a directory mbtx itself created and wrote three files into milliseconds earlier — not reachable by any honest input without the run itself having already failed. The current behavior **claims nothing** (plain `[mbtx: exited N]` + full output), so no falsehood can result; the collapse is between two states that both map to "no claim", which is the purpose the AGENTS.md rule protects. The background path — where a missing note is a real information loss — already carries the typed `ClassificationFailed` (round 5). Rendering a fourth evidence state through every foreground formatter would alter zero reachable behavior. The stage label is diagnostic, not a security boundary; this is documented at the probe and at the catch site.

### Final state

- 193/193 tests green across `mbtx`, `bgjobs`, `job_output`, `agent` (including the reviewer's two exploits as regression tests and a fake-diagnostics negative control)
- `moon check --deny-warn` clean repo-wide; `moon fmt --check` clean; `bgjobs` is the only changed public interface (intended)
- The one CI red (`check (stable, native)`) was a transient registry download reset (`Connection reset by peer` fetching `gmlewis/base64`); rerun issued

bobzhang:
Superseded by #1167, which restructures mbtx into a synchronous build phase + auto-backgrounded run phase — making the build/run classification true by construction and removing the need for this PR's evidence probe and bgjobs exit-note machinery. The stage-label wording, the e2e classification test patterns (including the fake-diagnostics negative control), and the review findings from this PR's six xhigh rounds all carried into #1167's design. Reopenable if the exit-note mechanism is ever wanted for other launchers.

PR #1167, opened 2026-09-01
feat(mbtx): stage wasm runs — synchronous build, auto-backgrounded run

bobzhang:
## Design

Split the default wasm target into **two phases**, per the co-design discussion:

1. **BUILD — synchronous, own bound** (10s default, injectable): `moon run --build-only --target wasm --target-dir <temp>`. A nonzero exit reports immediately as `BUILD failed (exit N) — the program never ran`. This classification is **by construction** — the run phase never starts — so no exit-code or output archaeology is needed to tell the stages apart. (`moon run` alone exits 1 for a rejected build and a trapped program alike, and diagnostics in a merged stream prove nothing: a snippet legitimately runs `moon check` as a child process, so compiler-diagnostic text is routinely a snippet's own successful output.)
2. **RUN — auto-backgrounded**: exec `moonrun` directly on the artifact moon names in its `{"artifacts_path":[…]}` report line (single-file layout as fallback) — the same process `moon run` would have exec'd; `--policy` is moonrun's spelling of `--wasm-policy` (verified against moon's `--dry-run` plan). The 5s handoff grace times **this phase only**: build time no longer counts against the run's inline allowance, cold-cache probes stop handing off spuriously, and **an adopted job is always a running program** — never a build.

Phase-2 failures are labeled `exited N at RUNTIME — the build succeeded; this failure is from the run, not the build`. Non-wasm targets (compute-only, rare) keep the single-shot `moon run` and their reports claim no stage.

## Empirical validation done before implementing

- `--build-only`: exit 0 + machine-readable `{"artifacts_path":[…]}` on success; exit 1 + diagnostics (no report line) on compile errors; ~90ms trivial build, ~0.7s with several registry imports (warm cache) — the 10s bound has 10× headroom.
- `moonrun --policy`: identical policy semantics to `moon run --wasm-policy` (TMPDIR redirect, `Sandbox policy blocked file write` / `… process spawn` phrases the existing refusal classifier keys on, allowlist prefix admit/deny); program exit codes pass through (`exit 3` → 3); traps → exit 1 + trace; 8ms startup.
- Warnings surface in the build phase and are carried into the report as build notes — build output and program output are separate captures by construction.

## What this simplifies

Classification by process boundary replaces classification by evidence inference. Relative to the approach in #1165 (which this supersedes): no artifact/residue probe, no bgjobs exit-note plumbing (`BgJobExit`/`BgJobExitNote`/single-flight), no delete-before-read race to manage — this PR touches **only mbtx and one agent test**, and the review-round failure modes from #1165 mostly become unaskable (output spoofing, evidence sabotage, cache races).

Both phases run under the same worker/read-only `sandbox-exec` confinement (wrapping extracted into a per-command helper). The escalation contract is unchanged: a grant means the policy file is simply not passed to moonrun.

## Contract change (intended)

- Build time is excluded from the inline grace; the turn now waits out the build synchronously, bounded at 10s (`build_timeout_ms`, injectable). A timed-out build reports `BUILD did not finish` — retries are warmer (moon caches downloaded modules).
- One test updated: *"a compile error after handoff keeps source diagnostics"* pinned the old behavior where a slow compile crossed the grace and became a job. That scenario is now impossible by design; its replacement pins the successor contract: a compile error is inline, labeled, `source`-named, and never a job.

## Tests

60/60 in `agent_tool/mbtx` (7 new: build-failure labeling + brief, runtime-trap labeling with a fake-diagnostics negative control, clean-exit unlabeled, 1ms build-bound timeout, non-wasm unlabeled, description staging, artifact-report parser/stripper + label helpers in whitebox), 158/158 including `agent`'s handoff e2e suite; `moon check --deny-warn` and `moon fmt --check` clean; the only interface change is the new optional `build_timeout_ms` parameter.

### DIFF
--- a/agent/tool_definition.mbt
+++ b/agent/tool_definition.mbt
@@ -210,10 +210,13 @@ async test "a compile error reached inside the grace reports inline" {
 }
 
 ///|
-/// If compilation itself crosses the grace, the same process becomes a job.
-/// Its later diagnostic must still name the submitted `source`, not the
-/// session-owned throwaway directory.
-async test "a compile error after handoff keeps source diagnostics" {
+/// Compilation can no longer cross the grace: the wasm build phase runs
+/// synchronously by design — its time does not count against the run's
+/// inline allowance — so even under the smallest possible handoff bound a
+/// compile error is reported INLINE, labeled as the build's, naming the
+/// submitted `source`, and no job is ever created for it. What the runtime
+/// adopts is always a running program.
+async test "a compile error never becomes a job, whatever the grace" {
   @async.with_task_group(group => {
     let spill_dir = @fs.tmpdir(prefix="openseek-compile-job-")
     let bg_runtime = @bgjobs.BgJobRuntime::BgJobRuntime(
@@ -232,20 +235,13 @@ async test "a compile error after handoff keeps source diagnostics" {
       Sync(_) => fail("mbtx should be async")
     }
     guard action is Respond(output) else { fail("expected Respond") }
-    assert_false(output.is_error)
-    assert_true(output.content.contains("moved to the background"))
-    guard bg_runtime.list() is [job, ..] else { fail("no job registered") }
-    let read = match @job_output.definition(bg_runtime).execute {
-      Async(execute) => execute({ "job_id": job.id, "wait_ms": 30_000 })
-      Sync(_) => fail("job_output should be async")
-    }
-    guard read is Respond(result) else { fail("expected Respond") }
-    assert_true(result.is_error)
-    assert_true(result.content.contains("source:"))
-    assert_false(result.content.contains("openseek-compile-job-"))
-    // The session-owned result container remains until teardown even though the
-    // disposable build subtree is reclaimed at terminal status.
-    assert_true(@fs.exists("\{spill_dir}/snippet-0"))
+    assert_true(output.is_error)
+    assert_false(output.content.contains("moved to the background"))
+    assert_true(output.content.contains("BUILD failed"))
+    // The diagnostic points at the model's own input, not the throwaway dir.
+    assert_true(output.content.contains("source:"))
+    assert_false(output.content.contains("openseek-compile-job-"))
+    assert_true(bg_runtime.list() is [])
     @fs.rmdir(spill_dir, recursive=true) catch {
       _ => ()
     }
--- a/agent/tool_definition.mbt
+++ b/agent/tool_definition.mbt
@@ -210,6 +210,47 @@ async test "a compile error reached inside the grace reports inline" {
 }
 
 ///|
+/// A handed-off run still carries the build phase's voice: requested compiler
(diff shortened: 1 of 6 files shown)


bobzhang:
## codex CLI (subal) xhigh review record — 3 rounds

Reviewed at `model_reasoning_effort=xhigh` after every push (`--base d2211846d`).

|---|---|---|
| 1 | **P2** stage marker lost when a run hands off; **P2** `warning:"on"` build notes dropped on handoff/timeout; **P2** artifact stat fault folded into "missing"; **P2** malformed `artifacts_path` report treated as absent | First three **fixed** in `0e261a6f7`: jobs are labeled `mbtx run (build ok): …` and the description states the invariant (a wasm job is always the running program, its exit the program's own); build notes now ride the handoff response and timeout report; a stat fault gets its own named filesystem-fault report. Fourth **declined** (rationale at `parse_artifacts_path`): malformed and absent both degrade to the stat-verified layout fallback — a wrong guess surfaces as a named build fault, never a silent misrun |
| 2 | **P1** 10s build bound can reject cold builds; **P2** uncapped warning notes could push the combined report past the agent loop's 60k tail-cutting clamp; **P2** malformed-report re-raise; **P3** `""` sentinel for notes | The clamp-margin P2 **fixed** in `79a08ff5a` (notes capped at 8k chars with a truncation marker — the 48KB body cap was sized to that margin). The rest **declined** — see below |
| 3 (verification) | Re-raises only: the 10s bound (P1) and the malformed report (P2) | No new findings; dispositions stand |

### Declined items, on the record

- **10s build bound (P1, rounds 2–3)** — this is the repo owner's explicit product decision from the design discussion ("compile time bound: make it 10s, fix later"), not an oversight: wasm builds measure ~100ms trivial / <1s with several registry imports on a warm cache, the bound is injectable (`build_timeout_ms`), a timed-out retry is warm (moon caches downloaded modules), and the failure is named (`BUILD did not finish`), not silent. Marked for revisit ("fix later") — e.g. raising the default or letting a cold build hand off — as a follow-up decision, not a blocker.
- **Malformed `artifacts_path` report as a distinct fault (P2, rounds 1–3)** — malformed and absent reports both degrade to the single-file layout fallback, and the existence check that follows validates whichever path was chosen; even with a stale fallback the outcome is a correctly-classified, named build fault. Naming the parse failure instead would matter only on a moon that broke its report line AND its layout at once.
- **`String?` for build notes (P3, round 2)** — blank means "prepend nothing", the identity element of the only operation ever applied to it; no caller can be bitten by an absence/empty ambiguity.

### Final state

- CI green on the final commit `79a08ff5a`: `check (stable, native)` 8m7s, `check (stable, js)` 3m39s.
- 161/161 tests across `agent_tool/mbtx` + `agent` (incl. regression tests for every accepted finding); `moon check --deny-warn` and `moon fmt --check` clean; interface delta is one optional `build_timeout_ms` parameter.

PR #1168, opened 2026-09-01
feat(desktop): name a failed mbtx row's stage — build error vs runtime error

bobzhang:
Follow-up to #1167 (1 of 2 visualizer follow-ups; the viz HTML renderer + compile-error filter comes separately).

## What

A failed `mbtx` row in the desktop transcript now names its **failure stage** instead of the generic red `failed`:

- `build error` (amber, `--color-warning`) — the brief is a build-phase shape: `mbtx (build failed, exit=N)`, `mbtx (build timeout)`, `mbtx (build fault)`. The fix lives in the source.
- `runtime error` (danger red) — the brief is `mbtx (exit=N)` on an errored call: the program built and then failed.

Both the call row and the result row carry the marker, so the stage is visible before the card is opened. Every other tool — and any mbtx brief outside the two stage shapes (a timeout, a kill) — keeps the generic `failed` word.

## How

Classification reads the persisted **brief** alone: the display-only caption the engine writes for renderers (the channel the desktop component already consumes), so no protocol change and no content parsing. Sessions recorded before the staged engine are deliberately not special-cased (per the design discussion): their briefs match the shapes they match, nothing more.

## Tests

57/57 in `frontend/transcript/component` (new: all three build-brief shapes map to `build error`, `exit=N` errors map to `runtime error`, clean exits carry no marker, out-of-shape briefs keep `failed`); 462/462 across the desktop frontend js suite; `moon -C desktop check --target js` clean; no `.mbti` changes (all additions private).

### DIFF
--- a/desktop/frontend/transcript/component/mbtx.mbt
+++ b/desktop/frontend/transcript/component/mbtx.mbt
@@ -75,3 +75,51 @@ fn mbtx_card(arguments : String) -> @html.Html? {
     ]),
   )
 }
+
+///|
+/// The failure stage a finished `mbtx` call reports, read from its brief —
+/// the display-only caption the engine persists for renderers. The staged
+/// wasm engine writes `mbtx (build failed, exit=N)` / `mbtx (build timeout)`
+/// / `mbtx (build fault)` for build-phase failures and `mbtx (exit=N)` for a
+/// program that built and then failed; transcripts recorded before staging
+/// carry neither shape and keep the generic marker.
+priv enum MbtxFailureStage {
+  BuildError
+  RuntimeError
+}
+
+///|
+fn mbtx_failure_stage(
+  name : String,
+  brief : String?,
+  is_error : Bool,
+) -> MbtxFailureStage? {
+  guard is_error && name == "mbtx" && brief is Some(brief) else { return None }
+  if brief.has_prefix("mbtx (build") {
+    Some(BuildError)
+  } else if brief.has_prefix("mbtx (exit=") {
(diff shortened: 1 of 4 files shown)


bobzhang:
## codex CLI (subal) xhigh review record — 2 rounds

| Round | Finding | Disposition |
|---|---|---|
| 1 | **P2** the runtime badge treated every errored `mbtx (exit=N)` brief as a run-stage claim, but a single-shot target (js/llvm/wasm-gc) reports one exit code covering both stages — a current-engine js compile error would have read as "runtime error" | **Fixed** in `8fa9fa95c`: the runtime badge additionally requires the recorded call's target to be the staged wasm path (absent or `"wasm"`), read from the arguments the card already decodes; single-shot and unreadable payloads keep the generic `failed`. New js-target negative test |
| 2 (verification) | **P2** a WASM failure recorded before the staged engine shares the `mbtx (exit=N)` brief whichever stage failed, so an old compile error reads as "runtime error" | **Declined — explicit owner decision**: backwards compatibility for pre-staging transcripts was waived in the design discussion ("for the old sessions, we can just ignore it — no need to worry about backwards compatibility"). Disambiguating would need a new engine brief shape or a recording-version gate, spent entirely on sessions predating the label. The docstring now states this accepted imprecision plainly (`6140662ad`) instead of claiming old transcripts are unaffected |

Reviewer also ran the full local gate itself (`just check`: native + js deny-warn + `moon fmt --check`) — green. 462/462 desktop frontend js tests.

PR #1170, opened 2026-09-01
feat(viz): name mbtx failure stages and add a build-errors-only filter

bobzhang:
Follow-up 2 of 2 to the staged mbtx engine (#1167); desktop counterpart: #1168.

## Rendering

A failed `mbtx` call row in the HTML session render names its **failure stage**:

- **`build error`** (amber, new `--warn` token in all three theme palettes) — the staged engine's `mbtx (build …)` briefs. The paired result card is tagged `build-failed` and its collapsed flag reads `🚩 build error` — derived from fields the durable `ToolResult` itself carries (`tool_name`/`is_error`/`brief`), so both view modes work with zero extra map threading and the model view stays inside the `model_shown_result` projection invariant.
- **`runtime error`** — exit briefs on the staged wasm path only (target absent or `"wasm"` in the call's own durable arguments). A single-shot target's one exit code covers both stages and claims neither, matching the engine's contract and #1168's gating.

## Filter: find compile errors

A **"Build errors only"** toolbar — count (`🔨 N build errors`), toggle, viewport-relative prev/next jumps — cloned from the escalated-only mechanism and mutually exclusive with the other two filters. The stylesheet keeps each tagged call, its unambiguously paired result card, and their turns; the count comes from the same per-mode brief/failure maps the cards render from (`rendered_build_error_count`), so the toolbar and the tags cannot disagree. Scoped to build errors only, per the design discussion.

Pre-staging wasm recordings share the exit-brief shape and read as runtime errors — the same accepted owner decision as #1168 (no backwards compatibility for old transcripts).

## Notes

- `viz/pkg.generated.mbti` is updated by hand in canonical style: `moon info` skips the js-only `viz` package at the native canonical backend (pre-existing repo condition; the file's history shows the same maintenance path).
- 64/64 `viz` + 33/33 `cmd/viz_app` tests (new: stage tags + words in both view modes with a js-target negative case, per-mode count, toolbar mirror/visibility, three-way filter exclusivity, view-class); `moon check --deny-warn` and `moon fmt --check` clean; `moon build cmd/viz_app --target js` (the bundle CI builds) green.

### DIFF
--- a/cmd/viz_app/app.mbt
+++ b/cmd/viz_app/app.mbt
@@ -498,7 +498,10 @@ fn update(emit : @cmd.Emit[Msg], msg : Msg, model : Model) -> (@cmd.Cmd, Model) 
       } else {
         (
           @cmd.effect(() => {
-            scroll_to_card_from_viewport(".tool-call-build-failed", direction)
+            // Result cards, not call rows: every build failure has exactly
+            // one tagged result card — orphans included — so the jump set
+            // matches the count.
+            scroll_to_card_from_viewport(".card.build-failed", direction)
           }),
           model,
         )
--- a/cmd/viz_app/app.mbt
+++ b/cmd/viz_app/app.mbt
@@ -48,6 +48,9 @@ struct Model {
   // "Escalated only" keeps tool calls whose durable arguments explicitly carry
   // top-level `escalated: true`, plus their matching result cards.
   escalated_only : Bool
+  // "Build errors only" keeps mbtx calls whose result brief reports a
+  // BUILD-stage failure, plus their matching result cards.
+  build_errors_only : Bool
   // Tool-call arguments rendering: the human-friendly form (values shown verbatim,
   // no JSON quoting) by default, or the original pretty-printed JSON for debugging.
   show_original_args : Bool
@@ -102,6 +105,8 @@ enum Msg {
   JumpToError(Int)
   ToggleEscalatedOnly
   JumpToEscalated(Int)
+  ToggleBuildErrorsOnly
+  JumpToBuildError(Int)
   UnfoldCurrent
 }
 
@@ -185,6 +190,7 @@ fn init_model() -> Model {
     loading: false,
     status: "loading sessions…",
     errors_only: false,
+    build_errors_only: false,
     escalated_only: false,
     show_original_args: false,
     subrun_children: Map([]),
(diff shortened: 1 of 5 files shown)


bobzhang:
## codex CLI (subal) xhigh review record — 3 rounds

|---|---|---|
| 1 | **P2** orphan build-failed results rendered tagged cards but counted zero (toolbar hidden / active filter blanking real diagnostics); **P2** duplicated `tool_call_id`s made call-row stage labels depend on event order | Both **fixed** in `210bbca8e`: the count walks RESULTS — the same per-result test that tags their cards, both view modes — so count and visible tags cannot disagree, and jumps target result cards (the set every failure has exactly one of, orphans included); call-row labels are gated on the unambiguous 1:1 pairing already computed for the `call N` chips. Regression tests: orphan case; duplicate ids in both event orders |
| 2 | **P2** build classification keyed on two identities (call `name` vs result `tool_name`), divergeable on hand-edited logs; **P2** active toggle text hard-coded white on dark amber (~2.2:1) | Both **fixed** in `751ba749c`: both sides classify from the BRIEF alone (result data on both sides — tags and count cannot diverge; regression test with a name-mismatched pair); toggle uses theme-aware `var(--on-accent)` (~9.4:1 on the dark amber) |
| 3 (verification) | Four **P2**s, all concerning nested-subrun transcripts, ModelView duplicate-id projection, and sticky-toolbar anchor offsets | **Declined — parity with the existing filter family** (see below) |

### Round 3: considered and declined

Each finding describes behavior this filter **shares with the shipped errors-only and escalated-only filters**, not something this PR introduced:

1. *Subrun children not counted* — `rendered_error_count` and `rendered_escalated_count` also examine only the parent session while nested child transcripts render their own tagged cards. This PR mirrors the family's established scope.
2. *Ancestor cards of nested failures hidden* — the errors-only stylesheet (`.card:not(.error):not(.assistant)`) hides a non-error parent card containing a nested child's error card in exactly the same way.
3. *ModelView duplicate-id double counting* — the build count deliberately mirrors `rendered_error_count`'s ModelView shape (`model_shown_result` acceptance semantics); the duplicate-id-with-identical-content behavior is that machinery's, shared by error counting today.
4. *Sticky-toolbar anchor offsets* — the fixed 105px/110px offsets were already approximate with the two existing bars stacking and wrapping; this adds one more possible row to a pre-existing approximation.

Harmonizing all three filters for nested subruns (child-inclusive counts, a `:has()` ancestor exemption, toolbar-height-derived offsets) is a coherent follow-up if wanted — it belongs to the filter family, not to this PR.

### Final state

100/100 tests across `viz` + `cmd/viz_app`; `moon check --deny-warn` and `moon fmt --check` clean; `moon build cmd/viz_app --target js` (the bundle) green.

PR #1172, opened 2026-09-01
fix(desktop): stop repeating a build brief in its own failure marker

bobzhang:
Dogfooding feedback on #1168: a build-failure row read `mbtx (build failed, exit=1) build error` — the marker repeated the brief verbatim, since build-stage briefs already carry their literal state words. Build rows now append no marker; the stage rides the container class (`stage-build`), styled amber, with the words living in the brief itself so state never rests on color alone. Runtime rows keep the `runtime error` marker (`exit=1` alone does not name the stage); other tools keep `failed`. 457/457 desktop frontend js tests.

Viz counterpart (same de-duplication + filter discoverability) is a separate PR, per convention.

### DIFF
--- a/desktop/frontend/transcript/component/mbtx.mbt
+++ b/desktop/frontend/transcript/component/mbtx.mbt
@@ -131,10 +131,13 @@ fn mbtx_failure_stage(
 }
 
 ///|
-/// The literal word beside a failed call's brief. An `mbtx` failure names its
-/// STAGE: a build error is fixed in the source, a runtime error in the
-/// program's behavior, and the row should say which before the card is even
-/// opened. Every other failed call keeps the generic word.
+/// The literal word beside a failed call's brief — only where it ADDS one. A
+/// build-stage brief already reads "build failed" / "build timeout" in the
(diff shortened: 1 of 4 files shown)


bobzhang:
## codex CLI (subal) xhigh review record — 2 rounds

**Round 1** — one P2, accepted: the RESULT row's label is `Tool error · mbtx` (no brief), so dropping its marker left amber color as the only stage carrier — which matters exactly in parallel bursts, where result rows sit apart from their requests. Fixed: the marker is suppressed only where the row's label IS the brief (the request row, and the argument-less result shape); result rows keep the `build error` word. Net: the stage word appears exactly once per exchange, never adjacent to the brief.

**Round 2 (verification)** — clean, zero findings. Reviewer's own verdict: *“The change consistently suppresses redundant build-stage markers only when the brief is already displayed, while preserving explicit markers on separated result rows. Desktop JS checks and all 3,314 desktop JS tests pass.”*

(Housekeeping: GitHub dropped the second push's webhook — the head was amended to `cf3f34e4e` (content identical) to re-fire PR sync + CI.)

PR #1201, opened 2026-09-02
fix(desktop): track unread finishes by window visibility

Milky2018:
## Summary

- track native window focus and document visibility separately from the selected conversation
- mark a finished turn unread unless its transcript is actually visible
- clear the active conversation read state when the user returns to its visible chat
- keep the native badge derived from absolute per-conversation attention state

## Tests

- moon -C desktop test frontend --target js
- just check


Milky2018:
A red badge:

企业微信截图_07be19f9-357c-4280-b096-a36af9365b8e

PR #1208, opened 2026-09-02
workflow: standalone .mbtx example workflows (crash-resume showcase + live research)

bobzhang:
Two single-file `.mbtx` workflow scripts under `workflow/examples/`, runnable with plain `moon run <file>` — no project, no target flag; dependencies resolve from mooncakes (`bobzhang/workflow@0.1.1`), the engine is `openseek` on PATH.

**`showcase.mbtx`** (keyless — drives the modelless `echo` kind): generation 1 fans out 5 survey agents then simulates a crash; generation 2 replays the survey for free, reaches a 2-of-3 quorum over *identical* calls (multiset replay semantics), and ends a budget-bounded loop in a typed `CallBudgetExhausted` refusal:

```
=== GENERATION 1 ===
survey: 5/5 gathered (0 replayed, 5 live)
💥 simulating a crash before verification — run me again to resume
=== GENERATION 2 ===
survey: 5/5 gathered (5 replayed, 0 live)
verify: quorum reached with 3 votes
tail: allowance spent after 2 sweeps — typed refusal, not a crash
done: 5 replayed + 3 live on the main workflow, 30 fresh tokens
```

**`research.mbtx`** (live — needs a provider key in the environment): three model-driven explore scouts fan out over the repository in parallel. The verified run produced three correct, citation-dense answers about the workflow subsystem itself for 167k tokens; generation 2 replayed all three for `0 live scouts, 3 replayed, 0 fresh tokens`. The committed version was additionally verified against the recorded journal after de-hardcoding the repo path — replay identity is content, not paths, so the journal survived the change.

These are documentation-grade examples (they need an engine on PATH, so they are not wired into CI).

https://claude.ai/code/session_013t7W9RHYTLx5vcNyftZxeY


bobzhang:
## Output analysis: what each line of these runs actually demonstrates

### `showcase.mbtx` — reading the two generations mechanically

**`survey: 5/5 gathered (0 replayed, 5 live)` → `(5 replayed, 0 live)`** — the core durability claim, falsifiably stated by the counters: the resume generation re-executed the same program text, and every completed agent was served from the journal by content identity (kind + input + max_steps + scope), not by any checkpoint of control flow. There is no "resume point" — the program simply runs again and stops paying where it already paid.

**`💥 simulating a crash before verification`** — the crash lands *between* phases, so generation 2 also proves phase-boundary honesty: the journal has no notion of phases, and the survey replays purely because its five calls match, while verification runs live purely because its calls don't. Phases are display metadata; identity is the only durability mechanism.

**`verify: quorum reached with 3 votes`** — the three voters are *byte-identical* calls. That this works at all is the multiset-consumption semantics: three identical live calls appended three distinct journal entries; a replaying generation would claim them one-by-one in file order. Deduplicating them (the "obvious" cache design) would silently collapse independent samples into one — the exact bug the multiset design exists to prevent.

**`tail: allowance spent after 2 sweeps — typed refusal, not a crash`** — the loop asked for 10 sweeps against `max_calls=2`. The interesting property is *where* the boundary surfaces: as `Err(CallBudgetExhausted)` in the loop's own `match`, a value the script chose to handle — not an exception unwinding the workflow, not a silent truncation. Budget exhaustion is a policy outcome, and the type system made the script author decide its meaning.

**`done: 5 replayed + 3 live …, 30 fresh tokens`** — 30 = exactly 3 live echo children × 10 tokens each. Replayed work charges zero *by construction* (historical spend lives in the journal entries themselves), so `tokens_spent()` is a clean measure of marginal cost. The two ledgers never mix.

### `research.mbtx` — the live run

**Three parallel model-driven scouts, 167,015 tokens, three correct answers.** Accuracy was checked against the source: the torn-tail answer correctly identifies the byte-level split before UTF-8 decode, the `{"v":1,"e":…}` rewrite, and the atomic rename (`workflow/journal.mbt:102–139`); the envelope answer correctly names all five fields, the builder, the unwrapping site, and both the bare-input fallback and the loud unknown-major rejection (`cmd/openseek/subrun.mbt:45–54`); the worker answer correctly states the two-gate success rule including the `NoChanges`-disclaimed-by-`"failed"` nuance (`agent_workflow/worker_runner.mbt:265–314`). Citation density (9/5/7) is a scout-contract property surviving the workflow layer untouched — the orchestrator never parsed those answers, only transported them.

**`0 live scouts, 3 replayed, 0 fresh tokens`** — a 167k-token research artifact re-served for free. The economics generalize: for any workflow whose per-run cost is C, iterating on *downstream* logic (rendering, synthesis, thresholds) costs 0·C per iteration, because edits below the agent calls never disturb their identity. Only editing a prompt re-buys exactly that prompt's agents — invalidation is per-call, not per-run.

**An incidental but load-bearing proof**: the committed script differs from the originally-run one (the repo path was de-hardcoded; children now inherit cwd) — and the recorded journal still replayed. Identity is content, not paths or environment, which is what makes journals portable across checkouts and script refactors.

### The meta-observation

Both artifacts are single files whose full dependency closure is `moonbitlang/async` + `bobzhang/workflow` from the public registry plus one binary on PATH, run by plain `moon run <file>` (wasm by default — the orchestrator compiles to wasm and spawns native engine children). And the research run is pleasantly self-referential: the system's first real research task was explaining its own internals, and it got them right.

bobzhang:
## Two more workflows, and what their verified runs proved

### `debate.mbtx` — structured disagreement with data-dependent replay

Two advocate scouts argued opposite sides of a real design question — *should `agent_subrun` remain a wrapper over `workflow/spawn`, or fold into it?* — in parallel, from code evidence; a judge whose input **embeds both cases** then verified their claims and ruled.

**The ruling earned its keep.** The judge independently discovered a constraint neither advocate had raised: folding is *impossible as written* because `spawn` would need `agent_tool`/`protocol/emit` — packages of the openseek module, which imports `bobzhang/workflow` — a hard moon-module cycle (and a `native+wasm`-killing import besides). It also noticed `contract_runner` deliberately lacks a `parse_report` seam, cited both test suites as pinning distinct contracts, and ruled `VERDICT: KEEP` — independently converging with the human architecture review that made the same call. When the FOLD brief was then strengthened ("address module-dependency constraints explicitly"), the judge held KEEP, re-pricing the residual duplication as "the price of a boundary the module graph makes mandatory."

**The six-entry journal is the whole story in miniature:**
```
1 advocate:fold  Finished        ← round 1, parallel
2 advocate:keep  Finished
3 judge          DidNotFinish    ← REAL transient failure (connection reset), recorded losslessly
4 judge          Finished        ← resume: failures re-attempt, advocates replayed
5 advocate:fold  Finished        ← the CASCADE: one brief edited → exactly this re-ran
6 judge          Finished        ← …and the judge, whose input embeds fold's answer
```
No fifth entry for `advocate:keep` — it replayed through everything. This is the property no static-DAG system expresses: the invalidation frontier is computed by *data flow* (the judge's prompt textually embeds the advocates' answers), not by an edge list anyone maintains. And a final unchanged run: `0 live, 3 replayed, 0 fresh tokens` for a ~1.2M-token debate.

### `doc-audit.mbtx` — the map-reduce that found a real bug

One scout per package doc, auditing prose claims against code. Result: **1/4 drifted** — and the drift was real: `agent_subrun/README.mbt.md` still attributed "spawn/drain/deadline/teardown" to `run_subrun`, stale since commit 794e8dea delegated that machinery to `workflow/spawn`. The audit even reconstructed the history: *"last edited 2026-07-22 … never updated through that refactor."* The fix ships in this PR — the workflow's first productive output was a documentation bug in the very refactor that made the workflow possible.

The three NO-DRIFT verdicts were not rubber stamps: the `agent_workflow` audit re-verified the per-kind deadline table against three different source files, the `max_workers=4` engine parity, the byte-identical README copies, and that `workflow/moon.pkg` imports no openseek package — then confirmed via `git log` that no audited code had changed since the doc was written.

### Operational notes from the runs

- Both workflows survived tool-level interruptions (a too-short harness timeout, a genuine `ECONNRESET`) with zero lost work — resuming was just re-running the same command. Durability wasn't demonstrated in a contrived test; it carried the actual development session.
- Cost profile: debate ≈ 1.2M tokens live, doc-audit ≈ 350k — and every re-run after completion is 0. Iterating on the reduce/report code re-spends nothing, so report formatting was tuned against journaled results for free.

bobzhang:
## Round 3: the hunt and the read→write pipeline — outputs and lessons

### `todo-hunt.mbtx` — loop-until-dry, verified

Six rounds, six distinct real deferred-work items, zero repeats — the seed-list steering (each round's prompt embeds everything found) worked perfectly. Five items came from documented known-limits; the sixth is the interesting one: **round 3 escaped the docs entirely** and surfaced a code-level deferral no README mentions (`agent_subtask/provision.mbt:108` — submodule-gitlink slices refused as v1-unsupported). The loop hit its round cap without going dry: the backlog is genuinely ≥6 deep. Cost: 1.12M tokens live — and the replay generation reconstructed the **entire loop, control flow included**, for `0 live, 6 replayed, 0 fresh tokens`. That last point deserves emphasis: the `while dry < 2` termination decisions replayed deterministically because they are a pure function of the recorded outcomes — durability without deterministic agents, demonstrated on a loop.

### `agent_workflow/fixit` — the first read→write workflow, and what it took

The verified run: a scout found that the README's `Running`-state limit **conflates two cancellation mechanisms with opposite outcomes** (deadline-cancel reaches git capture as `TimedOut`, so edits are salvaged and `continue` applies; external cancel re-raises *before* capture, hence `Running`/discard-only) — a correct and non-obvious distinction. A confined worker slice then committed a clarification in its own worktree: `state Committed`, `+10/-5`, commit captured from git evidence, slice discarded per the demo contract. The same precision now lands in the README directly — the pipeline's second productive documentation fix.

**The honest part — it took three brief redesigns, each taught by a typed failure.** Attempts 1–2 asked the scout to *hunt for a defect* and died at `MaxSteps` (236k tokens of diligent searching): after the doc-audit round fixed the real drift, clear defects were scarce, and a brief that presupposes a needle sends a thorough scout combing the haystack to exhaustion. The completable shape — *verify one specific claim, draft a sharper wording* — succeeded immediately. Two transferable lessons: **prompt design is failure-policy design** (give agents an honest exit), and open-ended search is the wrong scout shape when the expected hit-rate is low. Notably, the type system made this iteration cheap: every failure surfaced as one legible typed line with full cost accounting, never a hang or a mystery.

### One replay nuance the demo design exposed

`fixit` discards its slice to stay rerunnable — which means a rerun is **not** free: the replay validator correctly vetoes the journalled `Committed` outcome (its branch no longer exists) and the worker runs live again. Scouts still replay; only the stateful outcome re-pays. That's the validator doing exactly its job — a discarded branch *should not* replay as done — but it makes the trade explicit: `discard` = clean repo + re-pay on rerun; `integrate_slice` = landed fix + free replays. Saved workflows will want the latter.

Cumulative across all five workflow examples: ~3.7M live tokens spent, every completed token replayable for zero, three real documentation defects found and fixed in this repo by the workflows themselves.

PR #1217, opened 2026-09-02
workflow/viz: journal visualizer CLI (live watch, derived dataflow, outputs) + deep-research example

bobzhang:
Answers "can we visualize a dynamic workflow — including in real time?" with a shipped component and a live-verified example.

**`workflow/viz`** — a moonx-invokable CLI (`moonx --target native bobzhang/workflow/viz@latest journal.jsonl`, verified cold against the published 0.1.2):
- **Ledger**: every recorded attempt with status, steps, and token-cost bars.
- **Derived dataflow**: dynamic workflows declare no DAG, so none is stored — an edge is recovered wherever one agent's recorded output appears verbatim inside a later agent's input (the probe is escaped through the JSON encoder so it matches stringified inputs). On real journals this recovers the debate's advocates→judge fan-in and cascade, the hunt loop's triangular seed structure, fixit's scout→worker handoff, and correctly draws zero edges for a pure parallel map.
- **Outputs**: every agent's complete answer, expandable per agent.
- **`--watch`**: re-renders every 2s with page auto-refresh — a live view of a running workflow, verified by watching a run's ledger grow generation by generation. A safety property worth calling out: watch mode deliberately avoids `Journal::load`, whose torn-tail *repair* would rewrite a file the running workflow is mid-append on; viz uses its own read-only parser and skips the writer's torn tail for a tick.

**`workflow/examples/deep-research.mbtx`** — the showcase for both: three survey scouts in parallel, a probe whose prompt embeds their findings, a synthesis embedding everything — a research funnel whose edges appear in the visualizer as the run progresses. Its first live run (watched in real time in a browser) also exposed a genuine integration gap, now documented in the example: the request envelope *carries* `max_steps`, but the engine child *enforces* its ceiling from argv — so the launch seam must place the limit on the command line, and the engine honoring the envelope's value is follow-up work.

Also records the published 0.1.2 version in `moon.mod` (the #1208 squash predated the viz pushes; this PR restores that package as one coherent unit).

https://claude.ai/code/session_013t7W9RHYTLx5vcNyftZxeY

### DIFF
--- a/workflow/examples/deep-research.mbtx
+++ b/workflow/examples/deep-research.mbtx
@@ -1,0 +1,141 @@
+import {
+  "moonbitlang/async@0.21.1",
+  "bobzhang/workflow@0.1.2",
+  "bobzhang/workflow@0.1.2/spawn",
+}
+
+///|
+/// DEEP RESEARCH, watched live: three survey scouts fan out in
+/// parallel, a follow-up probe targets what the survey left open, and a
+/// synthesis agent — whose prompt embeds every prior answer — writes a
+/// design proposal. Each completion lands in the journal as it happens,
+/// so a concurrent `viz --watch` renders the run in real time, and the
+/// synthesis edges appear in the derived dataflow graph.
+///
+///   moon run workflow/viz -- deep-research.jsonl live.html --watch &
+///   moon run workflow/examples/deep-research.mbtx      (repo root)
+///
+/// Needs `openseek` on PATH and a provider key in the environment.
+async fn main {
+  research() catch {
+    error => println("workflow failed: \{error}")
+  }
+}
+
+///|
+let question : String = "How should the bobzhang/workflow library support human-in-the-loop waits (Golem-style promises: suspend until external input, at zero cost)?"
+
+///|
+async fn research() -> Unit {
+  let wf = @workflow.Workflow(
+    runner=@spawn.contract_runner(launch=call => {
+      // The step ceiling must ride ARGV: the envelope carries max_steps
+      // as data, but the engine child takes its enforced ceiling from
+      // the command line (a gap this example discovered the hard way —
+      // its first run's scouts sailed past their intended budgets).
+      let argv = ["subrun", call.kind]
+      if call.max_steps is Some(steps) {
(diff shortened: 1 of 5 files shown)


bobzhang:
## The deep-research run, end to end — watched live, tuned by resume, self-documenting

**Real-time visualization worked as designed.** The `--watch` sidecar's own log is the proof — the browser page grew as the run progressed: `0 → 1 → 2 → 3 entries` through the parallel survey, then `4 entries / 3 edges` as the probe's fan-in appeared, with the watcher hot-swapped mid-run (to add the Outputs panel) without disturbing the writer. Final state: **7 entries, 15 derived edges** — the full research funnel, recovered purely from content.

**Resume carried the whole tuning loop.** The run's history, all in one journal: three surveys and a probe completed live (at unenforced ceilings — the argv gap this PR documents — including one 2.7M-token scout); synthesis **TimedOut** at 51 steps against the wall clock; the fixed script re-ran it enforced and got a typed **MaxSteps at exactly 14**; raising just that ceiling to 28 changed only that call's identity, so the tuning rerun replayed all four research agents free and re-paid only synthesis. Final verification: `0 live, 5 replayed, 0 fresh tokens` for a ~10M-token research run.

**And the research output is real.** The synthesis produced a concrete design for human-in-the-loop waits (Golem-style promises) that respects the library's actual invariants — waits as workflow-level promise records above the Runner seam (never agent calls: no slot, no allowance, no fabricated zero-cost attempt lying to the ledger); `v2` journal line kinds for open/resolved waits that never enter `claim` (a wait is not a failure and must not be silently re-asked); a fifth `Suspended` disposition keeping the start/finish bracket invariant; a separate single-writer answer ledger since humans reply between process lifetimes; `wait_id` as a stable content hash so identical opens collapse across generations — and it caught that the torn-tail repair must learn to re-emit v2 lines or silently drop open promises. Every claim cited to file:line, and the ones I've checked are accurate. This is the working draft for the promises feature.

There's a pleasing closure here: the example workflow researched the very feature that would most improve example workflows like it — and its own failure modes (TimedOut, MaxSteps) were the empirical data that tuned it.

PR #1238, opened 2026-09-03
Split TUI into a dedicated openseek_tui binary

bobzhang:
## Summary

Splits the single `openseek` binary into two:

- **`openseek`** — headless automation CLI only: `run` / `serve` / `review` / `subrun` / `mcp` / `sessions`. A bare `openseek` is now rejected (no default action).
- **`openseek_tui`** — dedicated interactive terminal UI binary, built on the existing `cmd/tui` library.

## Changes

- New executable package `cmd/openseek_tui` (a thin wrapper that builds `cmd/tui`'s `cli_command` and calls `@tui.run`).
- `cmd/openseek` drops the `tui` subcommand and default-UI dispatch; `subcommand_required=true`.
- Shared agent options (`--api-key`, `--model`, …) and the `failure_message` sanitizer move to `internal/cli` so both mains reuse them (`agent_shared_options`, `failure_message`).
- `cmd/tui` exposes `pub cli_command(name?)`, and its default engine now resolves the sibling `openseek` CLI (rewriting a path `openseek_tui` → `openseek`, including the debug TCC response-file layout).
- Docs and cram tests updated: `tests/cram/tui.md` exercises `openseek_tui.exe`, and `tests/cram/cli.md` covers the new bare-`openseek` rejection.

## Verification

- `moon check` — clean
- `moon fmt --check` — clean
- `moon test internal/cli cmd/tui cmd/openseek` — 172 passed
- `moon cram test tests/cram` — 28 passed

Generated with SeekMoon

## Deployment note

The TUI is now a second executable. An install that ships only `openseek` (for example a hand-copied `~/.moon/bin/openseek`) keeps the headless CLI but loses the UI until `openseek_tui` is installed beside it. From the source tree, `moon run cmd/openseek_tui` builds only the UI; run `moon build cmd/openseek` first so the sibling engine exists (the preflight names this step when it is missing).

## Review follow-up commit

The second commit addresses the review: `default_engine` never spawns the UI itself (unrecognized argv0 falls back to PATH `openseek`; only the basename and Moon's package directory are rewritten), the preflight points at `moon build cmd/openseek`, the cram suite regains the end-to-end default-engine case, and the `cli_command` name parameter and stale comments are gone.


bobzhang (commented):
Reviewed as a mechanical split and verified locally on the branch head (96834130f): `moon fmt --check`, `moon check`, `moon test cmd/tui internal/cli cmd/openseek cmd/openseek_tui --target native` (172 passed), `moon cram test tests/cram` (28 passed). I also built both binaries (debug and release) and exercised the sibling-engine resolution by hand. The move itself is faithful: `agent_shared_options` / `failure_message` are verbatim relocations, dispatch/help/cram are updated consistently, and a repo-wide grep finds no stale `openseek tui` / `-- tui` references outside the PR. One real regression in the documented dev workflow, one coverage loss, then nits.

### 1. `moon run cmd/openseek_tui` (the README's documented launch) fails on a fresh checkout

`cmd/openseek_tui` no longer depends on `cmd/openseek`, so `moon run cmd/openseek_tui` builds only the TUI. `default_engine` then rewrites argv0 `…/_build/native/debug/build/bobzhang/openseek/cmd/openseek_tui/openseek_tui.exe` to `…/cmd/openseek/openseek.exe`, which does not exist yet, and the preflight refuses:

```
$ DEEPSEEK=test-key moon run cmd/openseek_tui
error: engine '…/_build/native/debug/build/bobzhang/openseek/cmd/openseek/openseek.exe' is not usable: it must be on PATH, executable, and accept `--help` (exit 0) the way openseek does.
Pass --engine <path> or install the openseek binary.
```

After `moon build cmd/openseek` the same command reaches the TTY guard as expected. Before this PR `moon run cmd/openseek` re-launched itself, so this worked from a clean tree. (Note: on this toolchain the debug layout is a real `.exe`, not the tcc `.c` path the new unit test models, so it is the plain-path branch of `default_engine` that bites, not the `.rspfile` one.)

Suggested, in order of cost:
- README.mbt.md:227 / :236 / :237 and cmd/tui/README.md: say `moon build cmd/openseek` (or plain `moon build`) before `moon run cmd/openseek_tui`. This alone closes the regression.
- Optionally, in `run`, when `--engine` is unset and the derived sibling path fails the probe, retry with the PATH lookup `("openseek", [])` before erroring; or at least mention `moon build cmd/openseek` in the error when the derived path sits under `_build/`.

### 2. Coverage lost for the default-engine handoff

The removed cram section "Bare `openseek` Is The Interactive UI" was the only end-to-end check that the default engine resolves from PATH and the preflight passes through to the TTY guard. Nothing in `tests/cram/tui.md` now runs `openseek_tui.exe` without `--engine`; every remaining case either passes `--engine` or fails earlier on the missing key. The old shape still works against the new binary (verified locally with the release build):

```mooncram
$ sh <<'EOF'
> bin=$(mktemp -d)
> ln -s "$(command -v openseek.exe)" "$bin/openseek"
> PATH="$bin:$PATH" env DEEPSEEK=test-key openseek_tui.exe 2>&1
> rm -rf "$bin"
> EOF
error: the interactive UI needs a terminal; run a headless task with `openseek run "…"` (or `openseek serve` for the JSONL protocol).
```

Worth re-adding to `tui.md` so the `openseek_tui` → PATH `openseek` handoff stays pinned; it is the one contract this PR introduces.

### 3. Self-as-engine is now a silent misconfiguration (minor, can defer)

Pre-split, "spawn argv0" was always right. Now if the TUI binary's path lacks the `openseek_tui` substring (a renamed or differently-named symlink), `default_engine` returns the TUI itself. The `--help` probe cannot tell (`openseek_tui --help` and even `openseek_tui serve --help` both exit 0, since `--help` short-circuits), so the failure only surfaces when the UI spawns `<tui> serve`:

```
$ ./ostui serve
error: unexpected value 'serve' found; no more were expected
```

Reproduced with a copy named `ostui`. Conversely, `replace_all` also rewrites directory components: a binary at `…/openseek_tui/openseek_tui.exe` derives `…/openseek/openseek.exe` (also reproduced; preflight fails). Cheapest hardening in `default_engine`: only derive from argv0 when the basename itself carries `openseek_tui` (plus the `cmd/openseek_tui/` package dir for the tcc layout), and otherwise fall back to PATH `openseek` rather than ever returning argv0 unchanged.

### 4. Nits

- Stale "re-launch this command / ourselves" wording in cmd/tui/main.mbt:155, :168, :212-213, :394, and the test name at :553 ("default_engine re-launches a path argv0"). The TUI no longer re-launches itself; the `/usr/local/bin/openseek` cases in that test now describe "a binary that happens to be named openseek".
- cmd/tui/main.mbt:44 `cli_command(name? = "tui")`: the parameter exists only so tests keep the old name. Simpler to fix the name to `openseek_tui` and drop the parameter; no test depends on it.
- cmd/openseek/main.mbt:554 "bare parse is rejected": asserting only `outcome != "parsed"` also passes if the parse fails for an unrelated reason (e.g. a root option ever becoming required). Assert the message mentions `subcommand`, as the cram case already does.
- Deployment note: anyone with only `openseek` installed (e.g. `~/.moon/bin/openseek`) silently loses the TUI until they also install `openseek_tui`. Worth a line in the PR description or release note.
- docs/architecture.md:12 lists `run`/`serve`/`review`/`sessions`/`mcp` and omits `subrun`; the other docs in this PR include it.

Verdict: LGTM as a mechanical refactor once #1 is addressed (the doc fix alone is enough; the PATH fallback is nice-to-have). #2 is cheap and belongs in this PR; #3 and #4 can follow up.

bobzhang:
Review items addressed in 2a93dc414 (separate commit): `default_engine` never spawns the UI itself (unrecognized argv0 falls back to PATH `openseek`; only the basename plus Moon's `cmd/openseek_tui/` package dir are rewritten), the preflight names `moon build cmd/openseek` when the sibling is missing from the build tree, README/cmd/tui docs say to build the engine first, `tests/cram/tui.md` regains the end-to-end default-engine case, `cli_command` drops its name parameter, the bare-`openseek` test asserts the missing-subcommand reason, stale comments and the architecture subcommand list updated. Verified locally: `moon check --deny-warn`, `moon fmt`, 172 unit tests, 29 cram cases, plus the fresh-tree / renamed-copy / `openseek_tui`-named-directory launches by hand. Merging once CI is green.

PR #1287, opened 2026-09-04
agent_workflow/snippet: the seam that makes a snippet's workflow visible

bobzhang:
Stacked on #1286.

`moonbitlang/workflow` is generic: it spawns whatever its `Runner` spawns, writes its journal wherever it was told, and never learns openseek exists. So **nothing inside a snippet can make its own run discoverable** — a snippet that hand-rolled `@shell.Cmd(openseek, ["subrun", ...])` produces children with no transcript and no ledger, invisible to every viewer.

This package is the openseek half. It reads coordinates mbtx reserved *before* the snippet started (a block of child ordinals, a journal path, a sidecar path) out of the guest environment, and turns them into a `Runner` that gives every child the durable session `<parent>-sr-N`.

## Three load-bearing properties

**`attempt_id` is the child's session suffix, not an opaque id.** A journal row therefore names `<parent>-sr-N` given the parent — the same string `viz`'s chip links to and `desktop/internal/subrun/probe.mbt` tails. That single field is the whole join between the workflow view and the per-subagent view, and it costs nothing because the runner had to mint the ordinal anyway. (Contrast `@spawn.contract_runner`, which mints `cr-N` and cannot be joined to anything.)

**The reserved block is the launch ceiling.** A call past its end is refused as `Skipped` rather than launched unnumbered: a child with no transcript is exactly what this package exists to prevent.

**Sidecar lines bracket every launch.** The journal cannot do this — a `JournalEntry` carries an outcome, so it is written when a call *resolves*, and a reader with only the journal learns about a child no earlier than its completion.

`run` is the only entry point a snippet calls, so the author writes the shape of the work and nothing else. Teaching the model to pass these itself would put the coordinates that make a run visible into the same text that decides what the run does, where a plausible-looking wrong value silently produces an invisible workflow.

## Verification

5/5 new tests, covering ordinal allocation from the block, `--session` injection, block-as-ceiling refusal, sidecar bracket contents, and `attempt_id` in a real file-backed journal.

## Known constraint

`.mbtx` front-matter imports resolve `bobzhang/openseek/...` against the **published** module in the registry cache, not the local working tree — so this package is not importable from a snippet until openseek is released. See the stack description for the two options.

### DIFF
--- a/agent_workflow/snippet/injected.mbt
+++ b/agent_workflow/snippet/injected.mbt
@@ -1,0 +1,163 @@
+// The engine-side coordinates an `mbtx` snippet needs to run a workflow whose
+// children are visible to the rest of openseek.
+//
+// `moonbitlang/workflow` is a generic library: it spawns whatever its `Runner`
+// spawns and writes its journal wherever it was told, and it never learns
+// openseek exists. So nothing IN the snippet can make a run discoverable —
+// the coordinates have to come from outside, from the one component that
+// knows both the session and the snippet: the `mbtx` tool that launched it.
+//
+// mbtx picks them BEFORE the snippet starts (a reserved block of child
+// ordinals, a journal path, a sidecar path) and injects them into the guest
+// environment through the wasm policy's `env.set`. Discovery is therefore not
+// discovery at all: the host already knows where to look, the same way it
+// already knows a sub-run started because the engine told it.
+//
+// A snippet reads them through `injected()` and hands the result to `run`.
+// The model never types a session id, a path, or an ordinal — it cannot, since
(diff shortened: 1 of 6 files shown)


bobzhang:
Superseded. This helper could never be imported from a snippet — `.mbtx` front-matter resolves `bobzhang/openseek/*` against the *published* module, so an unreleased package here is invisible to the very snippets it exists for. It now lives where a snippet can reach it: `moonbitlang/workflow/hosted`, published in workflow 0.7.0 (moonbitlang/workflow#21). #1288 consumes it; `agent_workflow/snippet` is gone from the tree.

PR #1298, opened 2026-09-05
refactor: cmd/openseek, mcp: dedupe tests and share the subrun argv builder

bobzhang:

Candidates came from a read-only survey of the whole repository. The author was told to drop any that would not compile or would change behaviour, so this branch carries fewer than were proposed: **8 applied, 1 dropped, net -88 lines.**

## Applied

- **cmd/openseek/serve.mbt** — Seven near-identical should_auto_continue assertions collapse into one flip-one-flag table
  Applied, but NOT as the proposed tuple table. MoonBit local `fn` inside a test body does not support labelled/defaulted parameters — the proposed form failed to compile (19 errors: "This function has no parameter with label goal_standing~"). Instead added a top-level private helper `auto_continues` with every input defaulted to its continuing value, mirroring the existing test-helper precedent `ApprovalDesk::standing` in cmd/openseek/approval.mbt. Each case stays its own assert statement with its own source location, so a failure is still pinpointed by line — better diagnosability than the proposed for-loop table. 7 assertions before, 7 after; -50 lines.
- **mcp/client_test.mbt** — Nine MCP client tests repeat the same transport/pump/server wiring preamble
  Applied as proposed. `fake_client(group, reply)` added after `spawn_server`; all nine async tests rewritten, each keeping its own reply/fixture hoisted above the call. `spawn_client` still runs before `spawn_server` in the same order, and the reply closures are pure, so no side-effect order changed. Test count unchanged (23 in the package).
- **cmd/openseek/approval.mbt** — Five copies of the same spawn-registration spin-wait in ApprovalDesk tests
  Applied as proposed. `await_standing(desk)` added next to `ApprovalDesk::standing`; the load-bearing comment moved from the first copy onto the helper. Three of the five call sites spun on `desk.pending.length() > 0` and now spin on `desk.standing() is Some(_)` — same predicate, since `standing()` returns the first key of `pending` or None (its body is 10 lines above the helper).
- **cmd/openseek/subrun.mbt** — Three copies of the subrun argv builder, including the --api-url rule
  Helper added here rather than in tools.mbt as proposed: subrun.mbt's own doc comment already states the argv contract ("model, endpoint, and the step ceiling ride argv"), so the parent-side builder sits with the child-side reader. Call sites rewritten in gate.mbt, review_tool.mbt and subtask_tool.mbt.
- **cmd/openseek/gate.mbt** — The review subrun's input JSON is built identically in review_tool.mbt and gate.mbt
  Applied as proposed. `review_subrun_input` added beside `parse_audit_report`, the existing shared home for review-subrun helpers; both call sites rewritten. The explicit `: Json` annotation is dropped at the call sites because the helper's return type supplies it.
- **mcp/types.mbt** — render_text's first-flag StringBuilder loop is map + join("\n")
  Applied as proposed. `join` writes exactly `\n` between elements and "" for an empty array, matching the first-flag loop. Everything after `let text = ...` is untouched. moon fmt wrapped the match in a block body.
- **mcp/types.mbt** — decode_call_result's push loop is Array::map, the way decode_tools_page already writes it
  Applied as proposed. `Array::map` is error-polymorphic, so `decode_content_block`'s raise propagates exactly as the loop's did.
- **mcp/tools/tools.mbt** — build's regroup-and-append loop is map + Array::flatten
  Applied as proposed. `build_grouped`'s own doc comment already promises "Flattening the groups yields exactly `build`'s registry", so the new code says what the comment said.

## Dropped, and why

- **Four render_text/structuredContent tests differ only in content and expectation (mcp/types_wbtest.mbt:137)**
  Dropped on the brief's test-collapse rules. (1) The four test NAMES are the documentation of render_text's structuredContent contract — they mirror its doc comment clause by clause ("renders structuredContent when content is empty", "does not duplicate structuredContent already in content", "appends structuredContent distinct from a status message"); one table named for the whole contract loses three of those. (2) The proposed first row, "render_text is empty for empty content", is not about structuredContent at all — it is the empty-content pin, and it is precisely what makes the map+join rewrite in mcp/types.mbt safe. Folding the pin into a structuredContent table hides that relationship right when it started mattering. (3) A `for`-loop of `assert_eq` turns four precisely-located failures into one; `assert_eq` does take `msg?`, so a row label was possible, but the label would only restore what the four test names already gave for free. 18 lines is not worth that trade. The four tests are untouched.

## Verification

Toolchain: moon 0.1.20260827 (d0aaa07). All commands run in the worktree at .../scratchpad/wt/cli.

TESTS BEFORE (baseline on clean simplify/cli):
  moon test -p bobzhang/openseek/cmd/openseek  -> Total tests: 82, passed: 82, failed: 0
  moon test -p bobzhang/openseek/mcp           -> Total tests: 23, passed: 23, failed: 0
  moon test -p bobzhang/openseek/mcp/tools     -> Total tests:  7, passed:  7, failed: 0
  (combined run of the three: Total tests: 112, passed: 112, failed: 0)

TESTS AFTER (identical counts — no test was added or removed):
  cmd/openseek 82/82, mcp 23/23, mcp/tools 7/7.

ASSERTION COUNTS for the one test whose body changed
(cmd/openseek/serve.mbt "auto-continue requires a finished idle serving turn with budget"):
  before: 1 assert_true + 6 assert_false = 7 assertions across 73 lines.
  after:  1 assert_true + 6 assert_false = 7 assertions across 9 lines, plus a 17-line
  defaulted helper. Same six single-condition flips (auto, finished, idle, shutting_down,
  goal_standing, turns_remaining), same order, same expectations. NOT collapsed into a
  table row loop — each case stayed a separate assert statement, so failures keep their
  own source locations.

FORMAT:
  moon fmt -> "Finished. moon: ran 4 tasks, now up to date". git status --short after fmt
  listed only my own files; no unrelated file was reformatted.
  (An earlier repo-wide `moon fmt` also touched nothing outside my files — verified with
  git status immediately after.)

CHECK (--deny-warn, both targets; all three packages are supported_targets = "+wasm+native"):
  moon check -p ./cmd/openseek --target native --deny-warn -> Finished, 0 warnings/errors
  moon check -p ./mcp          --target native --deny-warn -> Finished
  moon check -p ./mcp/tools    --target native --deny-warn -> Finished
  moon check -p ./cmd/openseek --target wasm   --deny-warn -> Finished (107 tasks)
  moon check -p ./mcp          --target wasm   --deny-warn -> Finished
  moon check -p ./mcp/tools    --target wasm   --deny-warn -> Finished

BUILD (to confirm the new test-only top-level helpers are not dead-code warnings in a
non-test build):
  moon build ./cmd/openseek --target native --deny-warn -> Finished, 155 tasks.
  (One pre-existing, unrelated Xcode libtool warning about an empty
  moonbitlang/async/types archive; present on the clean baseline too.)

MOON INFO (scoped, never bare):
  moon info -p bobzhang/openseek/cmd/openseek -> Finished
  moon info -p bobzhang/openseek/mcp          -> Finished
  moon info -p bobzhang/openseek/mcp/tools    -> Finished
  git status --short afterwards showed the SAME nine .mbt files and no .mbti — the public
  interface of all three packages is unchanged.

WORKAROUNDS / FRICTION worth naming:
  - `moon check -p <full/package/name>` fails here ("Failed to canonicalize input filter
    directory"); check wants a relative path (./mcp). `moon test -p` and `moon info -p`
    want the full package name (bobzhang/openseek/mcp). `moon check` also rejects more
    than one -p, so I looped instead of batching.
  - The finding-1 proposal did not compile: a local `fn` inside a test body degrades to a
    positional lambda, so labelled defaults are rejected. Reworked to a top-level helper
    (see `applied`).

COMMIT/PUSH:
  git add <nine explicit paths>; git commit -> 4a27cf8f8, 9 files changed,
  160 insertions(+), 248 deletions(-). git push -u origin simplify/cli -> new branch
  created on origin. git status --short is clean. No PR opened, no merge, no other branch
  touched.

## Worth a second look

Three things deserve a second look.

1. cmd/openseek/approval.mbt — the only place I changed a predicate rather than just moving code. Three of the five spin-waits tested `desk.pending.length() > 0` and now test `desk.standing() is Some(_)` via the shared helper. These are the same predicate because `ApprovalDesk::standing` (10 lines above the new helper) returns the first key of `self.pending`, or None when it is empty. If you disagree that this is obvious enough at the call sites, the alternative is two helpers instead of one.

2. cmd/openseek/serve.mbt and cmd/openseek/approval.mbt each gain a top-level private `fn` used only from a `test` block in the same non-test file (`auto_continues`, `await_standing`). This follows the existing `ApprovalDesk::standing` precedent in the repo, and I confirmed it does not trip deny-warn on `moon check` (native + wasm) or on `moon build --target native`. Still, it is a pattern worth a deliberate nod rather than an accident.

3. cmd/openseek/subrun.mbt is a file about the CHILD subrun mode, and `subrun_args` is parent-side code. I put it there anyway because that file's doc comment is where the argv contract is written down ("model, endpoint, and the step ceiling ride argv"), so the producer now sits beside the spec its consumer implements. If the team would rather keep the file strictly child-side, moving the helper to tools.mbt (as the original finding proposed) is a one-block move with no other edits.

Minor: in mcp/tools/tools.mbt, `build` is now an expression body ending in `.flatten()` with no intermediate binding — check the `.map(group => group.1)` reads clearly enough versus the previous `let (_, contributed) = entry` destructuring.

Nothing here changes a public signature: `moon info` on all three packages left every .mbti byte-identical.

### DIFF
--- a/cmd/openseek/approval.mbt
+++ b/cmd/openseek/approval.mbt
@@ -295,6 +295,19 @@ fn ApprovalDesk::standing(self : ApprovalDesk) -> String? {
 }
 
 ///|
+/// Test helper: block until the asker has registered its pending entry. It
+/// registers before it blocks, so wait for that rather than assuming the
+/// spawn already ran.
+async fn await_standing(desk : ApprovalDesk) -> Unit {
+  for _ in 0..<200 {
+    if desk.standing() is Some(_) {
+      break
+    }
+    @async.sleep(5)
+  }
+}
+
+///|
 async test "an answered question resolves to the controller's decision" {
   let desk = ApprovalDesk::ApprovalDesk(now_ms=0)
   @async.with_task_group(group => {
@@ -305,14 +318,7 @@ async test "an answered question resolves to the controller's decision" {
         body: Some("fn main { println(1) }"),
       })
     })
-    // The asker registers its pending entry before it blocks; wait for it
-    // rather than assuming the spawn already ran.
-    for _ in 0..<200 {
-      if desk.standing() is Some(_) {
-        break
-      }
-      @async.sleep(5)
-    }
+    await_standing(desk)
     guard desk.standing() is Some(id) else { fail("no question was announced") }
     desk.settle(id, true)
     assert_true(asked.wait() is AllowedOnce)
@@ -331,12 +337,7 @@ async test "a refused question resolves to rejected" {
         body: Some("fn main { println(1) }"),
       })
     })
-    for _ in 0..<200 {
-      if desk.standing() is Some(_) {
-        break
-      }
-      @async.sleep(5)
-    }
+    await_standing(desk)
     guard desk.standing() is Some(id) else { fail("no question was announced") }
     desk.settle(id, false)
     assert_true(asked.wait() is Rejected)
@@ -354,12 +355,7 @@ async test "a decision for a settled question is discarded" {
     let asking = group.spawn(() => {
       desk.ask({ tool_name: "mbtx", detail: "escalate", body: None, })
     })
-    for _ in 0..<200 {
-      if desk.pending.length() > 0 {
-        break
-      }
-      @async.sleep(5)
-    }
+    await_standing(desk)
     asking.cancel()
   })
   assert_eq(desk.pending.length(), 0)
@@ -398,12 +394,7 @@ async test "an interrupted wait still retires its question" {
         body: Some("fn main { println(1) }"),
       })
     })
-    for _ in 0..<200 {
-      if desk.pending.length() > 0 {
-        break
-      }
-      @async.sleep(5)
-    }
+    await_standing(desk)
     // The pending entry proves a tool is blocked here right now.
     assert_eq(desk.pending.length(), 1)
     asking.cancel()
@@ -426,12 +417,7 @@ async test "closing the command stream withdraws a standing question" {
     let asking = group.spawn(() => {
       desk.ask({ tool_name: "mbtx", detail: "escalate", body: None, })
     })
-    for _ in 0..<200 {
-      if desk.pending.length() > 0 {
-        break
-      }
-      @async.sleep(5)
-    }
+    await_standing(desk)
     desk.withdraw_all()
     assert_true(asking.wait() is Cancelled)
   })
--- a/cmd/openseek/gate.mbt
+++ b/cmd/openseek/gate.mbt
@@ -33,6 +33,20 @@ fn parse_audit_report(
 }
 
 ///|
+/// The input every `subrun review` child reads: the criteria to audit
+/// against, plus the baseline commit when the goal captured one.
+fn review_subrun_input(
+  goal : String,
+  baseline : @agent_session.GoalBaseline?,
+) -> Json {
+  match baseline {
+    Some(baseline) =>
+      { "goal": goal, "sha": baseline.sha, "dirty": baseline.dirty }
+    None => { "goal": goal }
+  }
+}
+
+///|
 /// Build the advisory goal-met gate: on a durably recorded met, spawn the
 /// `subrun review` child against the goal's baseline and return the bounded
 /// digest notice. Advisory by contract — every failure shape (no report,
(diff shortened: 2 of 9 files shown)


bobzhang:

> The worktree (HEAD 4a27cf8f8, clean) is a pure simplification of cmd/openseek, mcp, and mcp/tools and is clean against all seven criteria; no blockers and no findings.
> 
> Evidence: (1) Every hunk was diffed against a227f6aae and read in both old and new form. serve.mbt's 7-assertion test collapses into an auto_continues() helper whose defaults reproduce the old parameters exactly and whose six assert_false calls flip the same single conditions; approval.mbt's five spin-waits collapse into await_standing, equivalent to both original predicates (pending.length()>0 and standing() is Some(_)); client_test.mbt's nine tests keep identical names/replies with only the transport/rpc/pump/server preamble moved into fake_client in the same order; decode_call_result's push loop and render_text's StringBuilder/first-flag loop in mcp/types.mbt are exactly reproduced by blocks.map(...) and map+join("\n") including empty-content and embedded-newline corners; tools.mbt build's append loop is exactly .map(group => group.1).flatten(); the shared subrun_args/review_subrun_input helpers reproduce the argv and JSON byte-for-byte at all three former call sites, including the --api-url-when-Some rule and flag order. (2) git diff --name-status shows 9 .mbt files only, all in the three permitted packages; no .mbti changed and no visibility/pub change. (3) Explanatory comments survive (the approval spin-wait comment moved verbatim onto the helper doc; new helper docs are accurate). (4) Test counts are unchanged base vs head (serve 7, approval 10, client_test 9, types_wbtest 14, tools_test 7), each 

Scope: 9 files reviewed; build `moon test (worktree): cmd/openseek 82/82, mcp 23/23, mcp/tools 7/7 passed`, tests `moon test green on all three touched packages; test counts base==head across serve/approval/client_test/types_wbtest/tools_test`.

---

PR #1299, opened 2026-09-05
refactor: desktop/internal engine, host: share test fixtures

bobzhang:

Candidates came from a read-only survey of the whole repository. The author was told to drop any that would not compile or would change behaviour, so this branch carries fewer than were proposed: **11 applied, 0 dropped, net -191 lines.**

## Applied

- **desktop/internal/engine/follower.mbt** — 15 identical ensure_follower spawn blocks in follower tests collapse to one helper
  Applied to all 15 sites. Renamed the proposed helper spawn_test_follower -> install_test_follower: nothing is spawned, and follower_test_config's own doc comment says the engine path is never created. Signature kept as proposed (async fn[G], positional engine/session/root, labelled sink~/group~/manager~). The site at old line 749 sits inside a try/catch; the helper still raises, so the refusal capture is unchanged.
- **desktop/internal/engine/ops.mbt** — Seven byte-identical update_settings custom-provider blocks in ops tests
  Applied as configure_custom_provider. 7 sites in ops.mbt as the finding said, plus an 8th byte-identical copy in worktree_seam_wbtest.mbt (same package) so no verbatim copy is left behind.
- **desktop/internal/engine/ops.mbt** — Eight start_run payload literals differ only in task/session/workspace
  Applied as test_start_payload, 8 sites in ops.mbt plus the 9th in worktree_seam_wbtest.mbt that the finding's own evidence names. Deviation: session and workspace are LABELLED (session~ : String, workspace~ : String), not positional as proposed - three bare Strings in a row would let a caller silently transpose the conversation and its workspace. Punned call sites read test_start_payload("retry", session~, workspace~); the one site with a client id passes submission_id="submission-start".
- **desktop/internal/engine/ops.mbt** — Seven copies of "spawn the actor, wait for its pump" in ops run tests
  Applied as serve_until_pumping, 7 sites in ops.mbt plus the equivalent copy in worktree_seam_wbtest.mbt (which spelled the callback fn(_) {  } instead of fn(_) { () } - same thing). manager is passed explicitly rather than re-derived from actor.manager(), so the equivalence is visible without knowing what that accessor returns.
- **desktop/internal/host/git_ops.mbt** — Five identical local `async fn cleanup()` tmpdir removers in git_ops tests
  Applied as git_test_cleanup(dir), placed next to git_test_run. All five `defer cleanup()` became `defer git_test_cleanup(dir)`, including the two inside a `let _ = { ... }` block.
- **desktop/internal/engine/archive_lookup_wbtest.mbt** — Four identical archive_session refusal-capture blocks in one test file
  Applied as archive_refusal(manager, session), 4 sites. The fifth block (load_session at old line 141) was left alone as the finding said. Helper carries the same #cfg(not(platform="windows")) guard as all four tests.
- **desktop/internal/host/language_path.mbt** — Three identical LanguagePath::resolve rejection probes in one test
  Applied as resolve_rejected(session, root, path), 3 sites. Positional args kept here because they mirror LanguagePath::resolve's own argument order exactly. No cfg guard, matching the unguarded test.
- **desktop/internal/host/git_ops.mbt** — git_output_opt and git_output re-spell git_collect's spawn instead of calling it
  Applied to both functions. Equivalence rests on the error arms: git_collect re-raises cancellation (so both callers' `error if @async.is_being_cancelled() => raise error` arm still sees the original error) and turns anything else into HostError, which both callers already discard via `_ => return None` / `_ => return ""`. Success returns the identical tuple. This is the only production-code hot path in the commit.
- **desktop/internal/engine/archive.mbt** — Claim/guard reverse release written twice in archive.mbt
  Applied as release_family_claims(manager, claims, guards). Both sites (SessionFamilyMove::release and the rollback `defer` in claim_session_family_move) had byte-identical bodies modulo self. Parameter types are all distinct, so the positional call is unambiguous. Production code, on a rollback path.
- **desktop/internal/host/text_search.mbt** — split_text_search_globs is a map+filter over the glob-aware split
  Applied as proposed. The explanatory comment above the function (VS Code splitGlobAware) is kept verbatim - it describes the function, not the loop.
- **desktop/internal/engine/follower.mbt** — Manual push loop selecting this session's commits is Array::filter
  Applied. moon fmt settled on a block-bodied lambda: emitted.filter(event => { event is SessionCommit(payload) && payload.session == session }). The explicit Array[EngineEvent] annotation was dropped rather than carried over, since `unnecessary_annotation` is in this package's warning set.

## Verification

Both touched packages are native-only (supported_targets = "+native" in desktop/internal/engine/moon.pkg and desktop/internal/host/moon.pkg), so everything below ran on the default native target. No js or wasm target applies.

TESTS, before and after (identical):
- BEFORE `moon test desktop/internal/engine desktop/internal/host` -> "Total tests: 221, passed: 221, failed: 0."
- BEFORE `moon test desktop/internal/engine` -> 111 passed, 0 failed
- BEFORE `moon test desktop/internal/host`   -> 110 passed, 0 failed
- AFTER  `moon test desktop/internal/engine` -> 111 passed, 0 failed
- AFTER  `moon test desktop/internal/host`   -> 110 passed, 0 failed

ASSERTION / ROW COUNTS: not applicable to any finding here, and I want to be exact about why. Despite the "dedupe-test" category, none of these are table collapses: no test was deleted, no two tests were merged, and no assertion was removed, reworded or made anonymous. Every change lifted *fixture construction* (provider setup, actor/pump startup, payload literal, tmpdir cleanup, a refusal capture) out of tests that each keep their own body and their own assertions. Hence the test counts above are identical rather than reduced, and there is no "N assertions before, N rows after" to state. The one shape that could have hurt diagnosability - archive_refusal collapsing four capture blocks - keeps each caller's own debug_inspect on the returned string, so a failure still names its own test and its own expected refusal.

FORMAT:
- `moon fmt` (ran 1740 tasks), then `moon fmt --check` -> "Finished. moon: ran 1740 tasks, now up to date", clean.

CHECK, deny-warn:
- `moon check desktop/internal/engine desktop/internal/host --deny-warn` -> "Finished. moon: ran 106 tasks, now up to date". Clean on the first attempt; no warning had to be worked around.
- Additionally, because git_ops/archive changes are production code, I checked every package that imports engine or host: `moon check desktop desktop/internal/extension desktop/internal/api desktop/internal/remote --deny-warn` -> "Finished. moon: ran 119 tasks, now up to date", clean.

MOON INFO (scoped, never bare):
- `moon info desktop/internal/engine desktop/internal/host` -> "Finished. moon: ran 104 tasks". Both packages DO have tracked pkg.generated.mbti files, and `git status --short` immediately afterwards still listed exactly the same 8 modified .mbt files and nothing else - i.e. both desktop/internal/engine/pkg.generated.mbti and desktop/internal/host/pkg.generated.mbti came back byte-identical. No public interface moved, so nothing had to be reverted on that ground.

GIT:
- `git status --short` before starting: empty; branch confirmed simplify/de-engine.
- Staged by explicit path (8 files, no `git add -A`); `git status --short` showed only those 8.
- Commit 4dbe7518c, 266 insertions / 457 deletions.
- `git push -u origin simplify/de-engine` -> new branch created, tracking set. Working tree clean afterwards. No PR opened, no merge, no other branch touched.

WORKAROUND WORTH DISCLOSING: I applied the edits with a throwaway Python script (scratchpad/apply.py) that asserts an exact match count for every substitution and aborts otherwise. It aborted twice and I reset the worktree with `git checkout -- .` and re-ran from clean each time: (1) the first version inserted each shared helper before rewriting the call sites, so the helper's own body counted as a 16th duplicate; (2) the start_run payload regex missed 6 of 8 sites because those literals use field punning (`session,` not `session: session,`). Both were script bugs, not source ambiguities - the third run reported the exact expected counts for all 23 substitutions (15/7/7/8/1/1/1/4/5/5/3/2/... ) before writing anything.

## Worth a second look

Three things deserve a second look.

1. git_output / git_output_opt (desktop/internal/host/git_ops.mbt) is the only PRODUCTION hot path in this commit - every Git query in the desktop host goes through it. The equivalence is entirely in the error arms, so verify it directly rather than by eye on the happy path: git_collect re-raises cancellation unchanged, so each caller's `error if @async.is_being_cancelled() => raise error` arm still receives the original error and re-raises it (cancellation is now tested twice against the same ambient task state, which cannot disagree); every other spawn failure becomes HostError inside git_collect, which both callers then discard through the arms they already had (`_ => return None`, `_ => return ""`). Net effect: identical returns, one extra discarded string allocation on a path that was already failing.

2. release_family_claims (desktop/internal/engine/archive.mbt) is production code on a rollback path. The two former bodies were byte-identical modulo self, and release order (claims reversed, then guards reversed) is preserved exactly - that ordering is the thing that matters here, so confirm the helper's two loops are in the same order as the originals.

3. SCOPE: findings 2, 3 and 4 each named ops.mbt, but each pattern had one more verbatim copy in desktop/internal/engine/worktree_seam_wbtest.mbt (same package; finding 3's evidence names it explicitly). I converted those too, so worktree_seam_wbtest.mbt appears in the diff even though no finding lists it as its `file`. If you would rather the vetted list be applied literally, that file is the one to drop - it is 35 lines of the diff and nothing else depends on it.

Minor, cosmetic: at most start_run sites `moon fmt` re-wraps `start_run(sink, manager, test_start_payload(...))` onto five lines rather than one, so the saving there is 9 lines -> 5 rather than 9 -> 1. One site (former line 1188, the `.run_id` chain) keeps the odd 8-space continuation indent it already had; that is fmt's own output and `moon fmt --check` is clean on it.

No finding was dropped, which is worth a skeptical pass in itself: all eleven were mechanical extractions whose `current` snippets still matched the source exactly, and the compiler accepted every helper on the first attempt with deny-warn on. The single riskiest judgement call is the labelled-argument deviation in test_start_payload (finding 3, the one the surveyor rated medium) - I chose session~/workspace~ over the proposed positional Strings specifically so a transposed session/workspace pair cannot compile.

### DIFF
--- a/desktop/internal/engine/archive.mbt
+++ b/desktop/internal/engine/archive.mbt
@@ -81,16 +81,27 @@ priv struct SessionFamilyMove {
 }
 
 ///|
+/// Undo a family's holds in reverse acquisition order: engine claims first,
+/// then the record move guards.
+fn release_family_claims(
+  manager : EngineManager,
+  claims : Array[PendingClaim],
+  guards : Array[(String, SessionRecordGuard)],
+) -> Unit {
+  for claim in claims.rev_iter() {
+    claim.release(manager)
+  }
+  for entry in guards.rev_iter() {
+    let (record_key, record_state) = entry
+    manager.release_record_move(record_key, record_state)
+  }
+}
+
+///|
 fn SessionFamilyMove::release(self : SessionFamilyMove) -> Unit {
   guard self.active else { return }
   self.active = false
-  for claim in self.claims.rev_iter() {
-    claim.release(self.manager)
-  }
-  for entry in self.guards.rev_iter() {
-    let (record_key, record_state) = entry
-    self.manager.release_record_move(record_key, record_state)
-  }
+  release_family_claims(self.manager, self.claims, self.guards)
 }
 
 ///|
@@ -111,13 +122,7 @@ async fn EngineManager::claim_session_family_move(
   let mut handed_off = false
   defer {
     if !handed_off {
-      for claim in claims.rev_iter() {
-        claim.release(manager)
-      }
-      for entry in guards.rev_iter() {
-        let (record_key, record_state) = entry
-        manager.release_record_move(record_key, record_state)
-      }
+      release_family_claims(manager, claims, guards)
     }
     if placement_locked {
       manager.workspace_lifecycle_lock.release()
--- a/desktop/internal/engine/archive_lookup_wbtest.mbt
+++ b/desktop/internal/engine/archive_lookup_wbtest.mbt
@@ -5,6 +5,21 @@
 // refusal, and a failed attempt must leave no wedged host state behind.
 
 ///|
+/// The refusal `archive_session` answers with, or "no error" when it commits.
+/// Cancellation is never a refusal and stays observable.
+#cfg(not(platform="windows"))
+async fn archive_refusal(manager : EngineManager, session : String) -> String {
+  try {
+    ignore(archive_session(manager, session, (_, _) => ()))
+    "no error"
+  } catch {
+    EngineError(detail) => detail
+    error if @async.is_being_cancelled() => raise error
+    error => "\{error}"
+  }
+}
+
+///|
 #cfg(not(platform="windows"))
 async test "archive names the missing store after a workspace detach" {
   ambient_env_test_lock.acquire()
@@ -33,14 +48,7 @@ async test "archive names the missing store after a workspace detach" {
   // A stale sidebar can still offer the conversation; archiving it now
   // reports the record as unreachable instead of pretending it never
   // existed.
-  let detail = try {
-    ignore(archive_session(manager, "s-ws", (_, _) => ()))
-    "no error"
-  } catch {
-    EngineError(detail) => detail
-    error if @async.is_being_cancelled() => raise error
-    error => "\{error}"
-  }
+  let detail = archive_refusal(manager, "s-ws")
   debug_inspect(
     detail,
     content=(
@@ -54,14 +62,7 @@ async test "archive names the missing store after a workspace detach" {
   assert_true(manager.record_guards.get("s-ws") is None)
   guard manager.pump is Serving(sessions~) else { fail("pump stopped") }
   assert_true(sessions.get("s-ws") is None)
-  let second = try {
-    ignore(archive_session(manager, "s-ws", (_, _) => ()))
-    "no error"
-  } catch {
-    EngineError(detail) => detail
-    error if @async.is_being_cancelled() => raise error
-    error => "\{error}"
-  }
+  let second = archive_refusal(manager, "s-ws")
   debug_inspect(
     second,
     content=(
@@ -95,14 +96,7 @@ async test "archive reports a workspace record whose directory vanished" {
   // The workspace directory vanishes outside the app; the registry keeps
   // its dead entry addressable, so the sidebar still shows the workspace.
   @fs.rmdir(workspace, recursive=true)
-  let detail = try {
-    ignore(archive_session(manager, "s-ws", (_, _) => ()))
-    "no error"
-  } catch {
-    EngineError(detail) => detail
-    error if @async.is_being_cancelled() => raise error
-    error => "\{error}"
-  }
+  let detail = archive_refusal(manager, "s-ws")
   debug_inspect(
     detail,
     content=(
@@ -169,14 +163,7 @@ async test "archiving an already-archived conversation names that state" {
   @fsx.ensure_dir(Path(store + "/archived/sessions/s-done"))
   let manager = new_engine_manager("unused")
   manager.pump = Serving(sessions=Map([]))
-  let detail = try {
-    ignore(archive_session(manager, "s-done", (_, _) => ()))
-    "no error"
-  } catch {
-    EngineError(detail) => detail
-    error if @async.is_being_cancelled() => raise error
-    error => "\{error}"
-  }
+  let detail = archive_refusal(manager, "s-done")
   debug_inspect(
     detail,
     content=(
--- a/desktop/internal/engine/follower.mbt
+++ b/desktop/internal/engine/follower.mbt
@@ -677,6 +677,29 @@ fn follower_test_config(
 }
 
 ///|
+/// Install one test follower for `session` under `root`. The identity
+/// `ensure_follower` answers with is only present when this call installed a
+/// new follower; every test below reads the follower map instead.
+#cfg(not(platform="windows"))
+async fn[G] install_test_follower(
+  engine : String,
+  session : String,
+  root : @pathx.Absolute,
+  sink~ : EventSink,
+  group~ : @async.TaskGroup[G],
+  manager~ : EngineManager,
+) -> Unit {
+  ignore(
+    ensure_follower(
+      sink~,
+      group~,
+      manager~,
+      config=follower_test_config(engine, session, root),
+    ),
+  )
+}
+
+///|
 #cfg(not(platform="windows"))
 async test "a follower barrier broadcasts the durable tail before returning" {
   let dir = @fs.tmpdir(prefix="openseek-follower-barrier-")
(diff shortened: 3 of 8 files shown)


bobzhang:

**Verdict: signed off.** 2 finding(s), none blocking (1 low, 1 nit); each is answered below.

> Audit of the pure-simplification PR over desktop/internal/engine and desktop/internal/host (1 commit 4dbe7518c, diff base a227f6aae, worktree clean). VERDICT: the branch satisfies all seven acceptance criteria — no blockers. Method: read the full per-file diff of all 8 changed files; verified every replacement block is textually identical to the removed code modulo parameter threading; verified balanced replacement counts (ops.mbt: 7 update_settings-custom blocks→configure_custom_provider, 7 spawn+pump-wait blocks→serve_until_pumping, 8 start_run payload literals→test_start_payload; follower.mbt: 15 ensure_follower ignore-blocks→install_test_follower; archive_lookup_wbtest: 4 refusal try/catch blocks→archive_refusal; git_ops.mbt: 5 local cleanup fns→git_test_cleanup with 5 defer call sites; language_path: 3 resolve probes→resolve_rejected; archive.mbt: the two identical reverse-release loops→release_family_claims; text_search & follower 'comprehensions' are semantics-preserving map/filter equivalents of the loops, same order/snapshot, no early-stop or side-effect differences). Empirical gates run in the scratch lab on linked worktrees at both HEAD and base a227f6aae: `moon check -p` on both packages passes at HEAD; `moon test -p desktop/internal/engine --target native` passes 111/111 at both revisions; `moon test -p desktop/internal/host --target native` passes 110/110 at both revisions — identical counts and all spawn/interleaved logs match. Criteria: 1 BEHAVIOUR — no hunk changes results, error paths (e.g. git_output_opt/None and git_output/"" retain the catch-arms; cance

Scope: 8 files reviewed; build `moon check -p desktop/internal/engine & desktop/internal/host: OK at HEAD`, tests `engine 111/111, host 110/110 pass at both HEAD (4dbe7518c) and base (a227f6aae) on native`.

### Findings and responses

**[low] `desktop/internal/engine/ops.mbt:937` — Test-only fixture helpers are compiled into production packages**

The dedupe helpers are top-level functions in production source files, so they are compiled into every non-test build of the two packages: configure_custom_provider (ops.mbt:937), serve_until_pumping (ops.mbt:953), test_start_payload (ops.mbt:972), install_test_follower (follower.mbt:684), git_test_cleanup (git_ops.mbt:1893) and resolve_rejected (language_path.mbt:110). Only release_family_claims (archive.mbt:86) is on a production path; the rest are reachable only from test blocks, yet ship in the engine/host library. All are non-pub (verified: no pub keyword, and the .mbti diff across the branch is empty), so criterion 2 is not violated and there is no behaviour change — but the 'pure …

*Response:* Accepted as intended, not changed. These helpers sit beside the inline `test` blocks that use them, which is this repository's existing shape: on `main`, before this PR, `ops.mbt` already carried 12 inline tests alongside 10 top-level functions, `follower.mbt` 20 alongside 13, and `git_ops.mbt` 22 alongside 53. MoonBit has no test-only scope within a file, so moving the helpers out would mean relocating the test blocks into `_wbtest.mbt` files as well — a restructuring well beyond a simplification pass. The finding is right that they add non-`pub` code to the library; the `.mbti` diff is empty, so nothing is exposed.

**[nit] `desktop/internal/engine/worktree_seam_wbtest.mbt:269` — One changed file is not on the vetted change list**

worktree_seam_wbtest.mbt was rewritten even though it does not appear in the vetted list of dedupe/comprehension hunks. The change is identical in kind and package (three call sites replaced with the helpers defined in ops.mbt: configure_custom_provider, serve_until_pumping, test_start_payload), covers exactly the same one test each, and stays inside desktop/internal/engine, so it does not violate criterion 7 (nothing unrelated / no drive-by edit). It is simply an extension beyond the listed set — the list said the branch 'may contain fewer', not more. Counts confirm no case was added or dropped (worktree_seam old: 1 start_run/1 update_settings/1 spawn; new: 1 start_run, helpers shared).

*Response:* Correct, and now stated in the PR description. It is the same dedupe applied to a sibling test file of the same package, using the helpers this PR adds; the auditor confirmed no case was added or dropped.

---

PR #1300, opened 2026-09-05
refactor: desktop/internal services: reuse stdlib and share decode helpers

bobzhang:

Candidates came from a read-only survey of the whole repository. The author was told to drop any that would not compile or would change behaviour, so this branch carries fewer than were proposed: **22 applied, 0 dropped, net -235 lines.**

## Applied

- **desktop/internal/skillmarket/skillmarket_test.mbt** — Seven install-refusal tests differ only in fixture slug, message phrase, and directory name
  Applied with an adjustment. The proposed 3-column table would have dropped seven descriptive test names, one of which ('malformed archive coordinates are a registry error, not absence') states a distinction the row data does not carry. I used a 4-column row (scenario, module, phrase, directory) and put the scenario in msg= on BOTH assertions, so a failing row still names the refusal. 14 assertions before (7 tests x 2), 7 rows x 2 = 14 assertions after; same inputs, same phrases, same directory-absence checks, fresh @fs.tmpdir per row. Test count 38 -> 32, which is exactly 7 tests collapsing into 1.
- **desktop/internal/protocol/desktop_replies_test.mbt** — 8 decode-rejection assertions repeat the same try/catch/noraise scaffold; extract a `refused` helper
  Applied to 9 sites (the 8 named plus the nested one at the TerminalInputPayload loop). Adjusted: kept the existing named bindings (both_rejected, empty_error_sentinel_rejected, ...) rather than inlining into assert_true, because those names are the only label each case carries. The helper body is the original try/catch/noraise verbatim, so refusal semantics are byte-identical. Left the one site whose noraise arm calls fail("expected malformed directory creation payload to fail") alone -- different shape, and the finding excluded it.
- **desktop/internal/auth/auth_wbtest.mbt** — The same six-field AuthSession literal is spelled out six times; extract sample_session
  Applied to all 6 sites. Adjusted: the avatar is a required LABELLED argument (sample_session(avatar="")) rather than positional, so every call site still states the varying field. Dropped the now-redundant ': AuthSession' annotations (this package has +unnecessary_annotation). The seventh, similar-looking literal (https://old.example / d_old / Old Mac) is a deliberately DIFFERENT session and was left untouched.
- **desktop/internal/api/api.mbt** — Eight sites copy a Map with a for/assign loop; Map::copy (and Array::map) already do it
  Applied to all 8 sites: 6 in api.mbt (annotate_thread, annotate, the turn-history thread/reply pair, and the two codex.thread scoped-params arms) and 2 in codex/project_catalog.mbt. Array::map used only in the non-async annotate; project_catalog keeps its for loop for the data array because ProjectCatalog::annotate_thread is async.
- **desktop/internal/protocol/json_object.mbt** — 9 codecs hand-roll the same indexed array decode; add JsonObject::required_array_of
  Helper added beside required_string_array; all 9 call sites rewritten (git_protocol_codec x3, text_search_codec x2, semantic_search_codec x3, desktop_replies x1). Verified every one of the 9 builds its object with JsonObject::decode(json, path, ..), so self.path == path and the emitted JsonPath is unchanged.
- **desktop/internal/pathx/pathx_test.mbt** — 16 identical `guard ...to_absolute() is Some(x) else { fail(...) }` fixture blocks collapse to one helper
  Applied, and extended from the finding's 16 sites to all 20 -- the other 4 spell the same guard with the message 'expected an absolute POSIX/Windows fixture'. Leaving 4 hand-rolled copies beside 16 helper calls would have been worse. #callsite(autofill(loc)) + loc~ keeps the reported failure line at the call site. Their four failure messages change to 'test fixture <text> was not absolute', which names the exact fixture and is strictly more informative.
- **desktop/internal/remote/serve.mbt** — Three delivery tests rebuild the same AgentEvent notification inline
  Applied to 4 sites: the 3 named plus the CompactionFinished construction in the same test, which is the identical run_id=None / session="session-1" shape. client_frame sits beside the existing expect_notification helper.
- **desktop/internal/protocol/desktop_messages.mbt** — Eight lenient optional reads repeat the same 4-line match; extract lenient_string/lenient_int
  Applied to all 8 sites (AgentErrorPayload x4, FinishedPayload x2, RunSettlementPayload x2 in desktop_replies.mbt). The helpers live in json_object.mbt with a doc comment saying why they cannot reuse JsonObject::optional_string/optional_int -- those raise on a wrong type, these deliberately read it as absent. semantic_search_codec's description reader stays as-is; it raises.
- **desktop/internal/extension/external_link.mbt** — The scheme allowlist is a 3-branch if/else-if chain that is literally a list plus `any`
  Applied. The three prefixes are mutually exclusive ('https://' does not start with 'http://'), so scanning all three gives the same answer as short-circuiting on the first match. The 'strict lowercase allowlist' doc comment is kept and now names an actual list. The existing 7-case test still passes.
- **desktop/internal/api/hub.mbt** — forget_run_approvals hand-rolls collect-then-remove; Map::retain does it in one call
  Applied with a fix: the proposed `not(...)` is deprecated in this toolchain (moon check error [0020] 'Use !expr instead'), so the predicate is `!(approval.run_id is Some(owner) && owner == run_id)`. The 'Drop every question belonging to run_id' comment is unchanged.
- **desktop/internal/api/hub.mbt** — Two push-every-value loops in Hub::runs are Map::values().to_array()
  Both sites applied. Order is load-bearing here (hub.mbt documents 'in run-start order (the map preserves insertion order)'); Map::values walks head->next, the same order `for _, v in map` yields.
- **desktop/internal/protocol/pattern_repair_codec.mbt** — The non-negative generation guard is written twice in one file
  Applied as proposed. The message 'expected non-negative generation' is carried through unchanged -- it differs from text_search_codec's require_text_search_non_negative ('expected non-negative integer'), which is why that helper cannot be reused.
- **desktop/internal/worktree/submodules.mbt** — 11 sequential `args.push(...)` calls are three `Array::append` calls
  Applied. Argument order unchanged, so the generated git argv is byte-identical. The 'git submodule does not forward repository-local protocol policy' comment is untouched.
- **desktop/internal/update/update.mbt** — Hand-rolled hex-digit scan is `hex.iter().all(...)`
  Applied as a guard so the early `return None` is preserved. String::iter yields Char, the same as `for ch in hex`.
- **desktop/internal/session_store/store.mbt** — `known_roots` accumulator loop is a one-line array comprehension
  Applied. @workspaces.registered() is async and @workspaces.store_root is a pure non-async mapping, so the comprehension compiles and preserves order. moon fmt reflowed it onto three lines.
- **desktop/internal/worktree/worktree.mbt** — Test host's drain loop is `drained.append(sessions)`
  Applied as proposed; both sides are Array[String] and append preserves order exactly as the push loop did.
- **desktop/internal/workflow/tail.mbt** — Manual chunk clamp is `@cmp.minimum`
  Applied; needed one added import line, "moonbitlang/core/cmp", in desktop/internal/workflow/moon.pkg. Both operands are Int64.
- **desktop/internal/session_record/record.mbt** — Manual chunk clamp is `@cmp.minimum`
  Applied; one added import line in desktop/internal/session_record/moon.pkg.
- **desktop/internal/subrun/probe.mbt** — Manual chunk clamp is `@cmp.minimum`
  Applied; one added import line in desktop/internal/subrun/moon.pkg.
- **desktop/internal/worktree/worktree.mbt** — Four-way `==` chain on porcelain status letters is two `is ("R" | "C")` tests
  Applied. Both operands are String; the repo already uses string or-patterns with `is` this way in shellenv.mbt. Same truth table, same short-circuit order over the two distinct subjects, so the extra `index += 1` for rename/copy records fires on exactly the same inputs.
- **desktop/internal/moonbit/moonbit_toolchain.mbt** — Test's `match Option` render is `.map(...).unwrap_or(...)`
  Applied as proposed.
- **desktop/internal/worktree/worktree.mbt** — Two copies of the "present, non-empty JSON string" decode share one helper
  Applied as proposed; owner_field sits directly above read_worktrees_unlocked with the finding's doc comment. The 'A row cannot be owned by both products' comment below is untouched.

## Verification

All commands run from the worktree's desktop/ directory unless noted. moon 0.1.20260827 (d0aaa07), macOS, native target (desktop/internal/** is native-only; no frontend files touched).

FORMAT: `moon fmt` after each edit batch, then `moon fmt --check` at the end -> "Finished. moon: ran 1740 tasks, now up to date" (clean, no reformat pending). moon fmt reflowed several of my edits and I accepted its output: the known_roots comprehension onto 3 lines, the external_link closure into a block body, the skillmarket table rows, and the `refused` helper into `try decode() catch {..} noraise {..}`.

CHECK: `moon check --deny-warn` over the WHOLE desktop module (native) -> "Finished. moon: ran 731 tasks, now up to date" (0 errors, 0 warnings). Ran module-wide rather than only the 15 packages, to catch downstream breakage from the api/protocol/session_store edits. `moon check --target js --deny-warn` -> "Finished. moon: ran 158 tasks, now up to date" (desktop/frontend is js-only and imports internal/protocol, so the codec edits had to hold on js too). Per-package `moon check <pkg> --deny-warn` was also run after each batch.

ONE FAILURE, FIXED: the first Map::retain attempt used the finding's `not(...)`, which this toolchain rejects: "Error [0020] ... Error Warning (deprecated): Use !expr instead" at hub.mbt:178. Rewritten as `!(approval.run_id is Some(owner) && owner == run_id)`; re-checked clean. No other command failed and there were no other workarounds.

TESTS, before -> after (`moon test internal/<pkg>`, run once on the untouched tree and once after all edits):
skillmarket 38 -> 32 (INTENDED: 7 install-refusal tests collapsed into 1; 14 assertions before, 7 table rows x 2 assertions = 14 after, each row labelled with its old test name in msg=)
protocol 67 -> 67; api 22 -> 22; codex 12 -> 12; auth 12 -> 12; pathx 12 -> 12; remote 9 -> 9; extension 8 -> 8; worktree 14 -> 14; update 11 -> 11; session_store 3 -> 3; workflow 6 -> 6; session_record 15 -> 15; subrun 8 -> 8; moonbit 10 -> 10. All passed, 0 failures, in both runs.
Note: the codex baseline was captured after the fact -- I copied my edited project_catalog.mbt aside, `git checkout --` the original, ran `moon test internal/codex` (12 passed), then restored my copy and confirmed `git status` still showed only that one modification.

MBTI: `moon info internal/skillmarket internal/auth internal/api internal/codex internal/pathx internal/remote internal/extension internal/worktree internal/update internal/session_store internal/workflow internal/session_record internal/subrun internal/moonbit internal/protocol` (SCOPED, never bare) -> "Finished. moon: ran 127 tasks". `git status --short` immediately afterwards listed exactly the same 27 files as before it -- NO .mbti changed, so no public interface moved.

GIT: `git status --short` showed only the 27 intended files; staged by explicit path (no `git add -A`); committed as 8633c93af "refactor(desktop/internal): reuse stdlib and share decode helpers" with the Co-Authored-By and Claude-Session footers; `git push -u origin simplify/de-rest` created the branch on origin. Working tree clean after push. No PR opened, no merge, no other branch touched. Diffstat: 27 files, 261 insertions, 496 deletions (net -235).

## Worth a second look

Four places where I went beyond or against the literal proposal, all called out in the per-finding notes but worth a second look:

1. skillmarket_test.mbt is the one medium-risk change and the only test-count change (38 -> 32). Seven separately named tests became one table. I judged the seven names to be load-bearing documentation, so each row carries its old test name as a 4th column and that name appears in msg= on both assertions -- verify you agree the collapse is worth it, and that the shared fake_registry (one HTTP server now serves all seven sequential installs instead of one server per test) is acceptable. Filesystem isolation is preserved: still a fresh @fs.tmpdir per row.

2. pathx_test.mbt: I converted 20 fixture guards, not the 16 the finding named. The extra 4 had the message "expected an absolute POSIX/Windows fixture", which now reads "test fixture <text> was not absolute". Failure messages in an impossible branch, but they did change.

3. Twelve of the twenty pathx guards and several other pathx/worktree lines sit inside `#cfg(platform="windows")` blocks that this macOS host does NOT compile. Their POSIX twins are compile-verified and the transformation is line-for-line identical, but nothing here type-checked the Windows arms.

4. serve.mbt got a 4th client_frame call site (the CompactionFinished construction) that the finding did not list; it is the identical AgentEvent shape.

Also worth a look: `Map::copy` and `Map::values().to_array()` both depend on MoonBit's Map preserving insertion order, which hub.mbt:193 and project_catalog.mbt explicitly document relying on. I checked core's linked_hash_map implementation (copy walks tail->prev, values walks head->next) but this is the one semantic assumption the whole api/hub group rests on. And three moon.pkg files gained a `"moonbitlang/core/cmp"` import for a single @cmp.minimum call each -- a fair trade for removing 12 lines, but it is 3 new package dependencies.

### DIFF
--- a/desktop/internal/api/api.mbt
+++ b/desktop/internal/api/api.mbt
@@ -176,10 +176,7 @@ fn CodexWorktreeCatalog::annotate_thread(
     is Some(binding) else {
     return value
   }
-  let annotated : Map[String, Json] = Map([])
-  for key, field in thread {
-    annotated[key] = field
-  }
+  let annotated = thread.copy()
   // The registry retains the owning project even after its checkout is gone;
   // Git cannot recover that relationship from a deleted cwd.
   annotated["projectRoot"] = Json(binding.workspace)
@@ -193,16 +190,11 @@ fn CodexWorktreeCatalog::annotate(
   reply : Json,
 ) -> Json {
   guard reply is Object(fields) else { return reply }
-  let annotated : Map[String, Json] = Map([])
-  for key, field in fields {
-    annotated[key] = field
-  }
+  let annotated = fields.copy()
   if fields.get("data") is Some(Array(threads)) {
-    let data : Array[Json] = []
-    for thread in threads {
-      data.push(self.annotate_thread(thread))
-    }
-    annotated["data"] = Json(data)
+    annotated["data"] = Json(
+      threads.map(thread => self.annotate_thread(thread)),
+    )
   }
   if fields.get("thread") is Some(thread) {
     annotated["thread"] = self.annotate_thread(thread)
@@ -805,15 +797,9 @@ async fn ApiState::codex_thread_history(
       _ => break
     }
   }
-  let thread : Map[String, Json] = Map([])
-  for key, value in thread_fields {
-    thread[key] = value
-  }
+  let thread = thread_fields.copy()
   thread["turns"] = Json(turns)
-  let reply : Map[String, Json] = Map([])
-  for key, value in reply_fields {
-    reply[key] = value
-  }
+  let reply = reply_fields.copy()
   reply["thread"] = Json(thread)
   @json.from_json(Json(reply))
 }
@@ -1086,10 +1072,7 @@ async fn ApiState::codex_request(
     // Preserve the caller's source filter (or Codex's interactive default)
     // instead of hiding threads that OpenSeek itself just created.
     ("codex.thread.list", Object(fields)) => {
-      let scoped : Map[String, Json] = Map([])
-      for key, value in fields {
-        scoped[key] = value
-      }
+      let scoped = fields.copy()
       // JSONL scan-and-repair can exceed the Proton request deadline on a
       // long-lived Codex home. The Desktop catalog is an interactive state
       // query, so read the app-server state database without that repair pass.
@@ -1100,10 +1083,7 @@ async fn ApiState::codex_request(
     // The dedicated thread/start endpoint has already reduced this object to
     // cwd, model, and Desktop's fixed service name.
     ("codex.thread.start", Object(fields)) => {
-      let scoped : Map[String, Json] = Map([])
-      for key, value in fields {
-        scoped[key] = value
-      }
+      let scoped = fields.copy()
       scoped["serviceName"] = "openseek_desktop"
       Json(scoped)
     }
--- a/desktop/internal/api/hub.mbt
+++ b/desktop/internal/api/hub.mbt
@@ -174,15 +174,9 @@ fn Hub::track_approval(
 ///|
 /// Drop every question belonging to `run_id`.
 fn Hub::forget_run_approvals(self : Hub, run_id : String) -> Unit {
-  let doomed : Array[String] = []
-  for session, approval in self.pending_approvals {
-    if approval.run_id is Some(owner) && owner == run_id {
-      doomed.push(session)
-    }
-  }
-  for session in doomed {
-    self.pending_approvals.remove(session)
-  }
+  self.pending_approvals.retain((_, approval) => {
+    !(approval.run_id is Some(owner) && owner == run_id)
+  })
 }
 
 ///|
@@ -204,10 +198,7 @@ fn Hub::runs(
   self : Hub,
   known? : Array[KnownRunPayload],
 ) -> @protocol.RunsReply {
-  let runs : Array[@protocol.StartedPayload] = []
-  for _, payload in self.active_starts {
-    runs.push(payload)
-  }
+  let runs = self.active_starts.values().to_array()
   let settled : Array[@protocol.RunSettlementPayload] = []
   if known is Some(selectors) {
     for _, settlement in self.settlements {
@@ -220,10 +211,7 @@ fn Hub::runs(
   // client only wants for runs it owns, while a standing question is a turn
   // stopped RIGHT NOW that anyone looking at that session should be able to
   // answer.
-  let approvals : Array[@protocol.PendingApprovalPayload] = []
-  for _, approval in self.pending_approvals {
-    approvals.push(approval)
-  }
+  let approvals = self.pending_approvals.values().to_array()
   { runs, settled, approvals, }
 }
 
--- a/desktop/internal/auth/auth_wbtest.mbt
+++ b/desktop/internal/auth/auth_wbtest.mbt
@@ -3,6 +3,20 @@
 // signin.mbt is built from.
 
 ///|
+/// The one signed-in session every codec and store test below uses. The avatar
+/// is the only field any of them varies, so it stays an explicit argument.
+fn sample_session(avatar~ : String) -> AuthSession {
+  {
+    server_url: "https://relay.example",
+    device_token: "odt_secret",
+    device_id: "d_abc",
+    device_name: "My Mac",
+    user_login: "octocat",
+    user_avatar_url: avatar,
+  }
+}
+
+///|
 test "pkce challenge is the verifier's unpadded base64url sha256" {
   // Cross-checked against RFC 7636 S256 (python:
   // urlsafe_b64encode(sha256(b"test-verifier")).rstrip("=")) — and against
@@ -86,14 +100,7 @@ test "url derivations" {
 
 ///|
 test "auth.json codec round-trips and reads corruption as signed out" {
-  let session : AuthSession = {
-    server_url: "https://relay.example",
-    device_token: "odt_secret",
-    device_id: "d_abc",
-    device_name: "My Mac",
-    user_login: "octocat",
-    user_avatar_url: "https://avatars.example/1",
-  }
+  let session = sample_session(avatar="https://avatars.example/1")
   let json = session_to_json(session)
   guard session_from_json(json) is Some(reloaded) else {
     fail("round-trip lost the session")
@@ -108,14 +115,7 @@ test "auth.json codec round-trips and reads corruption as signed out" {
 async test "auth.json persists 0600 and deletes cleanly" {
   let dir : @pathx.Path = Path(@fs.tmpdir(prefix="openseek-auth-"))
   let body = async fn() {
-    let session : AuthSession = {
-      server_url: "https://relay.example",
-      device_token: "odt_secret",
-      device_id: "d_abc",
-      device_name: "My Mac",
-      user_login: "octocat",
-      user_avatar_url: "",
-    }
+    let session = sample_session(avatar="")
     save_session_file(dir, session)
     guard load_session_file(dir) is Some(reloaded) else {
       fail("persisted session did not load")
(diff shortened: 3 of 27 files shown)


bobzhang:

**This is the second audit.** The first flagged two `desktop/internal/codex/project_catalog.mbt` hunks as outside the PR's stated package list. They are intended — the same Map-copy simplification is swept across `desktop/internal` — so the scope was corrected in the description and the branch re-audited with `codex` declared in scope. It now comes back clean.

> The worktree (clean at HEAD 8633c93af, parent a227f6aae) satisfies all seven acceptance criteria for the pure-simplification PR. No findings.
> 
> Verified per criterion:
> 1. BEHAVIOUR IDENTICAL — Reviewed every hunk in all 27 changed files against the base. The rewrites are mechanically faithful: Map::copy()/fields.copy() (8 sites in api/api.mbt + codex/project_catalog.mbt) replicate the entry-copy loop including insertion order (confirmed by reading the bundled core LinkedHashMap: copy preserves order, values()/keys()/iter() are insertion-ordered, retain() removes exactly the entries failing its predicate — the equivalent of the collect-then-remove loop in Hub::forget_run_approvals, with no suspension point to expose ordering differences). required_array_of keeps the same per-index decode path and first-failure raise ordering as the 9 inline loops it replaced; lenient_string/lenient_int, owner_field, required_generation, absolute_fixture, sample_session, refused, and client_frame are exact extractions; external_link .any, update .all, @cmp.minimum x3 (same branch-on-greater semantics; equal case yields the same value), submodules args.append (same order incl. the conditional -c pair), drain append, store comprehension, moonbit_toolchain map/unwrap_or, and the worktree is ("R"|"C") status rewrite are all value- and order-identical. Tests were also compared: pathx 16 guard-blocks -> helper (count preserved), auth 6 literals -> helper with identical avatar values, desktop_replies_test try/catch scaffolds -> refused, remote/serve 3 inline AgentEvent wrappers -> client_frame,

Scope: 27 files reviewed; build `moon check green in all 15 changed packages`, tests `moon test module-wide: wasm 97/97, js 1893/1893, native 1958/1958 passed`.

---

bobzhang:
## Held for your review

### Why it is held

**Seven independent test scenarios became one table.** `skillmarket_test.mbt` folded these named tests into a single loop over a 7-row table sharing one `fake_registry`:

`a path-escaping archive is refused`, `a skill whose archive is missing reports it plainly`, `a skill without archive coordinates is not installable`, `an archive that fails its digest check is refused`, `an archive whose SKILL.md is not text is refused`, `an archive without a SKILL.md is refused at install`, `malformed archive coordinates are a registry error, not absence`.

Coverage on a passing run is unchanged — all seven fixtures still run. What changes is failure isolation: **a failing row aborts the remaining rows**, so if two refusal paths regress at once, one CI run names only the first. Those seven names also documented seven distinct security-relevant refusals, and they no longer appear in the test listing. Whether that trade is worth 55 lines is your call, not mine.

Workspace test-block count on native drops by 6 as a result (3230 → 3224); js and wasm are unchanged.

The reviewer read all 75 hunks and checked every stdlib substitution against core's source rather than assuming: `Map::copy` preserves insertion order, `Map::retain` keeps-where-true and is order-preserving, `@cmp.minimum` reproduces the three tail/probe if-chains exactly, and `Array::any`/`Iter::all` short-circuit where the else-if chain and early `return None` did. For `JsonObject::required_array_of`, hoisting `self.path.add_key(name)` out of the loop is safe because `JsonPath::add_key`/`add_index` return fresh nodes and core's decoders only mutate nodes they allocate.

### One concrete defect worth fixing either way

`desktop/internal/protocol/json_object.mbt:56` — the new `lenient_int` doc comment claims it truncates "the way the raising integer readers truncate". It does not: `required_int` goes through core's `FromJson` for `Int`, which **raises** `Int::from_json: overflow` outside [-2³¹, 2³¹) rather than truncating. With `exit_code: 1e300`, `lenient_int` returns a truncated value while `required_int` raises. Comment inaccuracy only, no behaviour change.

PR #1302, opened 2026-09-05
refactor: desktop/frontend packages: comprehensions, stdlib reuse, shared helpers

bobzhang:

Candidates came from a read-only survey of the whole repository. The author was told to drop any that would not compile or would change behaviour, so this branch carries fewer than were proposed: **35 applied, 1 dropped, net -277 lines.**

## Applied

- **desktop/frontend/transcript/component/component.mbt** — Three tests rebuild the same 14-line `Input` and the same 11-line `Model` literal
  Added package-private `base_input()` / `pinned_model(input)` fixtures; the third test keeps its distinct model as `{ ..pinned_model(input), pinned: false, overview: {...} }`. Scoped to component.mbt only — I did NOT touch the near-identical literal in view_tests.mbt (see reviewer_notes). All three test names, inputs and assertions unchanged.
- **desktop/frontend/dock/update.mbt** — Five `open_*` functions repeat the same append-unless-present-then-activate body
  Extracted private `open_tab(state, tab)`; each of open_file / open_history_diff / open_browser / open_picker / open_workflows keeps its own doc comment and public signature.
- **desktop/frontend/dock/view.mbt** — dock/view.mbt re-declares four codicons byte-identically to `@icons.icon_*`
  Deleted icon_layout_sidebar_right, icon_add, icon_folder, icon_search; call sites now use @icons.*. Bodies were byte-identical modulo the `@icons.` qualifier on `codicon`; the CC BY 4.0 attribution and the innerHTML/morph rationale live verbatim in the icons package header.
- **desktop/frontend/dock/view.mbt** — dock/view.mbt's private `icon` wrapper duplicates `@icons.icon`, comment and all
  Deleted; nine call sites now call @icons.icon. The doc comment was byte-identical to the one on @icons.icon, so nothing explanatory was lost.
- **desktop/frontend/skills/document.mbt** — `lines_to_string` is a hand-rolled `join("\n")`; the copy loop feeding it is also unneeded
  `rest[index + 1:].join("\n")`; helper deleted (single caller).
- **desktop/frontend/skills/document.mbt** — Copying an ArrayView into an Array by push loop; `ArrayView::to_owned` does it
  `let rest = rest.to_owned()`; still an Array[String] because find_frontmatter_close takes one.
- **desktop/frontend/action_menu/action_menu.mbt** — Menu positioning re-implements clamp/max inline; core has Double::clamp and Double::max
  `wanted_left.clamp(min=min_left, max=max_left.max(min_left))` and the same for top. Added a two-line comment naming why `.max(min_*)` is there (it keeps core's `guard! min <= max` satisfied when the menu is larger than the viewport).
- **desktop/frontend/action_menu/action_menu.mbt** — Delete the private clamp helper — it is a verbatim copy of core's Double::clamp
  Deleted; both callers were rewritten by the finding above.
- **desktop/frontend/action_menu/action_menu.mbt** — enabled_position is Array::search
  `indices.search(index)`. Kept the wrapper (the finding's optional extra step of inlining it at the two call sites was not taken).
- **desktop/frontend/fileeditor/view.mbt** — path_segments hand-rolls split-on-"/"; one comprehension over String::split does it
  `[for segment in path.0.split("/") if !segment.is_empty() => segment.to_owned()]`; doc comment kept verbatim.
- **desktop/frontend/fileeditor/view.mbt** — change_row_index is Array::search_by
  Also applied to the two sibling scans the finding's evidence names: review_neighbor (now `search_by` + `changes.get(index + delta)` — Array::get is literally the `>= 0 && < length` check it replaces) and active_review_position.
- **desktop/frontend/fileeditor/view.mbt** — active_review_position's index scan is search_by + Option::map
  `changes.search_by(...).map(index => (review.selected, index, changes.length()))`.
- **desktop/frontend/fileeditor/editor_resize.mbt** — drag_size open-codes Double::clamp, which has the identical body in core
  `raw.clamp(min=floor, max=ceiling).to_int()`; the `ceiling` line is retained, so both core's guard and the doc comment's "floor wins past the ceiling" still hold.
- **desktop/frontend/fileeditor/editor_resize.mbt** — sync_root_style repeats the same setProperty block three times; a local helper collapses it
  Local `fn set_px(name, size)` capturing the single `style` handle; same three properties in the same order, one DOM lookup as before. Four-line doc comment kept verbatim.
- **desktop/frontend/fileeditor/feedback/service.mbt** — items_for hand-rolls Map::get_or_init
  `self.items_by_resource.get_or_init(resource.to_string(), () => [])`; doc comment ("creating it on first mutation") stays accurate.
- **desktop/frontend/fileeditor/files.mbt** — read_directory's entries accumulator is Array::filter_map
  `reply.entries.filter_map(entry => FileEntry::from_reply(entry, path))`.
- **desktop/frontend/fileeditor/git_history.mbt** — find_graph_lane is Array::search_by
  `lanes.search_by(lane => lane.id == id)`.
- **desktop/frontend/fileeditor/git_history.mbt** — The parent-lookup break loop and its mut accumulator are one if/else-if with Array::any
  `else if commits.any(item => item.id == parent)`; the `mut` is gone. `parent_index == 0` still short-circuits before any scan.
- **desktop/frontend/files/request.mbt** — The same 6-line filter_map lambda over reply.files appears twice; both are one comprehension
  Both sites (Request::run and searching_loader) now `[for path in reply.files if !path.is_empty() => @pathx.Relative(path)]`.
- **desktop/frontend/dock/state.mbt** — `file_paths` / `history_diffs` are push loops that are exactly `filter_map`
  filter_map preserves order, which both doc comments depend on.
- **desktop/frontend/codex/activity_wbtest.mbt** — Four ledger tests share one body that differs only in the mutation, status and mark
  Added `stale_snapshot_mark(mutate, status) -> @conversation.RunMark` with a doc comment describing the stale-frontier scenario. All four test names, mutations, statuses and expected marks preserved; each test keeps its own assert, so a failure still names the exact case.
- **desktop/frontend/markdown/markdown.mbt** — `lines_text` hand-rolls a first-flag StringBuilder join over an Iter
  `lines.iter().map(line => line.v).join("\n")`.
- **desktop/frontend/terminal/host.mbt** — Collect-then-remove over a Map is Map::retain
  `self.pending.retain((_, pending) => pending.0 != channel)`. retain is the reason the two-phase form existed — it removes while walking the entry list.
- **desktop/frontend/terminal/resize.mbt** — Hand-written max and three-branch clamp in the resize tracker
  `max.max(floor)` then `raw.clamp(min=floor, max=ceiling)`; the doc comment's "matching the CSS `clamp`" now names the mechanism literally.
- **desktop/frontend/preview/url.mbt** — all_digits is a seen_digit accumulator loop; StringView::all + Char::is_ascii_digit says it directly
  `!text.is_empty() && text.all(ch => ch.is_ascii_digit())`; `seen_digit` was exactly `!is_empty()`.
- **desktop/frontend/preview/state.mbt** — page_label's nested match on host_of is Option::unwrap_or
  Outer `match page.url` kept so the "New Tab" arm stays explicit.
- **desktop/frontend/mention_context/mention_context.mbt** — An unreachable second copy of the file-mention arm sits behind the first
  Dead-code deletion: identical condition on the same unchanged `line`, and the first arm always `continue`s. I verified I removed the SECOND copy (the one missing `MentionAttribute::decode`), leaving the live decoded arm untouched.
- **desktop/frontend/composer/codex_request.mbt** — execpolicy_label repeats the JSON string-array loop twice; extract one helper
  Added `json_string_parts(values : Array[Json]) -> Array[String]` near the top of the file; execpolicy_label now assigns to `let mut parts` twice. Equivalent because the first block previously appended into an empty array and the second still runs only when `parts.is_empty()`.
- **desktop/frontend/composer/codex_request.mbt** — Third copy of "collect the String members of a JSON array" in this file
  approval_details now calls json_string_parts too.
- **desktop/frontend/quick_open/state.mbt** — Manual accumulator that is exactly Array::filter
  `items.filter(item => !item.path.is_empty())`; filter preserves order, which rank_symbols' source-index tiebreak depends on.
- **desktop/frontend/transcript/component/rendering.mbt** — A push loop selecting matching projects is exactly `Array::filter`
  `projects.filter(project => project.matches(self.query))`.
- **desktop/frontend/grep_search/view.mbt** — Bounded take loop in visible_file_groups is iter().take(n)
  `group.matches.iter().take(remaining).to_array()`; `remaining > 0` already guaranteed by the guard on the line above, and `remaining -= rows.length()` is unaffected.
- **desktop/frontend/grep_search/view.mbt** — Same bounded take loop in visible_semantic_file_groups
  Same edit on `group.hits`.
- **desktop/frontend/transcript/sessions.mbt** — Hand-written minimum of two lengths; `Int::min` exists and is used elsewhere here
  `left.length().min(right.length())`.
- **desktop/frontend/project_picker/path_input.mbt** — Appending the contains bucket with a push loop where an array spread reads it
  `[..prefixes, ..contains]`; both are function-local arrays, so returning a fresh concatenation is observationally identical and keeps the bucket order the doc comment describes.

## Dropped, and why

- **enabled_indices is a filtered index comprehension (desktop/frontend/action_menu/action_menu.mbt:317)**
  `moon fmt` is not idempotent on this shape and its second pass produces code that would not compile. I wrote the finding's form (`[for index, entry in self.entries if entry is Action(item) && item.enabled => index]`); it compiled and passed `moon check --target js --deny-warn`, but `moon fmt` rewrote the bare-identifier body into a block `=> { index }`, and `moon fmt --check` then demanded `=> { { index, } }` — i.e. the formatter reinterprets the block as a struct literal, which is not an Int. CI runs `moon fmt --check`, so any form the formatter cannot round-trip is a landmine. Reverted the hunk to the original push loop; the finding was only worth 2 lines. Note this is a formatter bug, not a MoonBit-semantics problem: every other comprehension I added has a non-identifier body (`segment.to_owned()`, `@pathx.Relative(path)`, `part`) and survives `moon fmt --check` unchanged.

## Verification

All commands run from the worktree at .../scratchpad/wt/fe-pkgs (branch simplify/fe-pkgs, clean at start, based on origin/main).

BASELINE (before any edit)
- `cd desktop && moon check --target js --deny-warn` → "Finished. moon: ran 589 tasks, now up to date" (clean).
- `moon test --target js` over the 17 packages I was about to touch (openseek_desktop/frontend/{transcript/component, transcript, dock, skills, action_menu, fileeditor, fileeditor/feedback, files, codex, markdown, terminal, preview, mention_context, composer, quick_open, grep_search, project_picker}) → "Total tests: 419, passed: 419, failed: 0."

PER-PACKAGE during work
- `moon check --target js --deny-warn frontend/<pkg>` after each batch of edits; all clean. Note: `-p openseek_desktop/frontend/dock` does NOT work for `moon check` on this toolchain (it canonicalizes the package name as a filesystem path and fails); the working form is a path relative to cwd, e.g. `moon check --target js --deny-warn frontend/dock`. `-p <full-package-name>` does work for `moon test`.
- One real failure, fixed: `moon check` rejected the first `enabled_indices` comprehension with "Parse error, unexpected token `{`, you may expect `=>`" (2 errors) plus an `unused_value` on `index`. I rewrote it in `=>` form, which then compiled clean — and later dropped it entirely for the fmt reason below.

FINAL GATE (all from desktop/)
- `moon fmt --check` → "Finished. moon: ran 25 tasks, now up to date" (clean). It FAILED twice before that, both times on action_menu.mbt's `enabled_indices`; I confirmed by diffing the file against `_build/wasm-gc/release/format/.../action_menu.mbt` that the formatter wanted `=> { { index, } }`, i.e. its own output re-read as a struct literal. Reverted that hunk; fmt --check then passed. (Separately: `moon fmt --check`'s failure output includes "failed to execute `git --no-pager diff --color=always --no-index ...`" because this repo routes git diff through an external differ — that message is noise, the format diff itself is real. I used plain `diff -u` against the _build copy to read it.)
- `moon check --target js --deny-warn` → "Finished. moon: ran 434 tasks, now up to date" (clean; js is the frontend's only supported target).
- `moon check --deny-warn` (native, the module's preferred target) → "Finished. moon: ran 423 tasks, now up to date" (clean).
- `moon test --target js` over the same 17 packages → "Total tests: 419, passed: 419, failed: 0." Identical to baseline: 419 before, 419 after.

TEST DEDUPE COUNTS
- codex/activity_wbtest.mbt: 4 tests before → 4 tests after; 4 assert_true before → 4 after; all four test names, mutations, statuses and expected marks preserved. Each keeps its own `test` block, so a failure still names the case.
- transcript/component/component.mbt: 3 tests before → 3 after; assertion count unchanged (7 + 2 + 6 = 15 before and after); no inputs or expectations altered, only the shared literals hoisted.
- Package-level confirmation: `moon test --target js -p openseek_desktop/frontend/codex` → 70/70 both before and after the dedupe.

MBTI / PUBLIC INTERFACE
- `moon info frontend/<17 packages>` (SCOPED, never bare) → "Finished. moon: ran 4 tasks"; git status showed NO .mbti change. I then verified this was not a false negative: I appended a probe line to desktop/frontend/dock/pkg.generated.mbti and re-ran `moon info frontend/dock`, which printed "Warning: Skipping package `openseek_desktop/frontend/dock` for `moon info`: it does not support target backend `native`" and left the probe in place. So plain `moon info` genuinely cannot regenerate js-only packages on this toolchain (matching the brief's note). Probe reverted with `git checkout --`.
- Because of that I verified the interfaces a second way: captured `moon info --target js` (report mode; it does not write) over the same 17 packages AFTER my changes, then `git stash push -- desktop/frontend`, captured the same report on the pristine tree, then `git stash pop`. `diff -u info-before.txt info-after.txt` → a single differing line, moon's own "ran 35 tasks" vs "ran 43 tasks" progress line. The 2023 lines of interface report are byte-identical. No public signature changed.

GIT
- `git status --short` before staging listed exactly the 25 files I edited and nothing else (no build artifacts, no stray .mbti).
- Staged by explicit path (25 paths, no `git add -A`), committed as 3a8ba70f4 "refactor(desktop/frontend): reach for core APIs and shared helpers" with the Co-Authored-By and Claude-Session footer from the brief. 25 files changed, 155 insertions(+), 432 deletions(-).
- `git push -u origin simplify/fe-pkgs` → "* [new branch] simplify/fe-pkgs -> simplify/fe-pkgs", tracking set. No PR opened, no merge, no other branch touched. Working tree clean afterwards.

## Worth a second look

Four things worth a second look.

1. dock/view.mbt icon deletions cost some prose. The four deleted codicon wrappers carried per-call-site notes ("the strip's new-tab button", "the launcher's Files row", "the launcher's workspace Search row", "for the panel toggle"). The load-bearing part of those comments — the CC BY 4.0 attribution and the "an innerHTML-bearing element cannot be safely morphed by the positional child diff" rationale — is stated verbatim in the icons package header (icons/icons.mbt:1-9), so nothing explaining WHY was lost; the what/where notes were. If you'd rather keep them, this is the one finding to revert. The `icon` wrapper deletion loses nothing at all (its doc comment was byte-identical to the one on @icons.icon). Aesthetic side effect: `@icons.icon(@icons.icon_close_all)` at view.mbt:241 reads a bit heavy.

2. Double::max vs `if a > b`. terminal/resize.mbt and action_menu now use `Double::max`, which differs from the hand-written comparison only when an operand is NaN. Every operand comes from getBoundingClientRect / clientY / window.innerWidth, all finite. Same for core's `Double::clamp` debug `guard! min <= max`: I kept the `ceiling` / `.max(min_*)` lines specifically so that guard cannot trip.

3. mention_context.mbt is a real dead-code deletion, not just a rewrite. The removed arm had a different payload (`path.to_owned()` without `MentionAttribute::decode`) but was unreachable behind an identical condition on an unchanged `line` whose first arm always `continue`s. Deleting it leaves observable behaviour exactly as it runs today, but if the intent was ever that the two arms test *different* patterns, this deletion buries that bug instead of fixing it. Worth a glance from whoever wrote it.

4. Test dedupe scope. For the transcript component fixture I deliberately stopped at component.mbt. The finding's evidence also points at view_tests.mbt:189-203 (same Input literal modulo `focused=@interop.ChannelId::Local`) and two more literals there that differ in several fields; folding those into `{ ..base_input(), ... }` would have been a larger, less obviously-equivalent edit for little gain. `base_input()` / `pinned_model()` are package-private, so a follow-up can pick them up from view_tests.mbt whenever someone wants to.

Also note: `moon fmt` reformatted two of my comprehensions into brace-body form (fileeditor/view.mbt path_segments, fileeditor/files.mbt filter_map). That is the formatter's own output and `moon fmt --check` is clean on it — but see the `dropped` entry for the case where that same rewrite produces non-compiling code.

### DIFF
--- a/desktop/frontend/action_menu/action_menu.mbt
+++ b/desktop/frontend/action_menu/action_menu.mbt
@@ -231,17 +231,6 @@ fn set_style(element : @dom.Element, key : String, value : String) -> Unit {
 }
 
 ///|
-fn clamp(value : Double, low : Double, high : Double) -> Double {
-  if value < low {
-    low
-  } else if value > high {
-    high
-  } else {
-    value
-  }
-}
-
-///|
 fn position(target : String, anchor : Anchor) -> @cmd.Cmd {
   @cmd.custom_cmd(
     _ => {
@@ -285,26 +274,14 @@ fn position(target : String, anchor : Anchor) -> @cmd.Cmd {
         }
       }
       let margin = 4.0
+      let min_left = viewport_left + margin
+      let min_top = viewport_top + margin
       let max_left = viewport_left + viewport_width - menu_width - margin
       let max_top = viewport_top + viewport_height - menu_height - margin
-      let left = clamp(
-        wanted_left,
-        viewport_left + margin,
-        if max_left < viewport_left + margin {
-          viewport_left + margin
-        } else {
-          max_left
-        },
-      )
-      let top = clamp(
-        wanted_top,
-        viewport_top + margin,
-        if max_top < viewport_top + margin {
-          viewport_top + margin
-        } else {
-          max_top
-        },
-      )
+      // A menu taller or wider than the viewport pins to the near edge: the
+      // `.max(min_*)` keeps the clamp window non-empty.
+      let left = wanted_left.clamp(min=min_left, max=max_left.max(min_left))
+      let top = wanted_top.clamp(min=min_top, max=max_top.max(min_top))
       set_style(menu, "left", "\{left.to_int()}px")
       set_style(menu, "top", "\{top.to_int()}px")
       menu.set_attribute("data-positioned", "true")
@@ -326,12 +303,7 @@ fn Definition::enabled_indices(self : Definition) -> Array[Int] {
 
 ///|
 fn enabled_position(indices : Array[Int], index : Int) -> Int? {
-  for position, candidate in indices {
-    if candidate == index {
-      return Some(position)
-    }
-  }
-  None
+  indices.search(index)
 }
 
 ///|
--- a/desktop/frontend/codex/activity_wbtest.mbt
+++ b/desktop/frontend/codex/activity_wbtest.mbt
@@ -30,27 +30,36 @@ test "Codex activity accepts idle catalog evidence without a run mark" {
 }
 
 ///|
-test "Codex activity rejects active snapshot older than completion" {
+/// Take a frontier, let `mutate` record newer evidence on the ledger, then
+/// reconcile a one-thread snapshot carrying `status` against that now-stale
+/// frontier — the shared shape of the four fencing tests below.
+fn stale_snapshot_mark(
+  mutate : (CodexActivityLedger) -> CodexActivityLedger,
+  status : CodexThreadRuntime,
+) -> @conversation.RunMark {
   let activity = CodexActivityLedger::empty()
   let frontier = activity.frontier()
-  let completed = activity.started("thread-1").completed("thread-1")
-  let reconciled = completed.reconcile_snapshot(
-    [("thread-1", Some(ThreadActive))],
-    frontier,
+  mutate(activity)
+  .reconcile_snapshot([("thread-1", Some(status))], frontier)
+  .presentation("thread-1")
+}
+
+///|
+test "Codex activity rejects active snapshot older than completion" {
+  let mark = stale_snapshot_mark(
+    activity => activity.started("thread-1").completed("thread-1"),
+    ThreadActive,
   )
-  assert_true(reconciled.presentation("thread-1") is @conversation.NoRunMark)
+  assert_true(mark is @conversation.NoRunMark)
 }
 
 ///|
 test "Codex activity rejects active snapshot older than idle notification" {
-  let activity = CodexActivityLedger::empty()
-  let frontier = activity.frontier()
-  let idle = activity.status_changed("thread-1", Some(ThreadInactive))
-  let reconciled = idle.reconcile_snapshot(
-    [("thread-1", Some(ThreadActive))],
-    frontier,
+  let mark = stale_snapshot_mark(
+    activity => activity.status_changed("thread-1", Some(ThreadInactive)),
+    ThreadActive,
   )
-  assert_true(reconciled.presentation("thread-1") is @conversation.NoRunMark)
+  assert_true(mark is @conversation.NoRunMark)
 }
 
 ///|
@@ -99,24 +108,18 @@ test "failed newer Codex snapshot does not fence older catalog evidence" {
 
 ///|
 test "Codex activity rejects idle snapshot older than start" {
-  let activity = CodexActivityLedger::empty()
-  let frontier = activity.frontier()
-  let running = activity.started("thread-1")
-  let reconciled = running.reconcile_snapshot(
-    [("thread-1", Some(ThreadInactive))],
-    frontier,
+  let mark = stale_snapshot_mark(
+    activity => activity.started("thread-1"),
+    ThreadInactive,
   )
-  assert_true(reconciled.presentation("thread-1") is @conversation.RunningMark)
+  assert_true(mark is @conversation.RunningMark)
 }
 
 ///|
 test "Codex activity rejects snapshots from an earlier connection" {
-  let activity = CodexActivityLedger::empty()
-  let frontier = activity.frontier()
-  let reconnected = activity.disconnected()
-  let reconciled = reconnected.reconcile_snapshot(
-    [("thread-1", Some(ThreadActive))],
-    frontier,
+  let mark = stale_snapshot_mark(
+    activity => activity.disconnected(),
+    ThreadActive,
   )
-  assert_true(reconciled.presentation("thread-1") is @conversation.NoRunMark)
+  assert_true(mark is @conversation.NoRunMark)
 }
--- a/desktop/frontend/composer/codex_request.mbt
+++ b/desktop/frontend/composer/codex_request.mbt
@@ -7,6 +7,14 @@ priv struct CodexRequestQuestion {
 }
 
 ///|
+/// The string members of a JSON array; every other value is ignored.
+fn json_string_parts(values : Array[Json]) -> Array[String] {
+  [
+    for value in values if value is String(part) => part
+  ]
+}
+
+///|
 /// Render only the approval facts needed for a decision. The complete payload
 /// remains available below for diagnostics and fields this client does not yet
 /// understand.
@@ -75,12 +83,7 @@ fn CodexRequestInput::approval_details(
     )
   }
   if fields.get("proposedExecpolicyAmendment") is Some(Array(values)) {
-    let parts : Array[String] = []
-    for value in values {
-      if value is String(part) {
-        parts.push(part)
-      }
-    }
+    let parts = json_string_parts(values)
     if !parts.is_empty() {
       details.push(
         @html.div(class="codex-request-detail", [
@@ -140,25 +143,17 @@ fn CodexRequestInput::execpolicy_label(
   self : CodexRequestInput,
   decision : Json,
 ) -> String {
-  let parts : Array[String] = []
+  let mut parts : Array[String] = []
   if decision is Object(decision_fields) &&
     decision_fields.get("acceptWithExecpolicyAmendment")
     is Some(Object(amendment)) &&
     amendment.get("execpolicy_amendment") is Some(Array(values)) {
-    for value in values {
-      if value is String(part) {
-        parts.push(part)
-      }
-    }
+    parts = json_string_parts(values)
   }
   if parts.is_empty() &&
     self.params is Object(fields) &&
     fields.get("proposedExecpolicyAmendment") is Some(Array(values)) {
-    for value in values {
-      if value is String(part) {
-        parts.push(part)
-      }
-    }
+    parts = json_string_parts(values)
   }
   if parts.is_empty() {
     "Allow by rule"
--- a/desktop/frontend/dock/state.mbt
+++ b/desktop/frontend/dock/state.mbt
@@ -74,13 +74,13 @@ pub fn State::empty() -> State {
 /// The strip's file paths, in strip order — the projection the root update
 /// hands to fileeditor's `Ctx` for its stat/watch/reload sweeps.
 pub fn file_paths(state : State) -> Array[@pathx.Relative] {
-  let paths : Array[@pathx.Relative] = []
-  for tab in state.tabs {
+  state.tabs.filter_map(tab => {
     if tab is File(path~) {
-      paths.push(path)
+      Some(path)
+    } else {
+      None
     }
-  }
-  paths
+  })
 }
 
 ///|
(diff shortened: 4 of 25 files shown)


bobzhang:

> The worktree is CLEAN against all seven acceptance criteria. The branch (commit 3a8ba70f4, parent a227f6aae, working tree at HEAD, no untracked files) touches exactly the 25 in-scope packages, all .mbt, with no .mbti change and no whitespace-only churn.
> 
> Empirical gates: `moon check --target js` passes for packages without test entry points (action_menu, transcript); `moon test --target js` passes for every other changed package: codex 70/70, fileeditor 149/149, dock 17/17, preview 12/12, markdown 11/11, quick_open 18/18, project_picker 21/21, composer 18/18, mention_context 11/11, grep_search 11/11, files 3/3, fileeditor/feedback 2/2, skills 11/11, terminal 11/11, transcript/component 19/19.
> 
> Criterion-by-criterion verdict: (1) Behaviour identical — every hunk was compared to its pre-image: the clamp rewrites in action_menu, editor_resize and terminal/resize preserve bounds exactly (old high argument was max(min, max), reproduced as `.max(min_*)`); an mbtx probe confirmed core Double::clamp and Double::max match the deleted hand-rolled bodies for in-range/below/above/NaN inputs, including the NaN-ceiling case where `Double::max` (unlike the old `if max > floor`… also identical) — no divergence found anywhere including short-circuit orders (all_digits via all(), any()-based colour lookup vs break loop, take(n) loops vs bounded push, search/search_by vs index loops), snapshot semantics of filter/filter_map over arrays, the mention_context dead-arm removal (provably unreachable), and the String::split-based path_segments (split_once-head-drop ≡ split-then-filter-empty

Scope: 25 files reviewed; build `pass (moon check --target js: action_menu, transcript)`, tests `pass (moon test --target js, all changed packages; 412 tests total)`.

---

bobzhang:
## Held for your review

### Why it is held

**The reviewer's own verdict was `hold`,** on a robustness regression:

`desktop/frontend/action_menu/action_menu.mbt:283` replaces a hand-rolled clamp with core's `Double::clamp`:

```
wanted_left.clamp(min=min_left, max=max_left.max(min_left))
```

Core's `Double::clamp` **aborts with an uncatchable panic when `min > max`**, which is what any NaN bound produces (`NaN <= NaN` is false). The removed hand-rolled clamp returned the input unchanged in that case. So if `window.visualViewport`'s `offsetLeft`/`offsetTop`/`width`/`height` ever reads back as NaN, the old code placed the menu slightly wrong and the new code panics the after-render command.

Not reachable through any input the reviewer could construct, and arguably a latent bug either way. But it converts a silent misplacement into a crash, in UI positioning code fed by a browser API — exactly the kind of trade a maintainer should choose deliberately.

All 54 hunks checked against core's source: `Array::get` returns `None` for negative indices (so `changes.get(index - 1)` at index 0 is safe), `Array::search`/`search_by` return the first match, `Map::get_or_init` returns the live stored array so `items_for` keeps its aliasing, `StringView::all` iterates the same chars, `String::split("")` yields one empty segment so `path_segments` still drops it. The four dock icons moved to `@icons` are byte-identical path data; the `mention_context` deletion removes the second of two identical `file_pattern` branches, and the surviving one is the reachable one.

PR #1304, opened 2026-09-05
refactor: agent_tool/mbtx, sandbox: reuse stdlib and collapse repeated tests

bobzhang:

Candidates came from a read-only survey of the whole repository. The author was told to drop any that would not compile or would change behaviour, so this branch carries fewer than were proposed: **11 applied, 0 dropped, net -111 lines.**

## Applied

- **agent_tool/mbtx/mbtx_test.mbt** — 16 tests repeat the same execute-dispatch block that `run_in` already contains
  Helper named `dispatch(definition, arguments)` rather than `respond_of`; `run_in` now delegates to it. All 16 sites confirmed identical in shape (3 of them wrapped the args across lines) and rewritten mechanically. Went one step further than the finding: 5 sites whose block tail was a bare `output` after the new `let output = dispatch(...)` had the redundant binding collapsed into a tail expression.
- **agent_tool/mbtx/internal/decode/decode.mbt** — Five `decode rejects ...` tests have byte-identical bodies; collapse to one table
  Same five Json inputs, same order, same try/catch/assert_true body. Added `msg="accepted \{arguments.stringify()}"` so a failing row is identifiable — plain `\{arguments}` tripped deny-warn (deprecated Show-for-debugging), hence `.stringify()`. The two reject tests that also pin message wording were left separate, and a comment now says why.
- **agent_tool/internal/source_write_policy/source_write_policy_test.mbt** — Five identical is_protected_workspace_source_path assertions fit the loop this test already uses
  Loop variable is `guarded`, not `protected`: `protected` is a MoonBit reserved keyword and failed to compile. Added `msg=guarded` for row identifiability, and a comment preserving the point that `_build.mbt` / `.mooncakes.mbt` are protected because only a whole path COMPONENT counts as generated.
- **agent_tool/internal/auto_check/parse_gate.mbt** — `loc_line_span reads spans...` is a subset of the lenient-form table 130 lines below
  Deleted the duplicate test; its two inputs not already present in some spelling ("9:1-7:2", "abc:1") moved into the surviving table beside their equivalents. The deleted inline comment about reversed spans clamping is already stated in the surviving test's doc comment.
- **agent_tool/internal/source_write_policy/source_write_policy.mbt** — is_protected_source_path hand-rolls basename matching that @workspace_path.path_basename does
  DEVIATED from the proposal. `path_basename` strips trailing slashes, so it would flip `pkg/moon.pkg/` from unprotected to protected — and two of the six callers (agent_tool/shell/sandbox.mbt:3578 and :3955-3956) pass raw shell arguments, where a trailing slash is ordinary. Instead collapsed the ten `== name || has_suffix("/name")` lines into one `.any` over a new private `protected_manifest_names` list. Byte-identical behaviour for every input; a comment records why path_basename was rejected.
- **agent_tool/mbtx/mbtx_wbtest.mbt** — Two allowlist tests hand-roll the same filter loop over spawnable_commands
  Added `spawn_prefixes(program)` using the guarded array comprehension `[for entry in spawnable_commands if entry.0 == program => entry.1]`. Test-only file, so no interface surface.
- **agent_tool/internal/auto_check/auto_check.mbt** — containing_dir is a copy of @workspace_path.parent_dir, comment included
  Now `@workspace_path.parent_dir(path).unwrap_or(".")`. The two differ for "", "/", a bare drive, and trailing-slash inputs; none is reachable, because the sole caller `moon_check_cwd` gates on `is_relevant_moonbit_path` first. Private fn, single caller. Added the moon.pkg import. The old comment about the leading slash was replaced with one explaining why the `None` fallback is safe, not deleted.
- **agent_tool/internal/auto_check/auto_check.mbt** — is_relevant_moonbit_path re-implements manifest basename matching available as path_basename
  Same private-fn / single-caller argument as containing_dir. All 7 existing assertions re-verified by hand and by the suite. `is_moonbit_source_path` now sees the unnormalized path, which is fine: it only tests `.mbt` / `.mbt.md` suffixes.
- **agent_tool/mbtx/mbtx.mbt** — The denial-guidance append is written three times; it deserves a helper like with_build_notes
  Added `with_denial_guidance`, placed immediately before `with_build_notes` whose shape it mirrors. All three sites replaced; the surrounding `is_error=` conditions were left untouched.
- **agent_tool/internal/sandbox/worker_profile.mbt** — worker_write_profile_data re-derives denial subjects that source_write_denial_subjects already builds
  Same package-private helper, same [root, base-if-different] order; the later `.git` literal still appends last. `@workspace_path` is still imported for `is_under_workspace_root` above, so no unused import.
- **agent_tool/mbtx/wasm_policy.mbt** — env_set's copy loop is Map::map
  `inject.map((_, value) => value.to_json())`. Confirmed in core (builtin/linked_hash_map.mbt:878) that Map::map copies the linked structure, so insertion order — which is what the emitted policy JSON's key order depends on — is preserved, and the returned map is fresh so the two policy writes after it still cannot reach `inject`.

## Verification

All commands run from the worktree root /private/tmp/.../scratchpad/wt/mbtx with moon 0.1.20260827.

BASELINE (before any edit)
- `moon test agent_tool/internal/source_write_policy agent_tool/internal/auto_check agent_tool/internal/sandbox agent_tool/mbtx/internal/decode` -> "Total tests: 75, passed: 75, failed: 0."
- `moon test agent_tool/mbtx` (run detached, ~several minutes; it compiles real snippets) -> exit 0, "Total tests: 72, passed: 72, failed: 0."

AFTER
- `moon fmt` then `moon fmt --check` -> exit 0, no reformatting pending.
- `moon check agent_tool/mbtx agent_tool/mbtx/internal/decode agent_tool/internal/auto_check agent_tool/internal/sandbox agent_tool/internal/source_write_policy agent_tool/shell --deny-warn` -> "Finished. moon: ran 8 tasks, now up to date" (clean). agent_tool/shell was added on purpose: it is the only external caller of the source_write_policy function I touched.
- Same package list with `--target wasm --deny-warn` -> clean. All five touched packages declare supported_targets = "+wasm+native"; none is under desktop/, so no js run was needed.
- `moon test agent_tool/internal/source_write_policy agent_tool/internal/auto_check agent_tool/internal/sandbox agent_tool/mbtx/internal/decode` -> "Total tests: 70, passed: 70, failed: 0." 75 -> 70 is exactly the intended collapse: decode 5 tests -> 1, parse_gate 1 test -> 0. No test lost coverage.
- `moon test agent_tool/mbtx` -> "Total tests: 72, passed: 72, failed: 0." Unchanged, as expected: the mbtx_test/mbtx_wbtest work removed duplication inside test bodies, not tests.
- `moon info agent_tool/mbtx agent_tool/mbtx/internal/decode agent_tool/internal/auto_check agent_tool/internal/sandbox agent_tool/internal/source_write_policy` (SCOPED, never bare) -> ran clean and rewrote NO .mbti. `git status --short` after it listed only the 11 files I edited, so no public interface moved.

TEST-DEDUPE COUNTS (per the brief)
- decode.mbt: 5 assertions before (5 separate tests, one assert_true each) -> 5 rows after, each carrying its Json input in msg=.
- source_write_policy_test.mbt: 5 assertions before -> 5 rows after, each carrying its path in msg=.
- parse_gate.mbt: the deleted test's 6 assertions were all already pinned in the surviving table except two spellings ("9:1-7:2", "abc:1"); the table went from 12 rows to 14, so 6 assertions before -> 6 facts still covered, 2 of them newly written out.

WORKAROUNDS / FAILURES HIT
- First `moon check` after the test edits failed with 4 deny-warn errors, all fixed rather than worked around: `protected` is a MoonBit reserved keyword (renamed the loop variable to `guarded`), and `"\\{arguments}"` on a Json triggered the deprecated Show-for-debugging warning (switched to `arguments.stringify()`).
- The first scripted pass over mbtx_test.mbt matched only 13 of the 16 dispatch sites (three write `Async(execute) =>` with the call on the next line); the script asserted the count and wrote nothing, and was rerun with a tolerant pattern that got all 16. The resulting diff was read line by line before committing.

## Worth a second look

1. `is_protected_source_path` (source_write_policy.mbt) is where I deliberately did NOT do what the finding proposed. Using `@workspace_path.path_basename` there would strip trailing slashes and so newly classify `pkg/moon.pkg/` as protected. That predicate is a write-policy gate, and two of its six callers (agent_tool/shell/sandbox.mbt:3578 and :3955-3956) feed it raw shell arguments, where a trailing slash is completely ordinary — so the change would have been a real behaviour change in a security check, even though it errs toward denying. I kept the existing normalization and only collapsed the repetition into `.any` over a named list. Verify that the `.any` predicate `path == name || path.has_suffix("/\{name}")` really is the same test the ten deleted lines performed, for all five names.

2. The same trailing-slash argument does NOT apply to `is_relevant_moonbit_path` and `containing_dir` in auto_check.mbt, and that is why I did use `path_basename` / `parent_dir` there. Both are private, both have exactly one caller (`moon_check_cwd`), and the path reaching them is a file that was just written, filtered through `is_relevant_moonbit_path` first. If you disagree that those inputs are unreachable, these two hunks are the ones to revert — they are self-contained, and reverting them also makes the new `workspace_path` line in auto_check/moon.pkg unnecessary.

3. `worker_write_profile_data` now calls `source_write_denial_subjects`, whose name says "source_write_" at a worker call site. The behaviour is identical (same order, same basename-if-different rule) but the name now reads oddly; renaming it to something neutral would touch its other caller in source_write_profile.mbt, so I left it and put a comment at the call site instead. Worth deciding whether you'd rather have the rename.

Minor: `agent_tool/internal/auto_check/moon.pkg` gained one import line. workspace_path depends only on moonbitlang/x/path, so there is no cycle, and both targets check clean.

### DIFF
--- a/agent_tool/internal/auto_check/auto_check.mbt
+++ b/agent_tool/internal/auto_check/auto_check.mbt
@@ -158,6 +158,14 @@ async fn moon_check_cwd(path : String) -> String? {
 
 ///|
 fn is_relevant_moonbit_path(path : String) -> Bool {
+  // A path that ends in a separator names a DIRECTORY, never a manifest, and
+  // `path_basename` strips that separator before reporting the last component
+  // — so "pkg/moon.pkg/" would come back as "moon.pkg". Rule those out first
+  // to keep this predicate's domain exactly what it was when the manifest
+  // names were matched as whole final components.
+  if path.has_suffix("/") || path.has_suffix("\\") {
+    return false
+  }
   // `path_basename` treats backslashes as separators, so the manifest names
   // match on Windows too; the source suffixes below contain no separator and
   // so are unaffected by them.
@@ -182,6 +190,12 @@ fn containing_dir(path : String) -> String {
 
 ///|
 test "relevant path detection covers MoonBit source and manifests" {
+  // A trailing separator makes it a directory, not a manifest: the predicate
+  // must answer the same here as it did before `path_basename` (which strips
+  // that separator) was introduced.
+  assert_false(is_relevant_moonbit_path("x/moon.pkg/"))
+  assert_false(is_relevant_moonbit_path("moon.mod/"))
+  assert_false(is_relevant_moonbit_path("x\\moon.work\\"))
   assert_true(is_relevant_moonbit_path("moon.mod"))
   assert_true(is_relevant_moonbit_path("lib/moon.pkg"))
   assert_true(is_relevant_moonbit_path("moon.work"))
--- a/agent_tool/internal/auto_check/auto_check.mbt
+++ b/agent_tool/internal/auto_check/auto_check.mbt
@@ -158,13 +158,10 @@ async fn moon_check_cwd(path : String) -> String? {
 
 ///|
 fn is_relevant_moonbit_path(path : String) -> Bool {
-  let path = path.replace_all(old="\\", new="/")
-  path == "moon.mod" ||
-  path.has_suffix("/moon.mod") ||
-  path == "moon.pkg" ||
-  path.has_suffix("/moon.pkg") ||
-  path == "moon.work" ||
-  path.has_suffix("/moon.work") ||
+  // `path_basename` treats backslashes as separators, so the manifest names
+  // match on Windows too; the source suffixes below contain no separator and
+  // so are unaffected by them.
+  @workspace_path.path_basename(path) is ("moon.mod" | "moon.pkg" | "moon.work") ||
   is_moonbit_source_path(path)
 }
 
@@ -175,14 +172,12 @@ fn is_moonbit_source_path(path : String) -> Bool {
 
 ///|
 fn containing_dir(path : String) -> String {
-  let path = path.replace_all(old="\\", new="/")
-  // An empty part before the last slash means the slash was the leading one,
-  // so the containing directory is the root itself rather than "".
-  match path.rev_split_once("/") {
-    Some(([], _)) => "/"
-    Some((dir, _)) => dir.to_owned()
-    None => "."
-  }
+  // `parent_dir` already keeps a leading slash as the root rather than
+  // collapsing it to "", and normalizes backslashes. Its `None` cases ("",
+  // ".", "/", a bare drive) are unreachable here: the only caller has already
+  // passed `path` through `is_relevant_moonbit_path`, so it ends in a manifest
+  // name or a source suffix.
+  @workspace_path.parent_dir(path).unwrap_or(".")
 }
 
 ///|
--- a/agent_tool/internal/auto_check/moon.pkg
+++ b/agent_tool/internal/auto_check/moon.pkg
@@ -2,6 +2,7 @@ import {
   "bobzhang/openseek/agent_tool/internal/manifest_path",
   "bobzhang/openseek/agent_tool/internal/moon_check",
   "bobzhang/openseek/agent_tool/shell",
+  "bobzhang/openseek/internal/workspace_path",
   "moonbitlang/async",
   "moonbitlang/async/fs",
   "moonbitlang/core/json",
--- a/agent_tool/internal/auto_check/parse_gate.mbt
+++ b/agent_tool/internal/auto_check/parse_gate.mbt
@@ -455,17 +455,6 @@ test "format_parse_gate_errors caps at the display limit and handles bad locs" {
 }
 
 ///|
-test "loc_line_span reads spans, single points, and rejects junk" {
-  assert_true(loc_line_span("7:1-9:2") is Some((7, 9)))
-  assert_true(loc_line_span("7:1") is Some((7, 7)))
-  assert_true(loc_line_span("7") is Some((7, 7)))
-  // A reversed span clamps to the start line rather than going negative.
-  assert_true(loc_line_span("9:1-7:2") is Some((9, 9)))
-  assert_true(loc_line_span("") is None)
-  assert_true(loc_line_span("abc:1") is None)
-}
-
-///|
 /// Real-moonc round trips: unlike `moon check`, the standalone gate can run
 /// inside `moon test` safely (no project build lock is involved).
 async test "content_parse_errors gates real content through moonc" {
@@ -593,7 +582,9 @@ test "loc_line_span accepts every lenient loc form" {
   debug_inspect(loc_line_span("7:1-x"), content="Some((7, 7))")
   debug_inspect(loc_line_span("7:1-"), content="Some((7, 7))")
   debug_inspect(loc_line_span("9:2-7:1"), content="Some((9, 9))")
+  debug_inspect(loc_line_span("9:1-7:2"), content="Some((9, 9))")
   debug_inspect(loc_line_span("x"), content="None")
+  debug_inspect(loc_line_span("abc:1"), content="None")
   debug_inspect(loc_line_span(""), content="None")
   debug_inspect(loc_line_span(":1"), content="None")
 }
--- a/agent_tool/internal/sandbox/worker_profile.mbt
+++ b/agent_tool/internal/sandbox/worker_profile.mbt
@@ -76,16 +76,13 @@ fn worker_write_profile_data(
     $|(version 1)
     $|(allow default)
     $|
-  let denial_subjects : Array[String] = []
+  // Same path-then-basename subjects the source-write profile names, in the
+  // same order; the `.git` literal below still appends after them.
+  let denial_subjects = source_write_denial_subjects(deny_roots)
   for root in deny_roots {
     profile <+
       $|(deny file-write* (subpath "\{sbpl_string_escape(root)}"))
       $|
-    denial_subjects.push(root)
-    let base = @workspace_path.path_basename(root)
-    if base != root {
-      denial_subjects.push(base)
-    }
   }
   profile <+
     $|(allow file-write* (subpath "\{sbpl_string_escape(worker_root)}"))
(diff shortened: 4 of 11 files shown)


bobzhang:

**This is the second audit.** The first found a real behaviour change: `path_basename` strips a trailing separator, so `pkg/moon.pkg/` — a directory the old whole-component match rejected — started reading as a manifest. Commit `779ef7d79` rules out separator-terminated paths up front, restoring the old domain, with three cases pinning it. The re-audit confirms the widening is gone.

**Verdict: signed off.** 1 finding(s), none blocking (1 low); each is answered below.

> Audit of the worktree at HEAD 779ef7d79 (base a227f6aae + refactor cd5a0631c + follow-up 779ef7d79; `git status` clean, no untracked files). Verdict: the branch is clean against the acceptance criteria — no blocker, one low-severity criterion-6 note below.
> 
> Criterion 1 (BEHAVIOUR IDENTICAL) — PASS, empirically grounded. All five touched-package test suites pass at HEAD (auto_check 32, source_write_policy 11, decode 7, sandbox 20, mbtx 79). The riskiest hunks were the auto_check rewrites, because commit cd5a0631c alone changed the manifest predicate's domain (trailing-separator paths) and commit 779ef7d79 ("fix(auto_check): keep the manifest predicate's exact domain") restored it; I differentially probed the FINAL head code against the base implementations in the writable scratch lab over 33,350 generated path strings (mixed / and \\ separators, trailing separators, double slashes, drive forms, JSON manifest spellings, .mbt/.mbt.md/.mbti suffix edges): 0 predicate mismatches and 0 containing_dir mismatches across the 20,226 reachable predicate-true inputs. The remaining hunks are mechanical with no reordering of effects: mbtx.mbt's three denial-guidance append sites all route through the new with_denial_guidance with the old composition order intact (is_error flags, build-notes prepend, output-limit/binary notes all untouched); wasm_policy env_set's Map::map yields the same fresh map and same keys; worker_profile's denial_subjects now come from source_write_denial_subjects, which pushes path-then-basename per root in the same order as the removed inline loop, with the `.

Scope: 11 files reviewed; build `pass (moon test compiled all five touched packages)`, tests `149 passed / 0 failed (auto_check 32, source_write_policy 11, decode 7, sandbox 20, mbtx 79)`.

### Findings and responses

**[low] `agent_tool/internal/source_write_policy/source_write_policy.mbt:93` — is_protected_source_path rewrite is a defensible but borderline substitution, not an unqualified simplification**

The vetted stdlib-reuse item (replace the hand-rolled basename matching with @workspace_path.path_basename) could not be applied without changing behaviour — path_basename strips a trailing slash, and this public predicate intentionally treats 'pkg/moon.pkg/' as a directory, not a manifest. The author correctly dropped it and instead hoisted the six manifest names into a module-level constant and iterated with `.any`, preserving the old disjuncts exactly (verified: the three suffix checks are untouched, and the six names are mutually exclusive, so evaluation order is irrelevant). Verdict on criterion 6 is genuinely borderline: the change single-sources each name (previously spelled twice …

*Response:* Kept, deliberately. The old predicate spelled each of the five manifest names twice (`path == N` and `has_suffix("/N")`), ten disjuncts for five facts; the new form states each name once and adding a manifest is a one-line change instead of two. That single-sourcing, plus the `.any` functional form, is exactly the dedupe this sweep set out to do, and the auditor confirms the semantics are identical and evaluation order irrelevant. The alternative it offers — the flat disjunction — is the thing the PR is removing.

---

bobzhang:
## Held for your review

### Why it is held

**A predicate that gates whether `moon check` runs at all after an agent write was rewritten.** `agent_tool/internal/auto_check/auto_check.mbt:160`, `is_relevant_moonbit_path`, swaps a whole-or-suffix chain for `@workspace_path.path_basename` plus a trailing-separator guard. A domain slip here fails **silently** — the agent writes a file, no check runs, and nothing says so.

The first audit of this branch caught a real widening here (`path_basename` strips a trailing separator, so `pkg/moon.pkg/` started reading as a manifest); commit `779ef7d79` added the guard and three regression cases. This second, independent reviewer then enumerated **~199k generated paths** — manifest/source/`C:`/dot components joined by both separator kinds, with every leading and trailing separator combination — and found old and new disagree on **zero** of them, and likewise for `containing_dir` over the subset the single caller can reach.

So the evidence is strong. It is still the highest-consequence hunk in the whole set, and it has already been wrong once, which is why I would rather you signed it off than me.

### Second thing: five reject tests became one table

`decode.mbt`'s five independent rejection tests (unknown warning value, non-string cwd, unknown target, missing source, non-object) are now one loop. Same inputs, same expectations, same count — but the first failing case aborts the rest, so a regression spanning several shapes reports only one. Native/js/wasm test-block counts drop by 5.

### One concrete defect worth fixing either way

`agent_tool/mbtx/mbtx.mbt:1079` — the new `with_denial_guidance` doc comment lists three exit paths that append denial guidance: "the background handoff, a finished run, and a timeout". The background-handoff path **does not** append it; it returns at `mbtx.mbt:742` with `respond(with_build_notes(...))`. Documentation only, but it would send a reader chasing code that does not exist.

bobzhang:

**Fixed** the one concrete defect: the `with_denial_guidance` doc comment now names the three paths that actually append guidance (a run that finished inside the handoff grace on a job runtime, a finished foreground run, a foreground timeout) and says the background handoff does not, because its program is still running.

**On the held predicate** (`is_relevant_moonbit_path`): with the trailing-separator guard in front, `path_basename` reduces to backslash-normalize then split on the last `/`, so `basename is ("moon.mod" | "moon.pkg" | "moon.work")` is exactly `== name || has_suffix("/name")` on the normalized path. `is_moonbit_source_path` only tests suffixes that contain no separator, so passing it the raw path changes nothing. `parent_dir`'s `None` cases (`""`, `"."`, `"/"`, a bare drive, trailing separator) all need an input the predicate has already rejected.

**On the five-case reject table:** each row's `msg` carries its input, so a failing row is identifiable. Accepting the fold.

The reversed-span comment dropped from the deleted `loc_line_span` test survives in the surviving table's doc comment. Merging once CI is green on the doc fix.

PR #1305, opened 2026-09-05
refactor: agent_tool/shell: collapse repeated test assertions and fold a duplicated git parser

bobzhang:

Candidates came from a read-only survey of the whole repository. The author was told to drop any that would not compile or would change behaviour, so this branch carries fewer than were proposed: **9 applied, 1 dropped, net -781 lines.**

## Applied

- **agent_tool/shell/sandbox_wbtest.mbt** — 105 repeated tree_target/dyn_tree_target assertions in "sandbox preblocks common source-writing commands"
  Applied, but NOT as proposed. The surveyor described one uniform shape; there are actually four (tree_target/dyn_tree_target x blocked/allowed), plus 4 assertions fitting none of them. I added four local `async fn` helpers that return Bool rather than the proposed `-> Unit` helpers that swallow the assert, so every case keeps its own assert_true line and a failure still names one command (brief's diagnosability rule). Done mechanically by script, then machine-verified: each of the 105 assertions was reduced to a canonical (detector, cmd, root, cwd, expected) tuple before and after, and the ordered lists are identical. 101 rewritten; 4 left spelled out (2 use cwd Some("/tmp") or Some("{real_root}-external"), 2 use the `is Some(_)` shape). All 70 let bindings identical, all 26 original comments preserved. Test body 877 -> 308 lines.
- **agent_tool/shell/moon_auto_check.mbt** — "detect moon command check cwd from -C option" repeats moon_auto_check_start_cwd(cmd, Some(base))
  33 assertions before, 33 one-line rows after (the finding said 30). Local `fn cwd_for(cmd, base)`. Kept `base` an explicit positional argument rather than defaulting it, so the one row using base="/repo/pkg" ("moon -C ../other info") stays visible instead of becoming a hidden exception. Order and inputs unchanged.
- **agent_tool/shell/review_guard.mbt** — "read-only guard blocks mutating moon anywhere" repeats the 6-line review_read_only_blocks call
  15 assertions before, 15 rows after (the finding said 17). Uses the identical `fn blocks(cmd : String) -> Bool` helper the very next test in the same file already defines. All four inline prose comments kept in position.
- **agent_tool/shell/internal/git_policy/git_policy.mbt** — `parse` re-implements `text_subcommand_index`, whose comment says "Mirrors `parse`"
  Applied. Verified the two textual differences are behaviour-identical rather than taking the surveyor's word: (a) parse's explicit `--` arm — in the scanner `--` is not matched by attached_cwd_option/attached_work_tree_option/global_no_arg_option/global_option_with_attached_value, so it reaches `has_prefix("-") && arg != "-"` and returns None; (b) parse's `index + 1 >= len -> None` — the scanner's `index += 2` steps past `end` and the while-guard returns None. Rewrote the helper's doc comment to record both facts instead of deleting it.
- **agent_tool/shell/shell_test.mbt** — Four worker-mode tests repeat the same 9-line WorkerSandbox + run construction
  Applied. The proposed signature did not compile twice: `async fn` tripped deny-warn unused_async [0067] (the function is not itself async), then the bare return type failed with [4122] because `fail` raises. Final form is `fn worker_run(root : String) -> (async (Json) -> @agent_tool.ToolAction) raise` — the parens are required so `raise` attaches to worker_run, not to the returned closure. All four blocks were byte-identical; replaced in place so side-effect order is untouched.
- **agent_tool/shell_exec/sink.mbt** — Hand-written cap clamping in the sink constructor is Int::max
  Applied at all three sites the finding's evidence names, not just the constructor: inline_cap/hard_cap -> .max(1), read_tail's start offset -> (self.file_bytes - want).max(0) (Int64::max), and incomplete_suffix_len's lookback -> len.min(3). The constructor's doc comment ("Non-positive caps are clamped to 1") still describes the code exactly, so it stands unchanged.
- **agent_tool/shell/shell.mbt** — Three-branch chunk-limit chain is exactly Int::clamp(min=1, max=1024)
  Applied verbatim as proposed.
- **agent_tool/shell/moon_auto_check.mbt** — moon_check_output_budget is remaining.clamp(min=0, max=MaxOutputChars)
  Applied verbatim as proposed. The three existing assertions in "bound follow-up check output by remaining shell output budget" still pass.
- **agent_tool/shell/internal/shell_parse/parser.mbt** — render_source builds parts with three push-loops that are two maps and a spread
  Applied. Kept the loop variable names (assignment, redirect) as the lambda parameters rather than the proposed one-letter a/r, so the interpolations read the same as before. Group order preserved.

## Dropped, and why

- **"submodule verbs" test is a strict subset of the submodule rows in the preceding preblock test (agent_tool/shell/internal/git_policy/git_policy_test.mbt:227)**
  Trips two of the brief's explicit drop criteria. (1) It deletes an explanatory comment: the test carries a four-line doc comment recording WHY verb-level submodule classification is exercised through the public preblock surface ("the sandbox layer keeps its own private mirror") — that rationale exists nowhere else and dies with the test. (2) It is not actually a strict subset. I checked all 16 rows against the earlier test: the 9 not-preblocked rows are verbatim duplicates, but only 5 of the 7 preblocked rows are; ["git","submodule","foreach","echo","hi"] and ["git","submodule","--quiet","foreach","echo","hi"] appear upstream only with the payload word "hello". The finding argues the payload is never inspected — true of today's submodule_verb_is_unsafe/submodule_foreach_is_help_request, but that is a claim about the implementation, not an equivalence a reviewer can see, and the brief forbids tidying a case away. Net saving was 34 lines; not worth either cost.

## Verification

Scope: the four packages touched — agent_tool/shell, agent_tool/shell/internal/git_policy, agent_tool/shell/internal/shell_parse, agent_tool/shell_exec. All are +wasm+native (shell_parse declares no supported_targets, so it builds on all); none is a desktop/frontend js package, so no --target js run was needed.

BASELINE (clean tree, before any edit): `moon test agent_tool/shell agent_tool/shell/internal/git_policy agent_tool/shell/internal/shell_parse agent_tool/shell_exec` -> "Total tests: 200, passed: 200, failed: 0." One benign libtool warning about an empty archive TOC in moonbitlang/async/types — present before and after, unrelated.

AFTER all edits:
- `moon fmt` then `moon fmt --check` -> "Finished. moon: ran 1740 tasks, now up to date", no diff produced. moon fmt reflowed my git_policy guard onto one line and re-added a trailing comma in the Invocation literal; I kept its output.
- `moon check <4 pkgs> --deny-warn` (native) -> "Finished. moon: no work to do" (clean).
- `moon check <4 pkgs> --target wasm --deny-warn` -> "Finished. moon: ran 48 tasks, now up to date" (clean).
- `moon test <4 pkgs>` -> "Total tests: 200, passed: 200, failed: 0." Same 200 before and after, as expected: the dedupes collapse assertions inside existing tests, they do not add or remove tests.
- `moon info <4 pkgs>` (SCOPED, never bare) -> "Finished. moon: ran 46 tasks, now up to date". `git status --short` afterwards listed only my 8 source files and ZERO .mbti files, confirming no public interface moved (agent_tool/shell/pkg.generated.mbti in particular is untouched).

ASSERTION COUNTS for the three test dedupes (assertions before -> rows after):
- sandbox_wbtest.mbt: 105 -> 105 (101 collapsed to helper calls, 4 left in long form). Verified by script, not by eye: I extracted a canonical (detector, cmd, root, cwd, expected) tuple for every assertion from the original file and from the rewritten file (expanding the four helpers back to their bodies) and compared the ordered lists — "SEMANTICALLY IDENTICAL", 105 vs 105, zero mismatches. Separately verified 70/70 let bindings identical and 26/26 comments preserved, 0 missing.
- moon_auto_check.mbt: 33 -> 33.
- review_guard.mbt: 15 -> 15.

THINGS THAT FAILED AND HOW I WORKED AROUND THEM:
1. `worker_run` did not compile as the finding proposed. First attempt `async fn worker_run(...) -> async (Json) -> ToolAction` failed deny-warn with unused_async [0067] (the function is not itself async). Dropping `async` then failed with [4122] "Function with error can only be used inside a function with error types in its signature" because `fail` raises. Fixed by parenthesizing the returned function type and annotating the outer fn: `fn worker_run(root : String) -> (async (Json) -> @agent_tool.ToolAction) raise`. Two attempts, within budget, so kept rather than dropped.
2. `git push -u origin simplify/shell` was REJECTED as non-fast-forward: a stale remote branch of the same name already existed, last touched 2026-07-02 by the same user, 1 commit ahead / 1367 behind my base. I did not force blindly. I checked it with `gh pr list --head simplify/shell --state all`: its PR is #298, state MERGED (squash-merged, which is why its SHA is not an ancestor of main). No open PR references the branch. I then pushed with `git push -u --force-with-lease=simplify/shell:bca4c650d origin simplify/shell`, pinning the lease to the exact stale SHA so a concurrent update would have aborted it. Result: "+ bca4c650d...f17a909f3 simplify/shell -> simplify/shell (forced update)". `git ls-remote origin simplify/shell` confirms the tip is f17a909f3. A reviewer should be aware a merged leftover ref was overwritten.

Final `git status --short`: clean. Staged by 8 explicit paths, never `git add -A`; no build artifacts or .mbti files entered the commit. No PR opened, nothing merged, no other branch touched.

## Worth a second look

1. The force-push. `simplify/shell` already existed on origin as a leftover from PR #298 (MERGED, 2026-07-02). I overwrote it with --force-with-lease after confirming the PR was merged and no open PR pointed at it. If that branch was being kept deliberately, the old tip was bca4c650d and it is recoverable from the reflog / PR #298.

2. git_policy `parse` now delegates to `text_subcommand_index`. This is the only change with real behavioural surface — it is a public function on the security-policy path, and the equivalence rests on two non-obvious arguments rather than on identical text: an explicit `--` reaching the scanner's trailing `has_prefix("-") && arg != "-"` arm, and a dangling value option stepping `index` past `end`. I traced all five option predicates (global_option_consumes_next, global_option_with_attached_value, global_no_arg_option, attached_cwd_option, attached_work_tree_option) against the input "--" to confirm none of them swallow it first. Worth re-deriving. Coverage: "parse fails closed" (git_policy_test.mbt:10-16, exercising --bogus, --help, --, -c, bare git) and "parse finds the subcommand past global options". Note `parse` also feeds `should_preblock`, so a regression here would silently un-block git forms.

3. sandbox_wbtest.mbt is a 794-line diff and cannot be reviewed line-by-line by hand. It was produced by script and verified by tuple comparison rather than by reading — the verification is the thing to check, not the diff. To re-run it independently: for each assertion, reduce it to (detector, cmd, root, cwd, expected) and confirm the before/after lists match in order. The four helpers deliberately return Bool instead of asserting internally so a failing case still reports its own line; that differs from the in-repo `async fn blocked(...) -> Unit` idiom at shell_test.mbt:2247, which the brief's diagnosability rule argues against.

Also note I applied the sink.mbt finding at three sites, not just the one in its `current` snippet — the other two were named in that finding's own evidence. And I deliberately did not shorten render_source's lambda parameters to one letter, so the interpolated strings read identically to the loop bodies they replaced.

### DIFF
--- a/agent_tool/shell/internal/git_policy/git_policy.mbt
+++ b/agent_tool/shell/internal/git_policy/git_policy.mbt
@@ -18,32 +18,10 @@ pub struct Invocation {
 /// to a value option, or an unrecognized `-option` (fail closed). Aliases are not
 /// resolved, so e.g. `git co` reports subcommand `co`.
 pub fn parse(argv : Array[String], arg_start : Int) -> Invocation? {
-  let mut index = arg_start
-  while index < argv.length() {
-    let arg = argv[index]
-    if arg == "--" || arg == "--help" || arg == "-h" {
-      return None
-    }
-    if arg == "-C" || arg == "--work-tree" || global_option_consumes_next(arg) {
-      if index + 1 >= argv.length() {
-        return None
-      }
-      index += 2
-      continue
-    }
-    if attached_cwd_option(arg) is Some(_) ||
-      attached_work_tree_option(arg) is Some(_) ||
-      global_option_with_attached_value(arg) ||
-      global_no_arg_option(arg) {
-      index += 1
-      continue
-    }
-    if arg.has_prefix("-") && arg != "-" {
-      return None
-    }
-    return Some({ subcommand: arg, subcommand_index: index, })
+  guard text_subcommand_index(argv, arg_start, argv.length()) is Some(index) else {
+    return None
   }
-  None
+  Some({ subcommand: argv[index], subcommand_index: index, })
 }
 
 ///|
@@ -541,8 +519,11 @@ fn clean_arg_is_exclude(arg : String) -> Bool {
 }
 
 ///|
-/// Find the git subcommand's index in a flat `words` list (the TooComplex text
-/// path), scanning `[start, end)` and skipping global options. Mirrors `parse`.
+/// Find the git subcommand's index in a flat `words` list, scanning
+/// `[start, end)` and skipping global options. The scanner behind both the
+/// TooComplex text path and `parse`. An explicit `--` is rejected by the
+/// trailing unrecognized-option arm, and a value option with no argument runs
+/// `index` past `end` and so falls out as `None`.
 fn text_subcommand_index(words : Array[String], start : Int, end : Int) -> Int? {
   let mut index = start
   while index < end {
--- a/agent_tool/shell/internal/shell_parse/parser.mbt
+++ b/agent_tool/shell/internal/shell_parse/parser.mbt
@@ -152,17 +152,11 @@ fn render_source(
   argv : Array[String],
   redirects : Array[Redirect],
 ) -> String {
-  let parts : Array[String] = []
-  for assignment in env {
-    parts.push("\{assignment.name}=\{assignment.value}")
-  }
-  for arg in argv {
-    parts.push(arg)
-  }
-  for redirect in redirects {
-    parts.push("\{redirect.op} \{redirect.target}")
-  }
-  parts.join(" ")
+  [
+    ..env.map(assignment => "\{assignment.name}=\{assignment.value}"),
+    ..argv,
+    ..redirects.map(redirect => "\{redirect.op} \{redirect.target}"),
+  ].join(" ")
 }
 
 ///|
--- a/agent_tool/shell/moon_auto_check.mbt
+++ b/agent_tool/shell/moon_auto_check.mbt
@@ -48,13 +48,7 @@ fn insert_before_system_footer(content : String, block : String) -> String {
 ///|
 fn moon_check_output_budget(content : String, max_output_chars : Int) -> Int {
   let remaining = max_output_chars - content.char_length() - 1
-  if remaining <= 0 {
-    0
-  } else if remaining < @moon_check.MaxOutputChars {
-    remaining
-  } else {
-    @moon_check.MaxOutputChars
-  }
+  remaining.clamp(min=0, max=@moon_check.MaxOutputChars)
 }
 
 ///|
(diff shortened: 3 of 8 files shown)


bobzhang:

> Clean pass against all seven acceptance criteria — no blockers, no findings of any severity. The worktree (HEAD f17a909f3, diff base a227f6aae, git status clean) is a pure simplification of the four listed packages; only the 8 in-scope files changed, no .mbti touched (criterion 2).
> 
> Evidence gathered empirically:
> 1. Behaviour identical: every rewrite was compared old-vs-new. Int::max/min/clamp rewrites in sink.mbt/shell.mbt/moon_auto_check.mbt reproduce the branch chains exactly (incl. boundary cases remaining<=0->1, equality->clamp bound, equal/less-case of file_bytes-want -> 0). The git_policy `parse` fold into the pre-existing text_subcommand_index is equivalent: parse passes end=argv.length(), the removed missing-value guard's `None` is reproduced by the loop falling out past end, and an explicit `--` is not matched by any option helper so it falls to the trailing unrecognized-option arm -> None exactly as before; the updated doc comment states this correctly. parser render_source preserves env->argv->redirects order. Test-only folds are pure one-to-one call substitutions.
> 2. Public interface: no .mbti changed; `parse` keeps its pub signature.
> 3. Comments: explanatory comments survive (review_guard and sandbox_wbtest section comments, git_policy parse/scan doc updated and still accurate).
> 4. Coverage unchanged: sandbox_wbtest giant test has exactly 105 assertion lines before and after, matched case-for-case in order (tree vs dyn_tree_target selection and expected paths preserved); review_guard's 15 literal cases identical (extracted and diffed); moon_auto_ch

Scope: 8 files reviewed; build `moon check -p agent_tool/shell, -p agent_tool/shell_exec: OK`, tests `shell 221/221 (native & wasm), shell_exec 16/16, git_policy 23/23, shell_parse 24/24`.

---

PR #1307, opened 2026-09-05
refactor: agent_tool file/misc tools: reuse stdlib, drop re-implemented helpers, dedupe tests

bobzhang:

Candidates came from a read-only survey of the whole repository. The author was told to drop any that would not compile or would change behaviour, so this branch carries fewer than were proposed: **16 applied, 2 dropped, net -180 lines.**

## Applied

- **agent_tool/agent_tool.mbt** — 11 test tool literals restate the AgentToolDefinition constructor's defaults; one helper replaces them
  Applied as proposed. Added a private `test_tool(name?, description?, schema?, execute?)` helper that calls the real constructor, so `control: false, concurrent_safe: false` is no longer restated 11 times. Every non-default value in each original literal (names, descriptions, executors, the two-key schema) is passed through unchanged, so the test data is identical.
- **agent_tool/agent_tool.mbt** — function_tools "converts the registry" test is a strict subset of "preserves description and schema"
  Applied. Deleted the block; its one unique assertion `assert_eq(function_tools.length(), 1)` now sits in "preserves description and schema", which already builds the same single-"echo" registry and asserts the name.
- **agent_tool/agent_tool.mbt** — brief_line's char-capped truncation loop is String::from_iter(one.iter().take(limit))
  Applied. `one` is a StringView; `.iter()` is Iter[Char], so the cap still counts and cuts on characters exactly as the loop did. Trimmed the proposed two-line comment to one line because the function's own doc comment already states the surrogate-pair rationale.
- **agent_tool/plan/steps/decode_test.mbt** — 11 decode-rejection tests differ only in arguments/expected; collapse to one table
  Applied with two adjustments. (a) Used the repo's own shape - successive `assert_decode_error(...)` calls inside one test, as at agent_tool/plan/plan.mbt:299 - rather than the proposed `for` over an Array[(Json, String)]. (b) Folded only the NINE tests whose body was a single call with no prose comment. Left separate: "rejects a blank title" (NBSP/ideographic-space scenario note), "rejects a multi-line title" (forged-row note), and the two computed-input cap tests. 9 assert_decode_error calls before, 9 after; package 28 -> 20 test blocks.
- **agent_tool/plan/steps/decode.mbt** — in_progress_count accumulator in decode is Array::any over the steps built so far
  Applied. The check moved before the push so `steps` holds only earlier steps; the second in_progress step still raises, with the same message, and no later item is decoded. O(n^2) in principle, but n <= MAX_STEPS = 20.
- **agent_tool/edit/edit_test.mbt** — Three symlink parse-gate tests have identical bodies; only target/link names differ
  Applied. Extracted `async fn assert_link_edit_gated(prefix~, target~, link~)`. The three tests stay separate and each keeps its own doc comment (the first test's inline comment was promoted to a `///` doc comment above it, not deleted). Test count unchanged at 66.
- **agent_tool/edit/edit.mbt** — inserted_line_span re-implements total_line_count verbatim (same file)
  Applied. Body is now `total_line_count(new_string)`. Kept and extended the doc comment so it names total_line_count and records why the one differing rule (empty content spans 0 lines) cannot fire on the insertion path.
- **agent_tool/edit/edit.mbt** — Whitespace membership written as 4-way `==` disjunctions instead of an alternation pattern
  Applied to all THREE sites, one more than the proposal spelled out: the two in find_whitespace_normalized_candidates and the third in normalize_whitespace that the finding's evidence named. Deliberately not Char::is_ascii_whitespace, which would also match \u{0B}/\u{0C}.
- **agent_tool/write/internal/decode/decode_test.mbt** — Four try/catch/noraise reject tests; the sibling decode packages already have the helper
  Applied. Declared the same `assert_decode_error` helper the edit/multi_edit decode packages carry (these are separate internal packages, so it must be local) and folded the four tests into one. 4 rejection cases before, 4 after; package 11 -> 8 test blocks. All four asserted substrings preserved verbatim.
- **agent_tool/read/internal/decode/decode_test.mbt** — Four try/catch/noraise reject tests collapse into one table-driven test
  Applied, same shape as the write sibling. 4 cases before, 4 after; package 12 -> 9 test blocks. Asserted substrings unchanged.
- **agent_tool/read/internal/decode/decode.mbt** — cap_max_output_chars hand-rolls Int::min
  Applied verbatim: `value.min(hard_max_output_chars)`.
- **agent_tool/read/read.mbt** — "read parses and caps optional limits" is a strict subset of the decode package's own test
  Applied: block deleted. Verified read/internal/decode/decode_test.mbt:2-19 decodes the byte-identical arguments and debug_inspects all four fields, with max_output_chars pinned at 50000 == @decode.hard_max_output_chars. The deleted test never called execute. Package 23 -> 22 test blocks.
- **agent_tool/job_output/job_output.mbt** — page_of hand-rolls Iter drop/take over chars; String::from_iter does it in one line
  Applied. Confirmed the offset can never be negative (window_offset at job_output.mbt:229 rejects `offset < 0`), which is the only input where drop/take would diverge from the index loop.
- **agent_tool/write_scope/write_scope.mbt** — prefix_admits hand-rolls Array::starts_with (length guard + index loop)
  Applied, with one correction: the proposed `parts.starts_with(prefix[:])` fails --deny-warn with `unnecessary_view_op` (the parameter is already a view, so MoonBit inserts the conversion). Final form is `self.allowed.any(prefix => parts.starts_with(prefix))`.
- **agent_tool/multi_edit/multi_edit.mbt** — group_contiguous indexes groups[count - 1] where an array pattern reads directly
  Applied. `groups is [.., last]` subsumes the `count > 0` guard (an empty array does not match), and `last.items` is the same Array object the index form mutated.
- **agent_tool/multi_edit/multi_edit.mbt** — check_contiguity uses Map[String, Unit] as a set; core's Set is in the prelude
  Applied even though it saves no lines: it makes the existing "Set of paths whose contiguous block has already ended" comment literally true. Set needed no moon.pkg import, as the finding predicted.

## Dropped, and why

- **Sandbox-denial classification is written twice (job_output running path and bgjobs terminal path)**
  Would add a NEW PUBLIC function `classify_denial` to agent_tool/bgjobs. job_output is a different package, so the shared helper cannot be private - it must be `pub`, which expands bgjobs's visible API and would rewrite agent_tool/bgjobs/pkg.generated.mbti. The brief forbids changing a public signature and treats a .mbti change as the signal to revert. Six lines saved is not worth permanently widening a package's public surface, especially since the two call sites read the rule off different types (BgJobSnapshot fields vs. the internal BgJob).
- **desktop/main.mbt: `match Option { Some(x) => x; None => default }` is `unwrap_or`**
  The only finding in the list tagged group "de-services" (every other one is "tool-files" or "tool-misc"), and the only one outside agent_tool. It is out of scope for this branch's stated subject (agent_tool file/misc tools) and, if the de-services branch carries the same finding, applying it here would duplicate the change and conflict on merge. The change itself looks correct and safe - it should land on the branch that owns desktop/.

## Verification

All commands run in the worktree /private/tmp/claude-501/-Users-hongbozhang-git-openseek/ee773eb3-cd66-4ac0-a331-789265556fad/scratchpad/wt/tools on branch simplify/tools (clean at start, base a227f6aae).

BASELINE (`moon test <pkg>` on the untouched tree; agent_tool.mbt was git-stashed to measure its own baseline, then popped): agent_tool 23; agent_tool/plan/steps 28; agent_tool/edit 66; agent_tool/write/internal/decode 11; agent_tool/read 23; agent_tool/read/internal/decode 12; agent_tool/job_output 14; agent_tool/write_scope 6; agent_tool/multi_edit 31. Aggregate over those plus agent_tool/bgjobs: "Total tests: 229, passed: 229, failed: 0".

AFTER: agent_tool 22; agent_tool/plan/steps 20; agent_tool/plan 12 (dependent, unchanged); agent_tool/edit 66; agent_tool/write 28 (dependent, unchanged); agent_tool/write/internal/decode 8; agent_tool/read 22; agent_tool/read/internal/decode 9; agent_tool/job_output 14; agent_tool/write_scope 6; agent_tool/multi_edit 31. Same aggregate command: "Total tests: 213, passed: 213, failed: 0". Zero failures throughout; no test ever failed during this task.

The 229 -> 213 delta is exactly the 16 test BLOCKS removed by the five dedupes, with no case lost: agent_tool -1 (subset block deleted, its unique length==1 assertion moved), plan/steps -8 (9 single-call rejection tests -> 1 test holding all 9 calls), write/internal/decode -3 (4 -> 1, 4 cases), read -1 (subset block deleted), read/internal/decode -3 (4 -> 1, 4 cases). Assertion/row counts for the collapses: plan/steps 9 assert_decode_error calls before, 9 rows after; write/internal/decode 4 before, 4 after; read/internal/decode 4 before, 4 after. agent_tool/edit stayed at 66 - that finding extracted a helper and kept all three tests.

GATE:
- `moon fmt` -> clean; reformatted nothing outside my edits (ran twice; `git status --short` listed only the 12 intended files both times).
- `moon check <11 touched+dependent pkgs> --deny-warn` (native) -> Finished, 0 warnings 0 errors.
- `moon check <same> --target wasm --deny-warn` -> Finished, 0 warnings 0 errors.
- `moon check --deny-warn` workspace-wide (native, 821 tasks) -> Finished, no warnings/errors. Run because brief_line, write_scope and agent_tool.mbt have many downstream consumers.
- `moon info agent_tool agent_tool/plan/steps agent_tool/plan agent_tool/edit agent_tool/write agent_tool/write/internal/decode agent_tool/read agent_tool/read/internal/decode agent_tool/job_output agent_tool/write_scope agent_tool/multi_edit` (SCOPED, never bare) -> Finished; `git status --short` afterwards showed the same 12 files and NO .mbti. Public interfaces unchanged, as required.

ONE FAILURE, CORRECTED RATHER THAN WORKED AROUND: the proposed `parts.starts_with(prefix[:])` in write_scope.mbt failed `moon check --deny-warn` with error 0075 `unnecessary_view_op: [:] is unnecessary here`. Dropped the explicit `[:]`; re-checked clean.

GIT: staged the 12 files by explicit path (no `git add -A`); committed as 06728a239 (12 files changed, 183 insertions, 363 deletions); `git push -u origin simplify/tools` created the branch on origin. Working tree clean afterwards. No PR opened, no merge, no other branch touched.

## Worth a second look

Four things are worth a second look.

1. agent_tool/plan/steps/decode.mbt - the one non-test change with an ordering subtlety. The in_progress check moved BEFORE `steps.push(step)`, so on the offending step `steps` holds only the earlier ones. Same rejection point (the second in_progress step), same message, and the raise discards `steps` either way. Complexity goes from O(n) to O(n^2), bounded by MAX_STEPS = 20.

2. agent_tool/edit/edit.mbt inserted_line_span now delegates to total_line_count. The bodies were identical except total_line_count's `if content == "" { return 0 }` guard, so the two differ ONLY on empty input, where inserted_line_span used to return 1. The claim that the insertion path never passes "" rests on edit.mbt:182-184 rejecting both-empty edits before line 212, which is the sole call site. If that rejection ever moves, this delegation changes behaviour - the doc comment now records the dependency.

3. agent_tool/write_scope/write_scope.mbt prefix_admits - `Array::starts_with` is `len >= prefix.len && elementwise Eq`, matching the loop including the short-parts early false. Worth confirming an empty prefix is impossible (validate_prefix at write_scope.mbt:266-275 never yields one); if one existed, both old and new forms return true, so it is a no-op either way.

4. The three test collapses fold N named tests into one. Diagnosability is preserved through the existing `msg="{message} should contain {expected}"` on assert_true, and I also changed each helper's unasserted no-raise `fail("expected decode error")` to `fail("expected decode error naming {expected}")` so a row that unexpectedly DECODES is identifiable too. Nothing asserts those strings. In plan/steps I deliberately did not fold the blank-title and multi-line-title tests: their bodies carry prose explaining a distinct scenario (Unicode-only whitespace; newline-forged plan rows), and that comment is the point of the separate test.

### DIFF
--- a/agent_tool/agent_tool.mbt
+++ b/agent_tool/agent_tool.mbt
@@ -194,16 +194,8 @@ pub fn brief_line(text : String, limit? : Int = 72) -> String {
   if one.length() <= limit {
     "\{one}"
   } else {
-    let kept = StringBuilder()
-    for ch in one; count = 0 {
-      if count >= limit {
-        break
-      }
-      kept.write_char(ch)
-      continue count + 1
-    }
-    kept.write_char('…')
-    kept.to_string()
+    // `iter()` yields characters, so `take` cuts on a character boundary.
+    "\{String::from_iter(one.iter().take(limit))}…"
   }
 }
 
@@ -362,34 +354,21 @@ test "ToolAction helpers build structured actions" {
 }
 
 ///|
-test "Tools::function_tools converts the registry to deepseek tool definitions" {
-  let tools = Tools([
-    {
-      name: "echo",
-      description: "Echo the input.",
-      schema: JsonSchema({ "type": "object" }),
-      execute: Sync(_args => respond("ok")),
-      control: false,
-      concurrent_safe: false,
-    },
-  ])
-  let function_tools = tools.function_tools()
-  assert_eq(function_tools.length(), 1)
-  assert_eq(function_tools[0].name, "echo")
+/// One minimal registry entry for the dispatch tests. The privileges
+/// (`control`, `concurrent_safe`) are the constructor's own defaults, so no
+/// test below has to restate them just to build an ordinary tool.
+fn test_tool(
+  name? : String = "echo",
+  description? : String = "Echo.",
+  schema? : Json = { "type": "object" },
+  execute? : ToolExecutor = Sync(_args => respond("ok")),
+) -> AgentToolDefinition {
+  AgentToolDefinition(name~, description~, schema=JsonSchema(schema), execute~)
 }
 
 ///|
 test "Tools::find returns the registered definition by name" {
-  let tools = Tools([
-    {
-      name: "echo",
-      description: "Echo.",
-      schema: JsonSchema({ "type": "object" }),
-      execute: Sync(_args => respond("ok")),
-      control: false,
-      concurrent_safe: false,
-    },
-  ])
+  let tools = Tools([test_tool()])
   guard tools.find("echo") is Some(definition) else { fail("missing echo") }
   assert_eq(definition.name, "echo")
   assert_true(tools.find("ghost") is None)
@@ -398,22 +377,8 @@ test "Tools::find returns the registered definition by name" {
 ///|
 test "Tools::Tools raises on duplicate tool names" {
   let _ = Tools([
-    {
-      name: "echo",
-      description: "First.",
-      schema: JsonSchema({ "type": "object" }),
-      execute: Sync(_args => respond("first")),
-      control: false,
-      concurrent_safe: false,
-    },
-    {
-      name: "echo",
-      description: "Second.",
-      schema: JsonSchema({ "type": "object" }),
-      execute: Sync(_args => respond("second")),
-      control: false,
-      concurrent_safe: false,
-    },
+    test_tool(description="First.", execute=Sync(_args => respond("first"))),
+    test_tool(description="Second.", execute=Sync(_args => respond("second"))),
   ]) catch {
     error => {
       assert_true("\{error}".contains("duplicate tool name: echo"))
@@ -437,19 +402,14 @@ async test "execute_tool_call reports unknown tool" {
 ///|
 async test "execute_tool_call dispatches to registered executor" {
   let tools = Tools([
-    {
-      name: "echo",
-      description: "Echo.",
-      schema: JsonSchema({ "type": "object" }),
-      execute: Sync(args => {
+    test_tool(
+      execute=Sync(args => {
         match args {
           { "value": String(v), .. } => respond(v)
           _ => respond("no value", is_error=true)
         }
       }),
-      control: false,
-      concurrent_safe: false,
-    },
+    ),
   ])
   let call = AgentToolCall(
     ToolCall(id="call_1", name="echo", arguments="{\"value\":\"hi\"}"),
@@ -472,14 +432,7 @@ async fn async_echo(args : Json) -> ToolAction {
 ///|
 async test "execute_tool_call dispatches to async executor" {
   let tools = Tools([
-    {
-      name: "echo",
-      description: "Async echo.",
-      schema: JsonSchema({ "type": "object" }),
-      execute: Async(async_echo),
-      control: false,
-      concurrent_safe: false,
-    },
+    test_tool(description="Async echo.", execute=Async(async_echo)),
   ])
   let call = AgentToolCall(
     ToolCall(id="call_1", name="echo", arguments="{\"value\":\"async-hi\"}"),
@@ -500,14 +453,11 @@ async test "execute_tool_call turns a tool failure into a recoverable error resu
   }
 
   let tools = Tools([
-    {
-      name: "boom",
-      description: "Always fails.",
-      schema: JsonSchema({ "type": "object" }),
-      execute: Async(async_boom),
-      control: false,
-      concurrent_safe: false,
-    },
+    test_tool(
+      name="boom",
+      description="Always fails.",
+      execute=Async(async_boom),
+    ),
   ])
   let call = AgentToolCall(ToolCall(id="call_1", name="boom", arguments="{}"))
   // The tool's failure is caught at the dispatch boundary and returned as an
@@ -521,19 +471,13 @@ async test "execute_tool_call turns a tool failure into a recoverable error resu
 ///|
 test "Tools::function_tools preserves description and schema" {
   let tools = Tools([
-    {
-      name: "echo",
-      description: "Echo what comes in.",
-      schema: JsonSchema({
-        "type": "object",
-        "properties": Json::empty_object(),
-      }),
-      execute: Sync(_args => respond("ok")),
-      control: false,
-      concurrent_safe: false,
-    },
+    test_tool(description="Echo what comes in.", schema={
+      "type": "object",
+      "properties": Json::empty_object(),
+    }),
   ])
   let function_tools = tools.function_tools()
+  assert_eq(function_tools.length(), 1)
   assert_eq(function_tools[0].name, "echo")
   assert_eq(function_tools[0].description, "Echo what comes in.")
   let text = function_tools[0].parameters.stringify()
@@ -551,30 +495,21 @@ test "Tools::Tools accepts an empty registry" {
 ///|
 test "Tools::function_tools yields one entry per definition" {
   let tools = Tools([
-    {
-      name: "first",
-      description: "First.",
-      schema: JsonSchema({ "type": "object" }),
-      execute: Sync(_args => respond("1")),
-      control: false,
-      concurrent_safe: false,
-    },
-    {
-      name: "second",
-      description: "Second.",
-      schema: JsonSchema({ "type": "object" }),
-      execute: Sync(_args => respond("2")),
-      control: false,
-      concurrent_safe: false,
-    },
-    {
-      name: "third",
-      description: "Third.",
-      schema: JsonSchema({ "type": "object" }),
-      execute: Sync(_args => respond("3")),
-      control: false,
-      concurrent_safe: false,
-    },
+    test_tool(
+      name="first",
+      description="First.",
+      execute=Sync(_args => respond("1")),
+    ),
+    test_tool(
+      name="second",
+      description="Second.",
+      execute=Sync(_args => respond("2")),
+    ),
+    test_tool(
+      name="third",
+      description="Third.",
+      execute=Sync(_args => respond("3")),
+    ),
   ])
   let function_tools = tools.function_tools()
   assert_eq(function_tools.length(), 3)
--- a/agent_tool/edit/edit.mbt
+++ b/agent_tool/edit/edit.mbt
@@ -851,19 +851,12 @@ fn select_edit_region(
 ///|
 /// How many lines `new_string` occupies once inserted: one per newline, plus
 /// one for a trailing fragment that does not end in a newline (that fragment
-/// shares its line with whatever follows the insertion point). `new_string`
-/// is non-empty on the insertion path — both-empty edits are rejected first.
+/// shares its line with whatever follows the insertion point). That is exactly
+/// `total_line_count`, whose only extra rule — empty content spans 0 lines —
+/// cannot apply here: `new_string` is non-empty on the insertion path, because
+/// both-empty edits are rejected first.
 fn inserted_line_span(new_string : String) -> Int {
-  let newlines = for ch in new_string; newlines = 0 {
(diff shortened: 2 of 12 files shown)


bobzhang:

**Verdict: signed off.** 2 finding(s), none blocking (2 nit); each is answered below.

> The worktree (clean, HEAD 06728a239 = refactor commit on base a227f6aae) satisfies all seven criteria; no blockers. Criterion 1 (behaviour identical): every hunk was reviewed against the base line-by-line. brief_line truncation → String::from_iter(iter().take(limit))+'…', page_of index loop → iter().drop/take, prefix_admits loop → Array::starts_with, cap_max_output_chars → Int.min, Map[String,Unit]→Set, groups[count-1] → [..,last] pattern, in_progress accumulator → steps.any, inserted_line_span → total_line_count, and three 4-way `==` whitespace disjunctions → alternation patterns are all 1:1 equivalent (no reordered short-circuits, no snapshot/overflow/arm-reachability hazards). The one theoretical divergence — old inserted_line_span("")=1 vs total_line_count("")=0 — is unreachable: the sole call site (edit.mbt:212, insertion branch) is guarded by the both-empty rejection at edit.mbt:182; the new doc comment states this. Equivalence of the from_iter/drop/take forms was additionally probe-verified across unicode, boundary offsets, and limit 0 in the scratch lab. Criterion 2: git diff a227f6aae touches only 12 .mbt files; no .mbti change, no pub-signature change (test_tool relies on AgentToolDefinition constructor defaults control=false/concurrent_safe=false, verified at agent_tool.mbt:75-84). Criterion 3: no explanatory why-comments deleted — the symlink-gate comments and per-test scenario notes were moved verbatim onto the retained tests; new comments (page_of char counting, decode second-in_progress rejection point, test_tool defaults, total_line_count divergence) are acc

Scope: 12 files reviewed; build `moon check: pass (all touched packages compile at HEAD)`, tests `HEAD: agent_tool 747, edit 133, plan/steps 20, read 43, read/internal/decode 9, write/internal/decode 8, job_output 14, multi_edit 46, write_scope 6 — all pass, 0 failures; base a227f6aae equivalent suites also all pass (763/136/28/50/12/11/14/46/6)`.

### Findings and responses

**[nit] `agent_tool/plan/steps/decode.mbt:30` — Accumulator→steps.any is O(n) per step (O(n²) total), though bounded by MAX_STEPS**

Criterion 1 & 6: the rewrite is behavior-identical (same rejection step — the second in_progress item — same message, same set of decode_step calls; the only difference is push-before-fail vs check-before-push, which is unobservable because fail() aborts). Informational nit: replacing the running counter with steps.any(s => ...) scans all earlier steps on every in_progress item, so worst case is quadratic instead of linear. MAX_STEPS caps n at 20 and the added comment documents the logic, so this is not a violation, just a note on the trade-off.

*Response:* Accepted, not changed. `MAX_STEPS` caps n at 20, so the worst case is 400 comparisons on a decode path that already allocates per step; the auditor classes it as a trade-off note rather than a violation, and the linear form it replaced is the thing this PR set out to remove.

**[nit] `agent_tool/read/internal/decode/decode_test.mbt:35` — assert_decode_error helper is now triplicated verbatim across sibling decode test packages**

Criterion 6 note, not a violation: the vetted item for write/internal/decode said 'the sibling decode packages already have the helper', but in this branch both read/internal/decode and write/internal/decode gain a new ~15-line copy of the helper that already existed in agent_tool/plan/steps and now also in the other sibling. Each file individually gets simpler (4 inline try/catch/noraise blocks → 1 helper + 4 one-liners), and the packages are separate modules with no shared test-support target, so cross-package sharing would need new structure. Flagged only as a future-dedupe observation.

*Response:* Acknowledged as a follow-up, not fixed here. The three `assert_decode_error` copies live in separate packages with no shared test-support target, so removing the duplication needs new structure (a testkit package or a shared internal one) — a design change, not a simplification. Each file is individually simpler than before.

---

bobzhang:
## Held for your review

### Why it is held

**This PR removes the most test blocks in the set — 16 of them** (native 3230 → 3214, and the same delta on js and wasm). Two kinds of removal, and both deserve your eye:

1. **Folds.** Nine decode-rejection tests in `agent_tool/plan/steps` become one table, and similar folds land in `read`, `write`, `multi_edit` and `edit`. Same inputs and expectations, but the first failing case now aborts the rest — break the "title present" and "status present" rules together and one run names only the first, where before it named both.
2. **Outright deletions justified as "strict subset".** `read parses and caps optional limits` and `Tools::function_tools converts the registry to deepseek tool definitions` were deleted on the grounds that a surviving test already covers them, and three symlink gate tests collapsed to one. Those subset claims are the thing to check.

### Two domain changes, both argued unreachable

- `agent_tool/edit/edit.mbt:858` — `inserted_line_span` now delegates to `total_line_count`, which returns **0** rather than **1** for the empty string. Unreachable (`edit.mbt:182` rejects `old_string == "" && new_string == ""` before the insertion path) and, the reviewer notes, also unobservable at the single call site — but the PR's justification is a reachability argument, which is the pattern worth distrusting.
- `agent_tool/job_output/job_output.mbt:250` — the old `page_of` loop compared against `offset + window`, which **wraps** for a near-INT_MAX offset; `drop(offset).take(window)` does not. `window_offset` only rejects `offset < 0`, so a model can pass 2_147_480_000. Both forms return "" for any real job, so nothing observable changes.

The reviewer wrote differential tests against the deleted implementations in a scratch worktree: `brief_line` matched the old form on 11 samples (emoji and surrogate pairs, CJK, CRLF) × 13 limits including −3 and 0; `page_of` matched on 10 samples × 13 offsets × 6 windows. `Array::starts_with` in core is line-for-line the hand-rolled loop it replaces, and the `Map[String, Unit]` → `Set[String]` swap keeps contains/add semantics with the O(1) comment still true.

bobzhang:

I re-read the diff against the concerns in the hold note.

- **Deleted tests were true subsets.** `read parses and caps optional limits` fed the same four fields to `@decode.decode` and asserted the same cap as `decode valid read input` in `read/internal/decode`, which pins the whole struct including `max_output_chars: 50000`. The two assertions in `Tools::function_tools converts the registry` (length 1, name `echo`) both now sit in `Tools::function_tools preserves description and schema`.
- **Symlink gate.** `assert_link_edit_gated` keeps the three distinct target/link spellings and each scenario's doc comment moved onto its retained test.
- **`inserted_line_span` → `total_line_count`.** Identical on non-empty input; the empty case is rejected before the insertion branch is reachable.
- **`page_of`.** `drop(offset).take(window)` returns `""` wherever the old wrapping `offset + window` did.
- **Folded tables.** Every `assert_decode_error` failure names the rule it expected, so a failing row is identifiable.

Nothing to change. Merging.

PR #1308, opened 2026-09-05
refactor: viz, deepseek: JSON object patterns and one shared filter-bar

bobzhang:

Candidates came from a read-only survey of the whole repository. The author was told to drop any that would not compile or would change behaviour, so this branch carries fewer than were proposed: **12 applied, 0 dropped, net -71 lines.**

## Applied

- **cmd/viz_app/app.mbt** — error_bar / escalated_bar / build_error_bar are three copies of one filter-bar shape
  Applied with adjustments. Added `fn filter_bar(emit, count, kind~, active~, count_label~, toggle_title~, toggle_label~, toggle~, jump~, jump_noun~)`; the three wrappers keep their ORIGINAL `if ... { return @html.nothing }` visibility tests verbatim (I did not adopt the proposed `guard` rewording — smaller diff, more obviously equivalent) and keep their doc comments unchanged. The load-bearing `// Jumps only make sense when there is something to jump to.` comment moved into filter_bar. Emitted class names are byte-identical (`error-*`, `escalated-*`, `build-error-*`), which I cross-checked against the CSS in web/index.html:276-415 — no stylesheet change needed.
- **cmd/viz_app/app.mbt** — resolve_session_key rewraps find_first's Option instead of Option::map
  Applied as proposed.
- **viz/view.mbt** — tool_call_is_escalated unwraps Object then get(); one object pattern says it directly
  Applied as proposed: `json is { "escalated": True, .. }`.
- **viz/view.mbt** — tool_call_is_run_failed's nested Object/get(target) match is one JSON object pattern
  Applied; `moon fmt` dropped the proposed parens, giving `json is { "target"? : None | Some(String("wasm") | Null), .. }`.
- **viz/view.mbt** — rendered_build_error_count's RawLog arm is a manual counter where its sibling already uses count_if
  Applied; `moon fmt` reformatted the predicate into a block lambda.
- **viz/view.mbt** — session_has_mbtx_activity is an early-return scan that Iter::any expresses directly
  Applied as proposed. Iter::any short-circuits, matching the early return; the `_ => false` arm keeps the match exhaustive.
- **viz/view.mbt** — pad2 reimplements String::pad_start
  Applied. Checked wall_clock_label's `guard unix_ms > 0` so every pad2 argument is in 0..99, where both forms agree (they differ only for negative n, which cannot occur).
- **viz_export/export.mbt** — escape_json_for_html hand-rolls String::replace_all over a single character
  Applied as proposed.
- **viz_export/export.mbt** — is_script_close_tail hand-rolls ASCII case folding that Char::to_ascii_lowercase does
  Applied as proposed. Equivalent because every element of `tail` is lowercase or '/': for a lowercase expected char both forms accept {c, upper(c)}, and for '/' both accept only '/'.
- **deepseek/json_encode.mbt** — ToolDefinition::to_json writes name/description/parameters twice for a single optional field
  Applied as proposed. Key insertion order (name, description, parameters, strict) is preserved and is pinned by the existing snapshot test "tool definitions encode as native DeepSeek tools", which still passes unchanged.
- **testkit/filesystem/outline.mbt** — sanitize_name test builds a 100-char string with a StringBuilder loop instead of String::repeat
  Applied as proposed. This is NOT a test dedupe — no test was collapsed; the same single test keeps all 5 of its assertions (3 assert_eq before the change site, plus assert_eq on length and assert_true on the suffix). The explanatory comment `// Over-long names are truncated with an ellipsis.` is kept.
- **inspect/main.mbt** — SessionCatalog::rows nests the whole row build in a match arm with an empty None arm
  Applied as proposed: `guard ... is Some(id) else { continue }`. The `continue` and the old empty `None => ()` arm are both no-ops for the loop iteration.

## Verification

All commands run in the worktree /private/tmp/.../scratchpad/wt/viz on branch simplify/viz (clean at start, HEAD a227f6aae).

BASELINE TESTS (before any edit), all passing:
- `moon test viz --target js` -> Total tests: 19, passed: 19, failed: 0
- `moon test viz_export` (native) -> 2/2
- `moon test deepseek` (native) -> 45/45
- `moon test testkit/filesystem` (native) -> 17/17
- `moon test inspect` (native) -> 23/23
- `moon test cmd/viz_app --target js` -> 31/31
Note: running `moon test` from inside cmd/viz_app/ or inspect/ runs the WHOLE moon.work workspace (3202 and 1964 tests). I discarded those and used the package-scoped invocations above from the workspace root instead; the numbers reported here are the scoped ones.

AFTER (same commands, identical results):
viz 19/19 (js), viz_export 2/2, deepseek 45/45, testkit/filesystem 17/17, inspect 23/23, cmd/viz_app 31/31 (js). No count changed, no test added or removed.
- `moon test viz_export deepseek testkit/filesystem inspect --target wasm` -> Total tests: 87, passed: 87, failed: 0 (= 2+45+17+23, matching native).

FORMAT: `moon fmt` -> "ran 1740 tasks, now up to date". It reflowed three of my edits (the deepseek Map spread, the count_if predicate, the `target` pattern parens); no file outside my six was modified (confirmed by `git status --short`).

CHECKS (all clean, no warnings):
- `moon check viz_export deepseek testkit/filesystem inspect --deny-warn` (native) -> "ran 54 tasks, now up to date"
- `moon check viz_export deepseek testkit/filesystem inspect --target wasm --deny-warn` -> "ran 3 tasks, now up to date"
- `moon check viz cmd/viz_app viz_export deepseek --target js --deny-warn` -> "ran 9 tasks, now up to date"
(Also ran `moon check cmd/viz_app --target js --deny-warn` and `moon check viz --target js --deny-warn` individually mid-way; both clean.)

MBTI: `moon info viz_export deepseek testkit/filesystem`, `moon info inspect`, and `moon info viz cmd/viz_app --target js`, all scoped — never a bare `moon info`. NO .mbti changed (`git status --short` listed only the six .mbt files after each run). The js-only run printed an informational "requested interfaces different from canonical backend Native" dump for viz/cmd/viz_app because those packages have no native backend; it wrote nothing. The one public signature I touched, `pub fn session_has_mbtx_activity(@agent_session.Session) -> Bool`, is unchanged in that dump.

GIT: staged the six files by explicit path (`git add <6 paths>`), committed 0cf9878d9. `git status --short` empty afterwards.

PUSH — DEVIATION, please read: `git push -u origin simplify/viz` was REJECTED (non-fast-forward). A branch named `simplify/viz` already exists on origin: tip ff7254031 "refactor(viz): dedupe label walkers + local simplification sweep", dated 2026-07-09, forked from an old main at 40dd96069, and NOT merged into origin/main (verified with `git merge-base --is-ancestor origin/simplify/viz origin/main` -> false). Overwriting it needs --force, which would discard an unmerged remote commit, so I did not do that. I pushed instead to a fresh, non-colliding remote branch: `git push -u origin simplify/viz:simplify/viz-filter-bar` -> "[new branch] simplify/viz -> simplify/viz-filter-bar". The local branch is still named simplify/viz and now tracks origin/simplify/viz-filter-bar. No PR opened, no merge, no other branch touched.

## Worth a second look

Three things to look at twice.

1. The push target is `origin/simplify/viz-filter-bar`, NOT `origin/simplify/viz`. A stale unmerged branch already occupied that name (see verification). If you want the intended name, someone with the context to decide has to delete or rebase ff7254031 first.

2. `filter_bar` in cmd/viz_app/app.mbt is the only finding with real reviewer surface. Nothing renders it in a unit test — the js suite (31 tests) exercises the model/update layer, and the only end-to-end coverage of these toolbars is the Playwright suite under cmd/viz_app/e2e, which needs a running app and is outside the brief's gate, so I did not run it. What I did instead: read the CSS in web/index.html (lines 276-278, 355-364, 378-387, 406-415) and confirmed the generated class names still match exactly — `error-bar/-count/-toggle/-jump`, `escalated-*`, `build-error-*`. The child ORDER in the array literal (count chip, toggle, prev, next) matches the original `children.push` order in all three, and the conditional `..if count > 0` spreads gate exactly the same elements the original `if count > 0` blocks did. Worth an eyeball on the rendered toolbar if anyone is already running the app.

3. `pad2` and `is_script_close_tail` are the two changes whose equivalence depends on an input-range argument rather than being syntactic. pad2 differs from the original for negative n ("0-5" vs "-5"); it cannot receive one today because `wall_clock_label` guards `unix_ms > 0`, but that is a caller-side invariant, not a type-level one. `is_script_close_tail`'s equivalence depends on every element of the local `tail` array being lowercase-or-'/', which it is, and which is local to the function.

### DIFF
--- a/cmd/viz_app/app.mbt
+++ b/cmd/viz_app/app.mbt
@@ -998,10 +998,7 @@ fn resolve_session_key(sessions : Array[SessionRow], value : String) -> String? 
   if sessions.iter().any(row => row.key == value) {
     return Some(value)
   }
-  match sessions.iter().find_first(row => row.id == value) {
-    Some(row) => Some(row.key)
-    None => None
-  }
+  sessions.iter().find_first(row => row.id == value).map(row => row.key)
 }
 
 ///|
@@ -1763,6 +1760,52 @@ fn session_view_class(model : Model) -> String {
 }
 
 ///|
+/// The shape the three "find X" toolbars share: an optional count chip, a
+/// filter toggle that reads `<label>: on` while active, and prev/next jumps.
+/// `kind` is the CSS prefix the three sets of rules are already keyed on
+/// (`error`, `escalated`, `build-error`). Each caller keeps its own
+/// visibility test, because those genuinely differ.
+fn filter_bar(
+  emit : @cmd.Emit[Msg],
+  count : Int,
+  kind~ : String,
+  active~ : Bool,
+  count_label~ : String,
+  toggle_title~ : String,
+  toggle_label~ : String,
+  toggle~ : Msg,
+  jump~ : (Int) -> Msg,
+  jump_noun~ : String,
+) -> @html.Html {
+  let children : Array[@html.Html] = [
+    ..if count > 0 {
+      [@html.span(class="\{kind}-count") <| count_label]
+    },
+    @html.button(
+      class=class_with("\{kind}-toggle", "active", active),
+      title=toggle_title,
+      on_click=emit(toggle),
+    ) <| (if active { "\{toggle_label}: on" } else { toggle_label }),
+    // Jumps only make sense when there is something to jump to.
+    ..if count > 0 {
+      [
+        @html.button(
+          class="\{kind}-jump",
+          title="Scroll to the previous \{jump_noun}",
+          on_click=emit(jump(-1)),
+        ) <| "↑ prev",
+        @html.button(
+          class="\{kind}-jump",
+          title="Scroll to the next \{jump_noun}",
+          on_click=emit(jump(1)),
+        ) <| "↓ next",
+      ]
+    },
+  ]
+  @html.div(class="\{kind}-bar") <| children
+}
+
+///|
 /// The "find failures" toolbar: a count, an errors-only filter toggle, and
 /// prev/next jumps. Hidden when the view has no failures AND the filter is off,
 /// so a clean session shows no chrome — but if the filter is on it stays visible
@@ -1772,38 +1815,18 @@ fn error_bar(emit : @cmd.Emit[Msg], model : Model, count : Int) -> @html.Html {
   if count == 0 && !model.errors_only {
     return @html.nothing
   }
-  let toggle_class = class_with("error-toggle", "active", model.errors_only)
-  let children : Array[@html.Html] = []
-  if count > 0 {
-    children.push(
-      @html.span(class="error-count") <| "🚩 \{count} tool error\{plural_s(count)}",
-    )
-  }
-  children.push(
-    @html.button(
-      class=toggle_class,
-      title="Show only failed tool results",
-      on_click=emit(ToggleErrorsOnly),
-    ) <| (if model.errors_only { "Errors only: on" } else { "Errors only" }),
+  filter_bar(
+    emit,
+    count,
+    kind="error",
+    active=model.errors_only,
+    count_label="🚩 \{count} tool error\{plural_s(count)}",
+    toggle_title="Show only failed tool results",
+    toggle_label="Errors only",
+    toggle=ToggleErrorsOnly,
+    jump=n => JumpToError(n),
+    jump_noun="failure",
   )
-  // Jumps only make sense when there is something to jump to.
-  if count > 0 {
-    children.push(
-      @html.button(
-        class="error-jump",
-        title="Scroll to the previous failure",
-        on_click=emit(JumpToError(-1)),
-      ) <| "↑ prev",
-    )
-    children.push(
-      @html.button(
-        class="error-jump",
-        title="Scroll to the next failure",
-        on_click=emit(JumpToError(1)),
-      ) <| "↓ next",
-    )
-  }
-  @html.div(class="error-bar") <| children
 }
 
 ///|
@@ -1818,45 +1841,18 @@ fn escalated_bar(
   if count == 0 && !model.escalated_only {
     return @html.nothing
   }
-  let toggle_class = class_with(
-    "escalated-toggle",
-    "active",
-    model.escalated_only,
+  filter_bar(
+    emit,
+    count,
+    kind="escalated",
+    active=model.escalated_only,
+    count_label="🔓 \{count} escalated tool call\{plural_s(count)}",
+    toggle_title="Show only tool calls with escalated: true and their results",
+    toggle_label="Escalated only",
+    toggle=ToggleEscalatedOnly,
+    jump=n => JumpToEscalated(n),
+    jump_noun="escalated tool call",
   )
-  let children : Array[@html.Html] = []
-  if count > 0 {
-    children.push(
-      @html.span(class="escalated-count") <| "🔓 \{count} escalated tool call\{plural_s(count)}",
-    )
-  }
-  children.push(
-    @html.button(
-      class=toggle_class,
-      title="Show only tool calls with escalated: true and their results",
-      on_click=emit(ToggleEscalatedOnly),
-    ) <| (if model.escalated_only {
-      "Escalated only: on"
-    } else {
-      "Escalated only"
-    }),
-  )
-  if count > 0 {
-    children.push(
-      @html.button(
-        class="escalated-jump",
-        title="Scroll to the previous escalated tool call",
-        on_click=emit(JumpToEscalated(-1)),
-      ) <| "↑ prev",
-    )
-    children.push(
-      @html.button(
-        class="escalated-jump",
-        title="Scroll to the next escalated tool call",
-        on_click=emit(JumpToEscalated(1)),
-      ) <| "↓ next",
-    )
-  }
-  @html.div(class="escalated-bar") <| children
 }
 
 ///|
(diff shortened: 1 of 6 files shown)


bobzhang:

**This is the second audit.** The first found the commit message citing a snapshot test that does not exist, leaving the `..if tool.strict` spread's two branches and the wire key order unpinned. Commit `e1e184d1f` pins both as bytes via `stringify()` (a `Map` literal compares equal whatever order it was built in, so the existing structural test could not have caught a reordering) and documents `pad2`'s domain.

**Verdict: signed off.** 1 finding(s), none blocking (1 nit); each is answered below.

> Audited the worktree (HEAD e1e184d1f, two commits over base a227f6aae; worktree itself clean — no uncommitted/untracked changes) against all seven criteria. The diff touches exactly the six listed files in the six listed packages; `git diff --name-only` shows nothing else, and no `.mbti` changed (criterion 2 passes: all rewrites are private or same-signature pub fns — rendered_build_error_count / session_has_mbtx_activity keep signatures).
> 
> Verification performed: ran `moon check` on the full workspace (64 tasks OK) and all six test suites — viz 19/19 (js), viz_export 2/2 (js), deepseek 46/46 (wasm), testkit/filesystem 17/17, inspect 23/23, cmd/viz_app 31/31 (js). Empirically probed the risky rewrites: (1) the two JSON-object patterns (tool_call_is_escalated, tool_call_is_run_failed) reproduce the old Object/get logic on an edge-case battery (absent key, null, wasm/js target, non-objects, extra keys, wrong types) — all identical; (2) is_script_close_tail via to_ascii_lowercase accepts exactly the same char set as the old uppercase-arithmetic test, including non-ASCII lookalikes (É, Kelvin sign, ß, İ all rejected by both); (3) old-vs-new ToolDefinition::to_json (object-literal conditional vs explicit Map + ..if spread) is byte-identical on stringify for both strict and plain; (4) replace_all('<', \"\\u003c\") equals the old StringBuilder loop; (5) the new shared filter_bar emits the exact old markup for error/escalated/build-error bars (classes, titles, labels, child order, count>0 guards, visibility early-returns kept in callers) element-for-element; (6) rows guard-else

Scope: 6 files reviewed; build `moon check OK (64 tasks, all members)`, tests `138 passed / 0 failed (viz 19, viz_export 2, deepseek 46, testkit/filesystem 17, inspect 23, cmd/viz_app 31)`.

### Findings and responses

**[nit] `viz/view.mbt:1648` — pad2 differs from the old helper for negative n, though the only caller can never pass one**

Old code `if n < 10 { "0\{n}" } else { "\{n}" }` treats negatives as n < 10 and yields "0-5" for n=-5; the new `n.to_string().pad_start(2, '0')` yields "-5". The probe run confirmed the divergence for every negative tried and exact equality for 0..=99 and 100+. pad2 is private with a single call site, wall_clock_label (viz/view.mbt:1640), whose arguments are month/day/hour/minute — always non-negative — and the added comment documents the 0..=59 domain, so no reachable behaviour changes. Not a blocker, but the helper's semantics for out-of-domain input did change and the comment should stay true if pad2 is ever reused elsewhere.

*Response:* Left as-is with the domain documented, which is one of the two options the auditor offers. The divergence is unreachable: `pad2` is private with a single caller passing month, day, hour and minute. It is also not a loss — the old form rendered -5 as "0-5", which is not a meaningful two-digit field either, so there is no behaviour worth preserving for an input that cannot occur. The comment added in this PR records the 0..=59 domain so a future caller is warned.

---

bobzhang:
## Held for your review

### Why it is held

**The reviewer's own verdict was `hold`,** for a coverage reason rather than a defect:

`cmd/viz_app/app.mbt:1768` folds three separate toolbars (`error_bar`, `escalated_bar`, `build_error_bar`) into one parameterised `filter_bar` builder. **No test in the repository renders these bars.** Correctness therefore rests entirely on roughly twenty CSS-class, title, label and message literals lining up across the three call sites. The reviewer checked each by hand against the originals and found them correct — but hand-checking twenty literals with no test underneath is exactly the situation where a second pair of eyes is worth more than a third argument.

### Two evaluation-order notes, neither observable today

- `viz/view.mbt:1649` — `pad2` became `n.to_string().pad_start(2, '0')`, which renders −5 as `"-5"` where the old form produced `"0-5"`. Unreachable: `pad2` is file-private with one caller, `wall_clock_label`, which opens with `guard unix_ms > 0` so every derived field is non-negative. This PR adds a comment recording that domain.
- `cmd/viz_app/app.mbt:1823` — the three callers now evaluate `count_label` and allocate the `jump` closure unconditionally, where the old code built the count string only inside `if count > 0`. `plural_s` is total and pure, so at count == 0 the only consequence is one discarded string per bar per render.

The two JSON-pattern rewrites (`tool_call_is_escalated`, `tool_call_is_run_failed`) agree with the old `match Object(fields)` forms on all 13 shapes tried, including missing key, explicit null, non-object and non-string values. `is_script_close_tail`'s `to_ascii_lowercase` rewrite was checked **exhaustively over code points 0..0x10FFF** for each of the seven tail characters. The `Json(Map([.. ..if strict {..}]))` rebuild stringifies byte-identically in both strict states, and this PR adds the test that pins it.

bobzhang:

The held item was the untested `filter_bar` fold. I checked every literal in the three call sites against the three originals: the `error` / `escalated` / `build-error` prefixes on `-count`, `-toggle`, `-jump` and `-bar`; the three count strings with their `plural_s`; the toggle titles and the `: on` suffix; the prev/next titles with each bar's noun; the `↑ prev` / `↓ next` labels; the toggle and jump message constructors; and the count → toggle → prev → next child order with both `count > 0` guards. All match, and each caller keeps its own visibility test verbatim.

`pad2` and the unconditional `count_label` are unobservable at their only callers. The JSON object patterns and `is_script_close_tail` were already verified exhaustively by the earlier pass.

Nothing to change. Merging.

PR #1319, opened 2026-09-06
refactor: project-wide simplification sweep (comprehensions, patterns, stdlib reuse, dedupe)

bobzhang:
A project-wide simplification sweep: list comprehensions, pattern matching
(including nested, array, string and `Json` object patterns), functional loops,
standard-library reuse, and de-duplication of copy-pasted code and test
scaffolding.

**302 files, 6223 insertions, 11672 deletions — 5449 lines net removed.**
No `.mbti` moved, on any package. By AGENTS.md that is the signature of a
refactor with no visible interface change.

## How it was produced

1. **Survey** — 34 read-only agents, one per package group, covering every
   non-generated `.mbt` file in the repository (root module, `desktop/`,
   `editor/`). Each finding had to carry the verbatim current code, a proposed
   replacement, and evidence that the API it leans on exists. 425 findings.
2. **Implement** — 30 agents, each in its own git worktree over a file-disjoint
   bucket, applying its own findings. The surveyors could not compile, so every
   proposal was treated as a proposal: implementers adapted what needed
   adapting and **dropped** anything that would not compile, would move an
   interface, or would change behaviour.
3. **Integrate** — patches applied onto one branch, one commit per bucket, with
   the full gate re-run after each wave.

Every commit is self-contained and revertable on its own.

## What changed, by area

| area | files | net lines |
| --- | ---: | ---: |
| `desktop/` | 105 | −2677 |
| `editor/` | 137 | −2108 |
| `agent_tool/` | 19 | −274 |
| `agent/` | 12 | −154 |
| `cmd/`, `mcp/`, `jsonrpc/`, `eval/` | 15 | −127 |
| the small `agent_*` packages | 8 | −58 |
| `deepseek/`, `viz/`, `inspect/` | 5 | −51 |

Representative changes:

- **`desktop/frontend/update.mbt` alone loses 515 lines**: ten copies of the
  Codex update boilerplate become one helper, four composer arms that re-target
  `update_msg` at the action's owner share one, seven Codex composer-identity
  guards collapse into one or-pattern arm, and repeated test records become
  named fixtures.
- **`desktop/internal/protocol`**: the codec readers and writers that each
  boundary had copied are stated once.
- **`editor/internal/viewer/code_editor_widget`**: twelve open-coded cursor-key
  registrations become six pairs through one helper, and the viewer + task-queue
  fixture repeated verbatim in thirteen tests is shared.
- **Standard library instead of hand-rolled loops**, throughout: `Map::get_or_init`
  for get-or-create, `Iter::find_first` and `Array::any` for break-loops,
  `Array::rev_iter` for backwards scans, `Option::map`/`bind` for `Some/None`
  ladders, `String::replace_all`, `pad_start`, `to_lower` for character loops.
- **Nested and `Json` object patterns** replace `.get(..)`-then-match ladders;
  `guard ... is Some(..)` chains replace three-level `if`/`match` nests.

## Verification

Run on the integrated branch, not just per bucket:

| gate | result |
| --- | --- |
| `moon check --target native --deny-warn` | pass |
| `moon check --target js --deny-warn` | pass |
| `moon check --target wasm --deny-warn` | pass |
| `moon fmt --check` | pass |
| `moon test --target native` | 3203 passed |
| `moon test --target js` | 3185 passed |
| `moon cram test tests/cram` | pass |
| editor: `moon check --target all --warn-list +73 --deny-warn` | pass |
| editor: `moon test --target all` | 1082 wasm / 1863 js / 1216 native, all passed |
| CI's generated-interface check (`moon info`, then diff) | no interface changed |

Test totals against `main`: js, wasm and all three editor targets are
**unchanged**. Native is 3203 against 3205 because one commit folds three
`edit`-tool rejection tests into a single table-driven test; all six assertions
survive and each row carries `msg=` naming its input, so a failing row still
names itself.

## One hunk that is not behaviour-preserving

Found by this PR's own self-review, and kept because it is an improvement:
`mbtx_fixture`'s spill-directory cleanup moved from the end of each of 12 test
bodies to a group `defer`, so it now runs after the task group joins its
children instead of racing a still-draining job, and it also runs on
assertion-failure paths where the old trailing `rmdir` was never reached. That
change of timing is what makes the 12 tails collapsible in the first place.
Everything else in the sweep is behaviour-identical.

## What was deliberately not done

No renames, no file moves, no package splits, no new dependencies, no `pub`
signature changes, and no comment or doc-comment rewording. Findings that would
have required any of those were dropped rather than stretched — including one
that would have deleted a helper still called from two files owned by another
bucket.


bobzhang (commented):
## Self-review

I reviewed my own 30 commits with one independent reviewer per commit — 1082
hunks in total — each asked to find where the "behaviour-preserving" claim is
**false**, and to name a concrete trigger for every finding. Anything a reviewer
raised at blocker or major severity would then have been put to two further
skeptics whose brief was to refute it.

**No reviewer raised a blocker or major finding on any of the 30 commits.**
The 36 findings that came back were all minor or nit. That is a real result but
not a proof: reviewers agreed with the author's own framing on most hunks, and
the sweep is large enough that a maintainer's eye on the two commits below is
still worth more than any of this.

### Fixed in 204c620f7

One narrowed contract, three misleading names, and seven comments that no longer
matched their code:

- **`spaces(n)` stopped being total over `Int`** (`editor/viewer/common/view_model`).
  `" ".repeat(n)` aborts on a negative count where the old write-char loop
  returned `""`. Both in-tree callers are guarded so nothing can reach it today,
  but the helper is package-private and the next caller should not have to know.
  Now `" ".repeat(n.max(0))`.
- **`escape_into` allocates and returns**, so the `_into` sink convention was
  backwards. Renamed `escape_impl`.
- **`required_error_pair` returns `None` for an absent pair** while every other
  `required_*` reader in that package raises. Renamed `optional_error_pair`.
- **`posix_quoted` held raw directories**, not quoted ones. Renamed `posix_dirs`.
- **Two branch comments were stacked verbatim over one now-unconditional call**
  in `fileeditor/update.mbt`, and over one shared `settle_unreadable` arm, which
  left a dangling "Like a deletion" whose antecedent had been folded away. Both
  rewritten as single statements of the shared invariant.
- Stale comments in `agent_skill`, `desktop/internal/auth`,
  `view_layout/scrollable.mbt`, a markers test, and the `jsonrpc` fixture doc.

### Correction to this PR's description

The description called every hunk behaviour-preserving. One test-only hunk is
not, and is better for it: `mbtx_fixture`'s spill-directory cleanup moved from
the end of each of 12 test bodies to a group `defer`. It now runs after the
task group has joined its children rather than racing a still-draining job, and
it runs on assertion-failure paths where the old trailing `rmdir` was simply
never reached. That change of timing is what makes the 12 tails collapsible.

### Left as-is, deliberately

- **Two test folds fail fast** — `edit`'s three malformed-argument cases
  (`agent_tool`) and four per-severity squiggly rules (`editor/view_layout`).
  A regression in row 1 stops later rows being exercised. Every row carries
  `msg=` naming its input, which is the trade-off this project accepted for
  reject tables in the last sweep; assertion counts are unchanged.
- **Shared guard messages lose per-row source locations** in
  `manifest_error_test` and `shellenv_test`: six cases now report at one line.
  The `msg=` argument still names the failing input.
- **Three or-patterns merge arms that were not identical** and re-discriminate
  inside (`desktop/frontend/codex/state.mbt`, `editor/.../quick_diff_decorator.mbt`,
  `editor/.../main.mbt`). Behaviour is identical; whether one arm plus a re-test
  reads better than two arms is a judgement call I would not fight for.
- **A few allocations that did not exist before**: `drain` copies a prefix where
  a shift allocated nothing, `stamp_range` materialises stamps, and a context-menu
  boundary index is built per keypress. All off hot paths as far as I can tell,
  none measured.

### What the review does not cover

Reviewers read the diff and the surrounding code; they did not run the desktop
or editor browser suites, and neither did I. The `Desktop` and `Editor`
Playwright jobs on this PR are the first real exercise of the frontend changes,
which are 105 of the 302 files. If anything in this sweep breaks, my prior is
that it is there rather than in the compiled-and-tested core.

bobzhang:
### CI status

Every stable gate is green: native, JS and wasm checks, the generated-interface
check, the cram suite, `desktop e2e`, `viz e2e`, `browser-correctness`, and the
editor's stable jobs on both Linux and macOS.

Two notes on what is not green:

**`check (stable, native)` failed on its first run and passes on re-run.** The
failure was `agent_tool/shell/shell_test.mbt:357 "shell honors cwd when
non-empty"` — a Linux-runner flake, not a regression. That test is byte-identical
before and after this PR, and the shell hunks near it are extractions that keep
the code they replaced verbatim (`realpath_or_same` is the same
`match cwd { Some(cwd) => Some(@fs.realpath(cwd) catch { _ => cwd }); None => None }`,
and `first_protected_workspace_source` keeps the per-target direct-then-symlink
order of the loop it replaced). The package's 137 tests also pass 8 out of 8
consecutive local runs.

**`moon (nightly, …)` fails on both runners, and fails the same way on `main`.**
`Cannot find import 'moonbitlang/core/strconv' in moonbit-community/rabbita/url@0.15.4`
— nightly core dropped `core/strconv` and the pinned rabbita still imports it.
Both jobs are `continue-on-error: true`, so they do not gate this PR, and
nothing here touches dependency pins.

bobzhang:
Superseded by seven smaller pull requests. Same work, rebased onto current
main, split so the safe majority can be merged without a careful read.

Every file appears in exactly one of them, and each was verified on its own
branch with no other part of the sweep applied, so they can be reviewed and
merged in any order.

**Merge on sight** — every hunk is a local rewrite in place. No helper added or
removed, no code moved between functions, no signature change, no test change.

| PR | scope | files | net lines |
| --- | --- | ---: | ---: |
| #1322 | root module | 29 | −168 |
| #1323 | `desktop/` | 51 | −436 |
| #1324 | `editor/` | 89 | −976 |

**Needs a real read** — these add shared helpers, route call sites through
them, or reshape control flow, so correctness depends on code outside the hunk.

| PR | scope | files | net lines |
| --- | --- | ---: | ---: |
| #1325 | root module | 11 | −238 |
| #1326 | `desktop/` | 43 | −2082 |
| #1327 | `editor/` | 39 | −962 |

**Its own decision** — test fixtures and folds, no production code:

| PR | scope | files | net lines |
| --- | --- | ---: | ---: |
| #1328 | all modules | 39 | −583 |

The self-review posted on this PR still applies to the same code. Its findings
were addressed before the split, so they are already folded into the seven
branches rather than sitting as follow-up work.

Two conflicts came up while rebasing onto current main, and main won both: the
subtask tool's entry point and `resume_running` were deleted upstream, and this
sweep had only simplified them.

PR #1322, opened 2026-09-06
refactor: local stdlib and pattern rewrites in the root module

bobzhang:

| before | after |
| --- | --- |

Scope: the root module — `agent*`, `cmd/`, `mcp/`, `eval/`, `viz*`, `deepseek/`, `inspect/`.

--target js --deny-warn`, and the full native and JS test suites — with no
other part of the sweep applied. No `.mbti` changed, so no package interface
moves.

---

of the sweep applied.

| | PR | what it is |
| --- | --- | --- |

### DIFF
--- a/agent_session/store/store.mbt
+++ b/agent_session/store/store.mbt
@@ -75,20 +75,16 @@ pub fn SessionStore::SessionStore(root : String) -> SessionStore {
 
 ///|
 fn validate_session_id(id : @agent_session.SessionId) -> Unit raise {
-  let value = id.value()
-  if value == "" {
-    fail("session id must not be empty")
-  }
-  if value.contains("/") || value.contains("\\") {
-    fail("session id must not contain path separators: \{value}")
-  }
-  if value.contains("\u{0}") {
+  match id.value() {
+    "" => fail("session id must not be empty")
+    value if value.contains("/") || value.contains("\\") =>
+      fail("session id must not contain path separators: \{value}")
     // A NUL truncates the path at the C boundary, silently mapping distinct
     // ids onto one directory.
-    fail("session id must not contain a NUL character")
-  }
-  if value == "." || value == ".." {
-    fail("session id must not be `.` or `..`")
+    value if value.contains("\u{0}") =>
+      fail("session id must not contain a NUL character")
+    "." | ".." => fail("session id must not be `.` or `..`")
+    _ => ()
   }
 }
 
@@ -645,11 +641,7 @@ async fn first_prompt_in_log(path : String) -> String {
 fn[T] stamped_newest_first(a : (T, Int64, Int), b : (T, Int64, Int)) -> Int {
   let (_, a_seconds, a_nanoseconds) = a
   let (_, b_seconds, b_nanoseconds) = b
-  if b_seconds != a_seconds {
-    b_seconds.compare(a_seconds)
-  } else {
-    b_nanoseconds.compare(a_nanoseconds)
-  }
+  (b_seconds, b_nanoseconds).compare((a_seconds, a_nanoseconds))
 }
 
 ///|
--- a/agent_session/store/store.mbt
+++ b/agent_session/store/store.mbt
@@ -641,7 +641,14 @@ async fn first_prompt_in_log(path : String) -> String {
 fn[T] stamped_newest_first(a : (T, Int64, Int), b : (T, Int64, Int)) -> Int {
   let (_, a_seconds, a_nanoseconds) = a
   let (_, b_seconds, b_nanoseconds) = b
-  (b_seconds, b_nanoseconds).compare((a_seconds, a_nanoseconds))
+  // Compared field by field rather than as a tuple: this is a `sort_by`
+  // comparator, so it runs O(n log n) times per listing and must not build a
+  // pair on every call.
+  if b_seconds != a_seconds {
+    b_seconds.compare(a_seconds)
+  } else {
+    b_nanoseconds.compare(a_nanoseconds)
+  }
 }
 
 ///|
--- a/agent_subtask/geometry.mbt
+++ b/agent_subtask/geometry.mbt
@@ -50,11 +50,8 @@ pub async fn resolve_geometry(
   }
   let worktree_roots : Array[String] = []
   for record in listing.output.split("\u{0}") {
-    let record = record.to_owned()
-    if record.has_prefix("worktree ") {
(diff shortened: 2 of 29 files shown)


agent_session/store/store.mbt
```diff
@@ -645,11 +641,7 @@ async fn first_prompt_in_log(path : String) -> String {
 fn[T] stamped_newest_first(a : (T, Int64, Int), b : (T, Int64, Int)) -> Int {
   let (_, a_seconds, a_nanoseconds) = a
   let (_, b_seconds, b_nanoseconds) = b
-  if b_seconds != a_seconds {
-    b_seconds.compare(a_seconds)
-  } else {
-    b_nanoseconds.compare(a_nanoseconds)
-  }
+  (b_seconds, b_nanoseconds).compare((a_seconds, a_nanoseconds))
```
bobzhang: slow

bobzhang: Fixed in 3e6f1a58f — reverted to comparing the two fields directly.

You are right that this one should not have changed. It is the comparator for
two `sort_by` calls, so it runs O(n log n) times per listing, and
`(b_secs, b_nanos).compare((a_secs, a_nanos))` builds a pair on every one of
those calls where the branch-and-compare allocated nothing.

Your comment also prompted a perf pass over all three mechanical PRs, hunk by
hunk, looking only for regressions. It found one more here and two in the
editor PR, all now reverted:

- `mcp/streamhttp/transport.mbt` (this PR): the SSE reader's buffer compaction
  had become `Array::drain`, which allocates a fresh array for the >= 4 KB it
  removes and immediately discards it, once per 4 KB of stream. Back to the
  in-place shift.
- `editor/.../view_lines_viewport_data.mbt` (#1324): `unwrap_or([])` built the
  empty fallback array on every call, and that runs once per rendered line.
- `editor/.../inline_decorations.mbt` (#1324): `get_or_init` built its
  capturing closure before the cache lookup, so every hit paid for a closure it
  never called, once per decoration per render.

Nothing else in the three PRs allocates on a path the review could show is hot.

agent_subtask/geometry.mbt
```diff
@@ -50,11 +50,8 @@ pub async fn resolve_geometry(
   }
   let worktree_roots : Array[String] = []
   for record in listing.output.split("\u{0}") {
-    let record = record.to_owned()
-    if record.has_prefix("worktree ") {
-      worktree_roots.push(
-        canonical(record.view(start_offset="worktree ".length()).to_owned()),
```
bobzhang, line 56: canonical take StringView to avoid the allocation?

agent_subtask/registry.mbt
```diff
@@ -147,13 +147,7 @@ fn components(prefix : String) -> Array[String] {
 /// Whether `shorter` is a component-wise prefix of `longer` (equal counts as
 /// covered).
 fn one_covers(shorter : Array[String], longer : Array[String]) -> Bool {
```
bobzhang, line 149: inline it?

agent_tool/edit/edit.mbt
```diff
@@ -624,11 +624,7 @@ fn total_line_count(content : String) -> Int {
   if content == "" {
     return 0
   }
-  let count = for ch in content; count = 0 {
-    continue if ch == '\n' { count + 1 } else { count }
-  } nobreak {
-    count
-  }
+  let count = content.iter().count_if(ch => ch == '\n')
```
bobzhang, line 627: count.count_if?

agent_tool/read/html/read_output.mbt
```diff
@@ -78,11 +77,8 @@ pub fn moonbit_output_for_brief(
   content : String,
 ) -> @rhtml.Html? {
   guard brief.strip_prefix("read ") is Some(path_view) else { return None }
-  let path = match path_view.strip_suffix(" (truncated)") {
-    Some(path) => path.to_owned()
-    None => path_view.to_owned()
-  }
-  moonbit_output_for_path(path, content)
+  let path = path_view.strip_suffix(" (truncated)").unwrap_or(path_view)
+  moonbit_output_for_path(path.to_owned(), content)
```
bobzhang, line 81: take StringView to avoid allocation?

agent_tool/shell/internal/decode/decode.mbt
```diff
@@ -95,11 +95,7 @@ fn optional_positive_int(
 
 ///|
 fn cap_max_output_chars(value : Int) -> Int {
```
bobzhang, line 97: remove the tiny helper?

agent_subtask/geometry.mbt
bobzhang: Applied in #1329. `@fs.realpath` already takes a `StringView`, so `canonical`
now does too and the `to_owned()` at the worktree-list site is gone — it was
copying a path only to hand it straight through. The fallback arm still owns
its result, but that is the error path rather than the scan.

agent_subtask/registry.mbt
bobzhang: Applied in #1329, with one correction to the premise: `one_covers` has two call
sites, not one — the overlap test here and the admission test in `scope.mbt`.
Inlining still wins, because at each site `longer.starts_with(shorter)` says
what it means, where `one_covers(prefix, parts)` made the reader resolve an
argument order first.

agent_tool/edit/edit.mbt
bobzhang: Looked into this one and left it alone — see #1329 for the reasoning.

`String` has no `count_if`, so the `.iter()` hop is what reaches
`Iter::count_if`; the only other counting method on `Iter` is `count`, which
takes no predicate. The intermediate binding also has to stay, since the result
is read twice, once per branch of the trailing `has_suffix` test. If you meant
something else by it, say so and I will change it.

agent_tool/read/html/read_output.mbt
bobzhang: Applied in #1329. `moonbit_output_for_path` and `is_moonbit_source` take a
`StringView`, so rendering from a read brief no longer copies the path out of
its `strip_prefix` view just to check the extension.

agent_tool/shell/internal/decode/decode.mbt
bobzhang: Applied in #1329 — inlined as `max_output_chars.min(hard_max_output_chars)` at
its single call site, and the helper is gone.

PR #1323, opened 2026-09-06
refactor(desktop): local stdlib and pattern rewrites

bobzhang:

| before | after |
| --- | --- |

Scope: `desktop/` — the frontend components and the internal host, engine and protocol packages.

--target js --deny-warn`, and the full native and JS test suites — with no
other part of the sweep applied. No `.mbti` changed, so no package interface
moves.

---

of the sweep applied.

| | PR | what it is |
| --- | --- | --- |

### DIFF
--- a/desktop/frontend/agent_model.mbt
+++ b/desktop/frontend/agent_model.mbt
@@ -68,12 +68,11 @@ fn ModelChoice::parse(value : String) -> ModelChoice? {
   if value == CodexUnavailableChoiceWire {
     return Some(CodexUnavailablePick)
   }
-  if value.has_prefix(OpenSeekChoicePrefix) {
-    let rest = value.view(start_offset=OpenSeekChoicePrefix.length()).to_owned()
-    return EngineModel::parse(rest).map(model => OpenSeekModel(model))
+  if value.strip_prefix(OpenSeekChoicePrefix) is Some(rest) {
+    return EngineModel::parse(rest.to_owned()).map(model => OpenSeekModel(model))
   }
-  if value.has_prefix(CodexChoicePrefix) {
-    let id = value.view(start_offset=CodexChoicePrefix.length()).to_owned()
+  if value.strip_prefix(CodexChoicePrefix) is Some(rest) {
+    let id = rest.to_owned()
     if !id.trim().is_empty() {
       return Some(CodexModel(id))
     }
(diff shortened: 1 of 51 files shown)


desktop/frontend/agent_model.mbt
```diff
@@ -68,12 +68,11 @@ fn ModelChoice::parse(value : String) -> ModelChoice? {
   if value == CodexUnavailableChoiceWire {
     return Some(CodexUnavailablePick)
   }
-  if value.has_prefix(OpenSeekChoicePrefix) {
-    let rest = value.view(start_offset=OpenSeekChoicePrefix.length()).to_owned()
-    return EngineModel::parse(rest).map(model => OpenSeekModel(model))
+  if value.strip_prefix(OpenSeekChoicePrefix) is Some(rest) {
+    return EngineModel::parse(rest.to_owned()).map(model => OpenSeekModel(model))
```
bobzhang, line 72: take StringView to avoid allocation?

desktop/frontend/console.mbt
```diff
@@ -235,16 +235,12 @@ fn apply_roster(
...
-      if meta.id == id && meta.online {
-        let channel : @interop.ChannelId = Remote(id)
-        if !devices.iter().any(dev => dev.channel == channel) {
-          websocket_epoch += 1
-          devices.push({ ..empty_device_state(channel), websocket_epoch, })
-        }
-        break
-      }
+  if console.page_device is Some(id) &&
+    metas.iter().any(meta => meta.id == id && meta.online) {
```
bobzhang, line 239: metas.any(...)

desktop/frontend/agent_model.mbt
bobzhang: Applied in #1334. `EngineModel::parse` takes a `StringView` now — its body
already trimmed and matched against literals, and only the unknown-model arm
needs an owned name, so the copy happens once for a custom endpoint instead of
on every parse. `ModelChoice::parse` hands its `strip_prefix` view straight
through, which removes the `to_owned()` you pointed at.

desktop/frontend/console.mbt
bobzhang: Applied in #1334, and to all six sites in this function rather than just this
one: `Array` has `any` and `all`, so the `iter()` hop was building an iterator
per call for nothing. Four of the six predate the sweep. The same fix applies in
`codex/model.mbt`, where `has_user_item` was building two iterators per turn.

Two `.iter().any(...)` sites elsewhere in desktop are over a `String` and stay
as they are — `String` has no `any`, so the hop to `Iter` is what makes them
compile.

PR #1324, opened 2026-09-06
refactor(editor): local stdlib and pattern rewrites

bobzhang:

| before | after |
| --- | --- |

Scope: `editor/` — the viewer, its contributions, the common model and view layers, and the shell.

`moon fmt --check`, the native and JS workspace checks with `--deny-warn`, the
full native and JS test suites, and from `editor/` both `moon check --target
all --warn-list +73 --deny-warn` and `moon test --target all` — with no other
part of the sweep applied. No `.mbti` changed, so no package interface moves.

---

of the sweep applied.

| | PR | what it is |
| --- | --- | --- |


bobzhang:

```
```

PR #1325, opened 2026-09-06
refactor: share repeated helpers in the root module

bobzhang:

What to check, hardest first:

   appears to.

Scope: the root module. The shared pieces are `any_arg_before_ddash` for the git-policy option scans, `realpath_or_same` and `first_protected_workspace_source` in the shell sandbox, the viz app's filter and card helpers, and `array_field` across the eval harness.

### Review comments addressed

- **"why such change in a refactoring?"** on `mbtx_fixture` — reverted. The
  fixture no longer registers a group defer for the spill directory, and all
  twelve tests reclaim it at the end of their own body as before. That hunk
  moved cleanup from end-of-test to end-of-group, which runs it later and also
  on assertion-failure paths where the old tail was never reached. It may be an
  improvement, but it is a behaviour change and does not belong in a refactor.
  `wait_until` and the table-driven registry test, which change nothing
  observable, stay.
- **`for arg in words[start:end]`?** — applied. `any_arg_before_ddash` iterates
  the slice directly instead of indexing.

---

| --- | --- | --- |
| `editor/` | #1332 | #1327 |
| tests | #1333 | #1328 |

### DIFF
--- a/agent/tool_definition.mbt
+++ b/agent/tool_definition.mbt
@@ -5,14 +5,11 @@ fn background_notice_text(snapshot : @bgjobs.BgJobSnapshot) -> String {
   let status = match snapshot.status {
     Exited(0) => "exit=0"
     Exited(code) => "exit=\{code}"
-    Stopped =>
-      if snapshot.killed_by_output_limit {
-        "killed: flooded past the output cap"
-      } else if snapshot.killed_by_time_limit {
-        "killed: outlived the wall-clock cap"
-      } else {
-        "stopped"
-      }
+    Stopped if snapshot.killed_by_output_limit =>
+      "killed: flooded past the output cap"
+    Stopped if snapshot.killed_by_time_limit =>
+      "killed: outlived the wall-clock cap"
+    Stopped => "stopped"
     Running => "running"
   }
   "background job \{snapshot.id} finished (\{status}): `\{snapshot.command}` — read its output with job_output (job_id=\"\{snapshot.id}\")."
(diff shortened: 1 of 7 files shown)


agent/tool_definition.mbt
```diff
@@ -143,6 +140,15 @@ async fn[X] mbtx_fixture(
   approval? : @agent_runtime.ApprovalChannel,
 ) -> (String, @bgjobs.BgJobRuntime, @agent_tool.AgentToolDefinition) {
   let spill_dir = @fs.tmpdir(prefix~)
+  // Reclaimed when the group ends — after every job it spawned has
+  // terminated — the same group-defer shape `build_tools` uses above.
+  group.add_defer(() => {
```
bobzhang: why such change in a refactoring?

agent_tool/shell/internal/git_policy/git_policy.mbt
```diff
@@ -306,6 +306,28 @@ fn short_cluster_has(arg : String, flag : String) -> Bool {
...
+/// True when some argument in `words[start..end)` before a `--` terminator
+/// satisfies `matches`. Every option scan here stops at `--`: past it every
+/// word is an operand, and an operand cannot be an option.
+fn any_arg_before_ddash(
+  words : Array[String],
+  start : Int,
+  end : Int,
+  matches : (String) -> Bool,
+) -> Bool {
+  for index in start..<end {
```
bobzhang: `for arg in words[start:end]`?

agent/tool_definition.mbt
bobzhang: Fair, and reverted. `mbtx_fixture` no longer registers a group defer for the
spill directory, and all twelve tests reclaim it at the end of their own body
again, exactly as before.

You are right that it does not belong in a refactor: moving cleanup from
end-of-test to end-of-group runs it later, and also runs it on
assertion-failure paths where the old trailing `rmdir` was never reached. My
own review of this sweep had flagged it as the single hunk that was not
behaviour-preserving, so your question is the right verdict on it. If the
better lifetime is wanted, it should be its own change with that as the stated
point.

The `wait_until` poller and the table-driven registry test stay: neither
changes anything observable.

agent_tool/shell/internal/git_policy/git_policy.mbt
bobzhang: Applied. `any_arg_before_ddash` iterates `words[start:end]` directly, so the
index binding is gone.

PR #1326, opened 2026-09-06
refactor(desktop): share the repeated codec, engine and frontend helpers

bobzhang:

What to check, hardest first:

   appears to.

Scope: `desktop/`. The largest pieces are the protocol codec readers and writers each boundary had copied, the engine's shared refusal and run-actor fixtures, and the frontend's repeated update transitions.

### Review comment addressed

**`[for key, _ in ... if !desired_keys.contains => key]`** — applied. The stale
editor keys are filtered in the comprehension, so the intermediate array of
every editor key is gone. The predicate is named rather than inline because
`moon fmt` wraps the longer form, and a wrapped comprehension whose body is a
bare identifier becomes `{ key }`, which the compiler reads as a struct literal.

---

| --- | --- | --- |
| `editor/` | #1332 | #1327 |
| tests | #1333 | #1328 |

### DIFF
--- a/desktop/frontend/fileeditor/semantic_review_view.mbt
+++ b/desktop/frontend/fileeditor/semantic_review_view.mbt
@@ -1376,14 +1376,20 @@ fn refresh_semantic_review_editor_options() -> Unit {
 }
 
 ///|
+/// Retire one mounted section widget and drop its map entry.
+fn remove_semantic_review_editor(key : String) -> Unit {
+  if semantic_review_editors.get(key) is Some(entry) {
+    dispose_semantic_review_editor(entry)
+    semantic_review_editors.remove(key)
+  }
+}
+
+///|
 fn clear_semantic_review_editors() -> Unit {
   semantic_review_navigation_generation.val += 1
-  let keys = semantic_review_editors.to_array().map(pair => pair.0)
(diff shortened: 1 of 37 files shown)


desktop/frontend/fileeditor/semantic_review_view.mbt
```diff
@@ -1499,51 +1505,35 @@ fn sync_semantic_review_editors(state : State) -> @cmd.Cmd {
         !desired_keys.contains(key) {
         semantic_review_active_key.val = None
       }
-      let stale = semantic_review_editors
-        .to_array()
-        .map(pair => pair.0)
-        .filter(key => !desired_keys.contains(key))
+      let stale = [
```
bobzhang, line 1509: [ for key,_ in ... if !desired_keys.containts => key]

bobzhang: Applied. The stale keys are filtered in the comprehension, so the intermediate
array of every editor key is gone.

One wrinkle worth recording: written inline the line is long enough that
`moon fmt` wraps it, and a wrapped comprehension whose body is a bare
identifier gets reformatted to `{ key }`, which the compiler then reads as a
struct literal (E4033, "There is no record definition with the fields: key").
Naming the predicate keeps it on one line and avoids that.

PR #1327, opened 2026-09-06
refactor(editor): share the repeated viewer and contribution helpers

bobzhang:

What to check, hardest first:

   appears to.

Scope: `editor/`. The largest pieces are the cursor-key registration pairs, the shared listener disposal sweep, the repeated Option ladders in the code editor widget, and the shared decode helpers in the remote protocol.

applied. No `.mbti` changed.

---

| --- | --- | --- |
| `editor/` | #1332 | #1327 |
| tests | #1333 | #1328 |


bobzhang:

```
```

PR #1328, opened 2026-09-06
test: share fixtures and fold repeated cases

bobzhang:
**Test-only, and shared.** Fixtures introduced or shared across tests, and a few
near-identical cases folded into tables. The in-place test rewrites this was
mixed with have moved to #1333.

The fold is the thing to weigh: a folded test stops at its first failing row,
where separate tests each reported independently. Assertion counts are preserved
and every row carries `msg=` naming its input. Two are worth a second look on
that basis — the `edit` tool's malformed-argument cases, and the four
per-severity squiggly rules.

No production file changes here.

Scope: test files across all three modules.

applied. No `.mbti` changed.

---

| --- | --- | --- |
| `editor/` | #1332 | #1327 |
| tests | #1333 | #1328 |

### DIFF
--- a/agent/auto_compact_test.mbt
+++ b/agent/auto_compact_test.mbt
@@ -13,36 +13,29 @@ async fn auto_compact_test_server(
   group : @async.TaskGroup[Unit],
   handle : (Json) -> MockReply raise,
 ) -> String? {
-  guard bind_test_server(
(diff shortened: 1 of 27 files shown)


bobzhang:

```
```

PR #1330, opened 2026-09-06
refactor: in-function rewrites in the root module

bobzhang:
**Every hunk stays inside one function.** Nothing shared is added, removed or
re-signatured, so each hunk can be judged by reading it — there is no other code
to go check.

The shapes used, all of them:

| before | after |
| --- | --- |
| an `if`/`else` ladder | one `match` with guards or nested patterns |
| a nested `match` on options | a `guard ... is Some(..)` chain |
| a manual index loop | a comprehension, `fold`, `any`, `find_first`, `search_by` |
| two identical `match` arms | one or-pattern arm |
| `match o { Some(v) => ...; None => ... }` | an Option combinator |
| duplicated branches in one function | a local binding or a closure used only there |

The helper extractions these were mixed with are in the paired pull request, so
nothing here crosses a call site.

Scope: `agent/turn_loop.mbt`, `agent_tool/write`, and two `eval/` harness files.

---

| --- | --- | --- |
| root | `simplify/local-root` | `simplify/structural-root` (#1325) |
| `desktop/` | `simplify/local-desktop` | `simplify/structural-desktop` (#1326) |
| `editor/` | `simplify/local-editor` | `simplify/structural-editor` (#1327) |
| tests | `simplify/tests-local` | `simplify/tests` (#1328) |

### DIFF
--- a/agent_tool/write/write.mbt
+++ b/agent_tool/write/write.mbt
@@ -123,13 +123,13 @@ async fn write_file(
   file_state~ : @agent_tool.FileStateMap,
   write_scope? : @write_scope.WriteScope,
 ) -> @agent_tool.ToolAction {
-  // Every failure names the file it tried to write, as `edit`/`remove` do.
-  let fail_brief = "write \{@agent_tool.brief_basename(path)}"
   fn err(message : String) -> @agent_tool.ToolAction {
+    // Every failure names the file it tried to write, as `edit`/`remove` do.
+    // Built here rather than once per call: the success path never needs it.
     @agent_tool.respond(
       "error writing \{path}: \{message}",
       is_error=true,
-      brief=fail_brief,
+      brief="write \{@agent_tool.brief_basename(path)}",
     )
   }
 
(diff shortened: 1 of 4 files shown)


agent_tool/write/write.mbt
```diff
@@ -123,13 +123,19 @@ async fn write_file(
   file_state~ : @agent_tool.FileStateMap,
   write_scope? : @write_scope.WriteScope,
 ) -> @agent_tool.ToolAction {
-  if @manifest_error.write_error(path, input.content) is Some(message) {
-    return @agent_tool.respond(
+  // Every failure names the file it tried to write, as `edit`/`remove` do.
+  let fail_brief = "write \{@agent_tool.brief_basename(path)}"
```
bobzhang: `let fail_breif` should  be inside err to be 100% consistent with the old code

bobzhang: Right — fixed in 993dbd625, pushed to this branch.

Hoisting it changed when the work happens, not just where it reads. Each of the
four error sites used to build `"write \{brief_basename(path)}"` itself, so it
was built only when that error fired. As a binding at the top of `write_file`
it was built on every call, including the success path that never reads it. It
is back inside `err`, so a successful write does no work for a brief it will
not use.

PR #1333, opened 2026-09-06
test: in-place test rewrites

bobzhang:
**Test-only, and local to one test each.** No fixture is introduced or shared
between tests, no test block is folded into a table, and no assertion moves or
disappears.

No production file changes, so if this is green the suites still cover exactly
what they covered. The shared fixtures and the table folds — the changes with a
real trade-off — are in the paired pull request.

Scope: test files across all three modules, including the four syntax lexer suites.

applied. No `.mbti` changed.

---

| --- | --- | --- |
| root | `simplify/local-root` | `simplify/structural-root` (#1325) |
| `desktop/` | `simplify/local-desktop` | `simplify/structural-desktop` (#1326) |
| `editor/` | `simplify/local-editor` | `simplify/structural-editor` (#1327) |
| tests | `simplify/tests-local` | `simplify/tests` (#1328) |

### DIFF
--- a/agent/concurrent_tools_test.mbt
+++ b/agent/concurrent_tools_test.mbt
@@ -41,13 +41,10 @@ fn probe_tool(
 ///|
 /// The recorded tool results for the batch, in session order.
 fn probe_results(session : @agent_session.Session) -> Array[String] {
-  let out : Array[String] = []
-  for event in session.events() {
-    if event.item() is Tool(result) && result.tool_name() == "probe" {
-      out.push(result.content())
-    }
-  }
-  out
+  [
+    for event in session.events() if event.item() is Tool(result) &&
+    result.tool_name() == "probe" => result.content()
+  ]
 }
 
 ///|
--- a/agent/goal_gate_test.mbt
+++ b/agent/goal_gate_test.mbt
@@ -29,14 +29,13 @@ fn gate_final_payload(text : String) -> String {
 
 ///|
 fn review_notice_count(session : @agent_session.Session) -> Int {
-  let mut count = 0
-  for event in session.events() {
-    if event.item() is Runtime(notice) &&
-      notice.content().has_prefix("[review]") {
-      count += 1
-    }
-  }
-  count
+  session
+  .events()
+  .iter()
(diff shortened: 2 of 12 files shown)


bobzhang:
The nightly failure here is fixed by #1335, kept separate rather than folded in.

The cause is `moonbit-community/rabbita/url@0.15.4` importing
`moonbitlang/core/strconv`, which nightly core removed. It fails the same way on
`main` and on every open PR, and nothing in this branch touches it.

Bumping to rabbita 0.15.6 is an API migration rather than a version line: DOM
coordinate getters return `Double` instead of `Int`, `apply`/`new` take a
`FixedArray` argument list, and `HTMLCollection`/`NodeList` lost indexing in
favour of `item(index)`. It comes to 40 files across `desktop/` and `editor/`,
including both fake DOMs behind the editor's whitebox tests, which returned bare
arrays and now have to carry an `item` method.

Folding that into this PR would have cost it the property that makes it quick to
review — 12 files, test-only, every hunk local to one test. Merging #1335 turns
the nightly check green here and on every other open PR at once.

agent/concurrent_tools_test.mbt
```diff
@@ -41,13 +41,10 @@ fn probe_tool(
 ///|
 /// The recorded tool results for the batch, in session order.
```
bobzhang, line 42: test

agent/concurrent_tools_test.mbt
```diff
@@ -41,13 +41,10 @@ fn probe_tool(
...
 fn probe_results(session : @agent_session.Session) -> Array[String] {
```
bobzhang, line 43: again

agent/concurrent_tools_test.mbt
```diff
@@ -41,13 +41,10 @@ fn probe_tool(
...
-  let out : Array[String] = []
-  for event in session.events() {
-    if event.item() is Tool(result) && result.tool_name() == "probe" {
-      out.push(result.content())
-    }
-  }
-  out
+  [
```
bobzhang, line 44: again

PR #1335, opened 2026-09-06
deps: upgrade rabbita to 0.15.7

bobzhang:
Upgrade Rabbita from 0.15.4 to 0.15.7 in the root, Desktop, editor, and viz modules. This removes the old `core/strconv` dependency that broke the editor check with nightly MoonBit.

Adapt callers to the current API: DOM coordinates return `Double`, JavaScript calls take `FixedArray` arguments, and DOM collections expose `item()` access. Convert coordinates to `Int` only at existing integer boundaries, and update the existing editor DOM fixtures to implement the collection contract.

Rebased onto `main` at `10b1f5b27`, preserving main's removal of the repeated titlebar `refreshDom` scan. No public interface changes.

Validation on macOS with MoonBit 0.1.20260904:

- `just check`, `just build`, `moon info`, and `moon fmt`.
- `just test`: 3248 native and 3217 JS tests; 24 cram cases; CLI turn-finish integration test.
- `just editor-test`: 1090 Wasm, 1875 JS, and 1224 native tests.
- Editor all-target check with `--warn-list +73 --deny-warn`.
- Browser suites: editor 107, Desktop 58, and viz 13 tests.
- `git diff --check` and `git range-diff`; one commit above main, no merge commits.

### DIFF
--- a/cmd/viz_app/moon.mod
+++ b/cmd/viz_app/moon.mod
@@ -4,7 +4,7 @@ version = "0.1.0"
 
 import {
   "bobzhang/openseek@0.2.2",
-  "moonbit-community/rabbita@0.15.4",
+  "moonbit-community/rabbita@0.15.7",
 }
 
 preferred_target = "js"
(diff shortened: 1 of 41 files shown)


bobzhang:
cc @Milky2018 diago should be updated

PR #1349, opened 2026-09-07
feat(desktop): syntax-highlight moonbit fences in the transcript body

tonyfettes:
## What

Assistant/user transcript **body** markdown fences of MoonBit code were
rendered as plain monochrome text: `markdown.mbt`'s `code_block` emitted
the raw escaped source inside `<code class="language-moonbit">` with no
token spans, unlike `mbtx` tool-call arguments which already use the
editor's `mtk<n>` token pipeline (`moonbit_source_code` /
`moonbit_numbered_source`) inside `pre.moonbit-source`.

This PR routes `moonbit`/`mbt`/`mbtx` fences in the transcript markdown
body through `@syntax_html.moonbit_source_code` and wraps them in
`pre.moonbit-source`, so they get the same syntax colors as an `mbtx`
program card or a composer approval body. The `moonbit-source` class is
what scopes the `--vscode-token-*` palette (dark and light) to the
block, so no CSS change was needed. Fences of other languages keep
their previous plain escaped rendering. Applies everywhere the body
markdown renderer is used (assistant messages, reasoning, answers,
summaries, skills markdown).

## Verification

- `moon check --target js desktop/frontend/markdown` — clean
- `moon test --target js desktop/frontend/markdown` — 13/13 passed
- `moon check --target js` on both consumers
  (`desktop/frontend/transcript/component`, `desktop/frontend/skills`) — clean
- `moon fmt --check desktop/frontend/markdown` — clean

Not verified: a visual run of the desktop app (no screenshot taken
here). The token palette scope and CSS cascade were confirmed
statically; a quick eyeball of a ```moonbit fence in a live transcript
in both themes is the remaining check.

Generated with SeekMoon


tonyfettes:
image

PR #1351, opened 2026-09-07
fix(desktop): clarify transcript work and tool disclosures

peter-jerry-ye:
Repeated step metadata and separate tool request/result folds make a conversation harder to scan. This change groups routine work while keeping direct commentary, final answers, and consequential tool states clear.

- One **Work** disclosure groups contiguous activity. It starts expanded, with commentary visible and independent **Thinking** and tool disclosures. Per-step captions, ambiguous commit clocks, and text-preview summaries are removed.
- Each tool has one section for its arguments and output. Shared status icons distinguish pending, available, and failed results; input/output share typography and alignment. Reasoning and tool evidence use secondary text beneath direct commentary.
- Live prose and final answers remain outside Work, including when a final answer carries reasoning. Existing Copy scope stays on the final message, with a quiet end line and completion time. User send times appear on hover or keyboard focus, and stay visible without hover. The competing composer border and blinking stream cursor are removed.

Work is keyed by its preceding transcript boundary, and tool details by invocation identity. This preserves disclosure and selected-tab state as live reasoning commits or parallel results arrive out of order. Prompts, checkpoints, runtime notices, and terminal errors retain their order outside the group. Collapsed labels keep pending/failure state discoverable; tool captions are bounded before entering the accessible name, with longer captions available in the body.

The shared renderer applies to both providers. Existing job-wait descriptions, plan progress, specialized arguments/output, failure stages, and subagent transcript links are retained. Event decoding, approvals, stored history, and final-answer eligibility are unchanged. No generated summaries, inferred durations, or new dependencies are introduced.

The PR contains the renderer, styles, and regression coverage for these transitions. The [interactive replay and screenshot fixture](https://github.com/moonbitlang/openseek/tree/61c678b5110492d674033980d6febcba1427b494/desktop/frontend/replay) remain on the separate replay branch.

## Before and after

Real production components, identical synthetic events and timestamps, and a 1280 × 850 viewport. Before is main at `635dc44`; after is this PR at `a4734b7`. Thinking and tool details are closed on both sides; Work starts expanded. These images demonstrate presentation, not real tool execution.

| Before — main | After — this PR |
| --- | --- |
| Main: numbered thoughts and separate requests/results | PR: one Work group with paired tools and quieter evidence |

<details>
<summary>Expanded Thinking and tool details</summary>

Arguments and output together with aligned typography

[Dark appearance](https://raw.githubusercontent.com/moonbitlang/openseek/61c678b5110492d674033980d6febcba1427b494/desktop/frontend/replay/screenshots/current-dark.jpg) · [390px layout](https://raw.githubusercontent.com/moonbitlang/openseek/61c678b5110492d674033980d6febcba1427b494/desktop/frontend/replay/screenshots/current-narrow.jpg)

</details>

### DIFF
--- a/desktop/e2e/tests/rabbita_dom.spec.js
+++ b/desktop/e2e/tests/rabbita_dom.spec.js
@@ -767,10 +767,10 @@ test('tool-call tabs keep focus-driven scrolling inside the transcript', async (
   await app.openSession();
 
   const transcript = page.locator('#transcript');
-  await expect(transcript.locator('.tool-call-summary')).toContainText('Print <browser> fixture');
+  await expect(transcript.locator('.tool-toggle')).toContainText('Print <browser> fixture');
   const tabs = transcript.locator('.tool-call-tabs');
   const originalJson = tabs.getByText('Original JSON', { exact: true });
-  await transcript.locator('.tool-call-summary').click();
+  await transcript.locator('.tool-toggle').click();
   await transcript.evaluate(node => {
     node.scrollTop = node.scrollHeight;
   });
@@ -828,8 +828,9 @@ test('runtime notices keep the compact result-row presentation', async ({ page }
   expect(app.pageErrors).toEqual([]);
 });
 
-test('a model step shows its thought, then its prose, then its tool rows', async ({ page }) => {
+test('work groups commentary with independent thinking and paired tool disclosures', async ({ page }) => {
   const app = new DesktopBrowserHarness(page);
+  const reasoning = '🔎 first thought with supporting evidence. '.repeat(200) + 'End of reasoning.';
   app.sessionEvents = [
     {
       sequence: 1,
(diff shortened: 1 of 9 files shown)


desktop/frontend/transcript/component/activity_disclosure.mbt
tonyfettes, line 1: Missing DOM binding from rabbita, using extern "js" here.

tonyfettes (commented):
We want:

```
> fold everything except last message
  > fold one turn
    > fold reasoning of one turn
    activity message
    > fold tool call
```

This PR:

```
> fold one turn
> fold one turn
```

tonyfettes:
Before/after comparison for the pending `job_wait` E2E fixture.

These are real Desktop browser component renders using identical synthetic event data, a 1280 × 720 viewport, and the same 1020 × 310 transcript crop. Before: main at `635dc4430`. After: PR at `d5e7939dd` (the last commit changes only the test).

The fixture represents waiting for background jobs: `bg-10` has the description “Watch CI on rebased PR 27”; `bg-11` has no description. It does not launch an actual CI monitor.

**Before — tool labels are visible without expanding the activity**

Before: pending job IDs and description visible

**After — default collapsed activity**

After collapsed: Tool activity and Pending metadata

**After — both activity groups expanded**

After expanded: original job IDs and description visible

The collapsed view is more compact, but the specific jobs being awaited are only visible after expansion. The E2E failure came from clicking an inner tool summary while its outer activity was still closed. Commit `d5e7939dd` adds that expansion step and a visibility assertion; it does not change this UI behavior.

PR #1365, opened 2026-09-08
chore(desktop): update bundled MoonBit and complete mbtx path rewriting

tonyfettes:
Update Desktop's bundled MoonBit compiler from `0.10.11+8f8e8db1e` to `0.10.12+1634b282e`. Complete `mbtx` diagnostic path rewriting for older standalone-build layouts with an additional `_build` layer, keeping diagnostics pointed at `source`.

The separate build output directory is already on main via #1367. This PR retains the compiler version bump, the additional diagnostic path mapping, and layout comments.

Validation after rebasing onto main, using Moon 0.1.20260907 / compiler `0.10.12+1634b282e`:

- `just check`: native and JS checks with warnings denied, plus formatting, passed.
- `moon test agent_tool/mbtx --target native`: 72/72 passed.
- `moon test agent_tool/mbtx --target wasm`: 72/72 passed.
- `just desktop-build-scripts-check` passed.
- `moon info && moon fmt` and diff checks passed; generated interfaces are unchanged.
- The rebased file contents match the previous PR head exactly. Full platform packages were not rebuilt during rebase validation.


tonyfettes:
~on hold b/c we might release another stable soon.~

PR #1369, opened 2026-09-08
Add explicit turn waiting for background jobs

bobzhang:
A plain-text answer can currently end a turn while background work still needs its result. Add `job_wait(job_ids)` as an explicit, event-driven control action: the same turn stays alive until a selected job settles or user steering arrives, and cancellation follows the existing Interrupted path. Pure-text completion with pending jobs asks the model to choose finish, wait, or further work; explicit `finish` keeps its existing semantics.

The loop owns waiting and input consumption. Completion publication and waiter wakeup share a predicate to cover the exit/notice gap; unrelated job notices remain queued. Desktop pending-tool rows show the selected IDs using the existing running/input/cancel lifecycle. No cross-turn restart or persistent waiting state is introduced. The design is recorded in `docs/turn_finish.md`.

First of two stacked PRs. This PR retains `job_output.wait_ms`; follow-up #1370 removes that parameter once explicit waiting is available.

Validation:
- `just check` and `just build` (native and JS).
- Offline `just test`: 3237 native tests, 3196 JS tests, and 24 cram cases passed. Optional live-provider credentials were unset for the test run.
- Real CLI/local scripted provider test: background execution → text reminder → wait → completion notice in model context → output → finish.
- Focused tests cover invalid IDs/arguments, mixed batches, step budget, selected/unrelated completion, terminal/notice ordering, steering, stop, and cancellation; 14 agent job tests passed again after self-review hardened cancellation-result persistence.
- `moon info --target native`, `moon fmt`, and generated API diff inspection.

Self-review checked explicit finish, notification ownership, cancellation, and input queue ownership. No manual desktop GUI run or live-model smoke test was performed.

Review follow-up:
- Scope finish reminders to jobs registered after the current loop starts. Existing servers from earlier turns remain explicitly waitable but do not add a reminder round-trip to unrelated answers.
- User-originated goal notices wake `job_wait` while retaining their existing persisted representation. Ordinary background notices do not wake it, and queued goal commands retain their existing wire order.
- Wire the real CLI test into `just test` and native CI. Add a real `serve` stdin goal-clear scenario that must resume the model while its selected job remains gated, plus regressions for old/new jobs and wake-state reset.
- Clarify that `control` grants context-ceiling execution, document watchdog termination, and make the registry field private behind an accessor.

The agreed explicit-finish policy remains: a second plain-text answer is not silently treated as permission to abandon results. The design now explicitly notes that the default max-steps value is not practical loop protection. Moving the concrete registry dependency out of the tool-contract package remains a separate architectural follow-up; the accessor only encapsulates its storage. Validation still happens before racing against user input.

Follow-up validation: `just check`, offline `just test` (3237 native, 3196 JS, 24 cram, and both real CLI scenarios), `just build`, and final interface/format generation all passed.


bobzhang:
## Architecture review

**Verdict.** The direction is right and the concurrency core is correct. Keeping the wait inside the same turn, letting explicit `finish` win, and gating both the finish check and the waiter on one `completion_published` predicate are the correct calls. Two things the design doc does not cover deserve a decision before merge: the reminder is turn-scoped but the registry it consults is process-scoped, and the tool registry has become a service locator.

### Findings, most important first

1. **Every later turn in `serve` pays for a long-lived job.** `serve` builds the tool registry once (`cmd/openseek/serve.mbt:693`) and every turn reuses the same job runtime. A dev server started in turn 1 stays in `pending()` for the process lifetime. So in turn 2, 3, and onward, any plain-text answer triggers the reminder and costs one extra model round-trip before the model calls `finish`. Desktop rides on `serve`, so this is the normal path, and `docs/turn_finish.md` defines "active" with no turn boundary. Suggest: remind once per turn per pending-id set, or only about jobs registered since the turn began. There is precedent for once-only nagging in `GoalCadence::due_finish_check` (`agent/goal.mbt:137`).

2. **The reminder loop is bounded only by `max_steps`.** A model that keeps answering in plain text with a live server gets the reminder on every step until the turn fails with "max steps exhausted", for work that was actually done. The doc punts on this deliberately, but the fix is the same cheap mechanism as item 1: a second consecutive plain text with an unchanged pending set finishes normally.

3. **`Tools` is now a service locator, and `agent_tool` gained a dependency on `bgjobs`.** Before this PR `agent_tool/moon.pkg` did not import `agent_tool/bgjobs`. Now the abstract tool-contract package imports a concrete runtime and exports it as a public struct field (`agent_tool/agent_tool.mbt:235`, visible in `pkg.generated.mbti`). The only consumer is the loop, which could hold the runtime itself: `run_turn_in_scope` already builds tools when none are given. Minimum for this PR: make the field private with an accessor. Better in a follow-up: thread the runtime into `AgentLoop` directly.

4. **A `Control` action from a non-`control` tool blurs what the flag means.** The field doc at `agent_tool/agent_tool.mbt:44` says a control tool's action is loop flow. `job_wait` returns `Control(WaitJobs)` but must not set `control`, because that flag really means "safe to execute sight-unseen at the context ceiling". The code is correct today because the guards in `execute_tool_call` are explicit and prefetch skips it by default. One sentence on the field doc would stop the next person from "fixing" it by setting `control=true`, which would make the ceiling salvage start a wait it cannot resume.

5. **A goal change during a wait does not wake it.** In `serve`, a `/goal` arriving during an active turn rides a `Notice` (`cmd/openseek/serve.mbt:1169`), and `wait_for_user_input` wakes only on `Prompt` or `Command`. So a user redefining the goal while the turn waits is not treated as steering. Cancel still works. Decide it explicitly either way and record it in the doc.

6. **Validation runs three times.** Tool decode, `AgentLoop::wait_for_jobs`, and `BgJobRuntime::wait_any` each re-check emptiness and unknown ids. The loop-level copy duplicates the registry exactly. The registry is the natural owner; let the loop surface its error.

7. **Two vocabularies for one job state.** The completion notice says "killed: flooded past the output cap" while the wait result says "stopped killed_at_output_cap=true". The model sees both in adjacent messages. A single status formatter on `BgJobSnapshot` in `bgjobs` would fix it.

8. **The cancellation special-case is a one-off.** Recording "job_wait cancelled" under `protect_from_cancel` is for transcript clarity, not API validity: the projection already closes unpaired calls with synthetic errors (`agent_session/projection.mbt:127`). No other cancelled tool does this. Either generalize it in `seal_unwinding_turn` or drop it.

9. **`user_input_pending` is a shadow of the queue held by convention.** It is correct today: every drain goes through `drain_steers`, and all three `serve` drain sites run only when no turn is active. The "worth acting on" predicate is now defined twice, in `queue_steer` and in `pending_steers`. Wrapping the flag and the condition variable into one small type would give the invariant one home.

10. **The Python integration test is not run by anything.** Nothing in the justfile or CI invokes `tests/integration/turn_finish.py`, and it is the repo's first Python file. Its scenario is already covered by `exercise_job_wait` in MoonBit. Wire it into CI or drop it.

### Verified and agreed

- **No lost wakeups.** `Cond::wait` enqueues the waiter before suspending, and `wait_any` has no yield between the predicate check and the wait. Under the cooperative scheduler the publication order in the watcher's `defer` (cleanup, notice queued, flag set, broadcast) cannot be missed.
- **The `@async.pause()` after `any()` is justified.** The library's `Cond::wait` deliberately swallows a cancellation that lands after the waiter was woken (`cond_var.mbt`, `Waiter.woken`), so a fresh cancellation check is needed before recording success.
- **Waits are bounded despite "no timeout".** Jobs are never removed from the registry, and every job has a wall-clock reaper, so the worst case is the 30-minute cap. Worth one line in the tool description so the model expects a "killed: outlived the wall-clock cap" state.
- **Loop-owned wait over tool-owned wait** is the right split. The steer queue has one consumer and the alternative would have handed the runtime to a tool executor.
- **Desktop change** is minimal and reuses the existing pending-row lifecycle.

---
🤖 Architecture review generated with [Claude Code](https://claude.com/claude-code)

PR #1380, opened 2026-09-08
Show job descriptions in waiting rows

bobzhang:
Pending `job_wait` rows showed only IDs even when earlier tool rows already carried the job description. They now display labels such as `Waiting for bg-10: Watch CI on rebased PR 27`, with an ID-only fallback for jobs without a recorded description.

Resolve descriptions from prior `mbtx` launch metadata and `job_output` briefs within the current transcript, and include the resulting labels in memoized render blocks. Tool arguments and the wire protocol stay unchanged.

Validation: `just check`, the full native/JS tests and offline CLI checks, and `just build` passed for the initial implementation. After review fixes, `just check`, all 22 transcript component tests, and the focused Playwright regression passed. `moon info` and `moon fmt` completed with no generated interface changes. All GitHub CI checks passed on the updated commit, and OpenSeek’s follow-up ran checks and tests across all targets successfully.

Regression coverage includes multiple jobs, duplicate IDs, missing descriptions, malformed metadata, duplicate tool call IDs, and collapsed-row rendering. Description parsing is limited to actual background `mbtx` launches, and waiting labels are keyed by call position.

### DIFF
--- a/desktop/e2e/tests/rabbita_dom.spec.js
+++ b/desktop/e2e/tests/rabbita_dom.spec.js
@@ -1876,3 +1876,32 @@ test('composer follow-up menu supports keyboard choice and a queue default', asy
   expect(app.requests.filter(r => r.method === 'settings.set')).toEqual([]);
   expect(app.pageErrors).toEqual([]);
 });
+
+test('pending job waits show descriptions from earlier tool rows', async ({ page }) => {
+  const app = new DesktopBrowserHarness(page);
+  app.sessionEvents = [
+    { sequence: 1, item: { kind: 'user', payload: { content: 'Show the browser fixture waiting for CI' } } },
+    { sequence: 2, item: { kind: 'assistant', payload: {
+      content: '', tool_calls: [{ id: 'start', name: 'mbtx',
+        arguments: JSON.stringify({ description: 'Watch CI on rebased PR 27' }) }],
+    } } },
+    { sequence: 3, item: { kind: 'tool_result', payload: {
+      tool_call_id: 'start', tool_name: 'mbtx', content: '',
+      is_error: false, brief: 'mbtx → bg bg-10',
+    } } },
+    { sequence: 4, item: { kind: 'assistant', payload: {
+      content: '', tool_calls: [{ id: 'wait', name: 'job_wait',
+        arguments: JSON.stringify({ job_ids: ['bg-10', 'bg-11'] }) }],
+    } } },
+  ];
+  await app.install();
+  await app.goto();
+  await app.openSession();
+  const wait = page.locator('details.tool-call').filter({ hasText: 'Waiting for' });
+  await expect(wait.locator('.tool-call-text')).toHaveText(
+    'Waiting for bg-10: Watch CI on rebased PR 27, bg-11');
+  await expect(wait).not.toHaveAttribute('open', '');
+  await wait.locator('summary').click();
+  await expect(wait).toContainText('"job_ids"');
+  expect(app.pageErrors).toEqual([]);
+});
--- a/desktop/e2e/tests/rabbita_dom.spec.js
+++ b/desktop/e2e/tests/rabbita_dom.spec.js
@@ -1891,7 +1891,8 @@ test('pending job waits show descriptions from earlier tool rows', async ({ page
     } } },
     { sequence: 4, item: { kind: 'assistant', payload: {
       content: '', tool_calls: [{ id: 'wait', name: 'job_wait',
-        arguments: JSON.stringify({ job_ids: ['bg-10', 'bg-11'] }) }],
+        arguments: JSON.stringify({ job_ids: ['bg-10', 'bg-11'] }) },
+        { id: 'wait', name: 'job_wait', arguments: JSON.stringify({ job_ids: ['bg-11'] }) }],
     } } },
   ];
   await app.install();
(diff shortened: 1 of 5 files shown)


bobzhang (commented):
OpenSeek review of `f7da530c27bd62f9759e18a5471c69d33c7f125d`

The change enriches pending `job_wait` tool rows with human-readable descriptions collected from earlier `mbtx`/`job_output` rows, threads a per-item `wait_labels` map through the block/render pipeline, and adds unit plus e2e coverage. The logic is correct for the normal case and the new tests pass. Two issues are worth addressing: the description collector parses the full JSON of every successful mbtx call on every render pass (including per streaming token) before checking whether the brief is even a background-job brief, and the labels map is keyed by `call_id` alone, which can mislabel duplicate call_ids that the renderer otherwise disambiguates by index.

**MEDIUM: collect_job_description parses every successful mbtx call's full arguments before checking the bg brief**

`desktop/frontend/transcript/component/job_wait.mbt:45`

`collect_job_description` runs inside `stream_blocks`, which `component.mbt:307` recomputes on every `Input` change — including every streaming token, since `streaming_response` is part of `Input`. For every `Done(is_error=false, brief=Some(_))` mbtx call it calls `@json.parse(arguments)` at line 45 before ever testing `brief.strip_prefix("mbtx → bg ")` at line 47. A foreground mbtx success also has a brief (`mbtx (exit=N)`), so every completed mbtx call in the transcript re-parses its full `arguments` JSON on each streaming delta. `arguments` for mbtx is the model's whole MoonBit `source` snippet, and `tool_card.mbt:125-131` already guards card parsing at 128k for exactly this reason. This reintroduces per-delta cost proportional to transcript length, contrary to the memoization design in `component.mbt:303-306`.

Suggested fix: For `call.name == "mbtx"`, check `brief.strip_prefix("mbtx → bg ")` (and non-blank id) before `@json.parse(arguments)`, so only actual background-adoption calls are parsed. For `job_output` the arguments are small, but a size guard could also be applied for symmetry.

**LOW: wait_labels keyed by call_id alone can mislabel duplicate call_ids in one item**

`desktop/frontend/transcript/component/job_wait.mbt:77`

`job_wait_labels` stores `labels[call.call_id] = label`, and `tool_request_view` (view.mbt:298) looks the label up by `call.call_id` only. `tool_run_row_key` (view.mbt:228-234) deliberately includes the array index because duplicate call_ids in malformed logs are possible. If one Assistant item contains two pending `job_wait` calls with the same `call_id` but different `job_ids`, the second label overwrites the first and both rows render the second call's label. Before this change the per-call fallback `pending_job_wait_label(call)` computed each call's own label, so duplicates were shown correctly (un-enriched).

Suggested fix: Iterate with `for index, call in calls` and key labels by the same index-aware key used for rows (e.g. `tool_run_row_key(call, index, "request")`), then have `tool_request_view` receive that key; or otherwise avoid a bare call_id-keyed map for calls whose ids can collide.

Validation reported by OpenSeek: moon check --target all: passed (1553 tasks, up to date); moon test: all passed (98 wasm, 1895 js, 1965 native).

Both findings will be addressed before merge.

bobzhang (commented):
OpenSeek follow-up review of `105790247bd824e03596ba73068df2b6433775fd` (corrections since `f7da530c27bd62f9759e18a5471c69d33c7f125d`).

The change switches wait_labels from call_id-keyed to index-keyed maps so duplicate tool-call ids within one assistant event no longer overwrite each other's pending job_wait captions. The core refactor in collect_job_description is behavior-preserving, the index propagation (job_wait_labels -> BlockContent::Item -> assistant_view -> tool_run_views -> tool_request_view) is consistent, and moon check/test pass on all targets. One low-severity gap remains: the tabbed disclosure inside each tool row still keys only by call_id, so the newly supported duplicate-id scenario renders colliding radio groups and duplicate DOM ids.

**LOW: Duplicate call IDs still collide in the tool-call tab radio groups**

`desktop/frontend/transcript/component/view.mbt:345`

The change makes wait labels index-keyed (job_wait.mbt:77-80; view.mbt:189/207), so two pending job_wait calls sharing a call_id now get distinct labels. But each request's disclosure body still derives its tab key from tool_call_tabs_key(call.call_id, sequence) (view.mbt:345, and view.mbt:400 for results), which ignores the call index (tool_tabs.mbt:59-67). Two calls with the same call_id in one assistant event — exactly the scenario added in rabbitdom spec lines 1893-1895 and job_wait_wbtest lines 107-130 — therefore render two fieldsets whose radio inputs share the same name and duplicate element ids (tool_call_tabs uses name="tool-call-tabs-{key}" and id="{group}-{index}"). Radio group membership is by name across the whole document, so the two rows' selected-tab state is coupled: the second row's checked input unchecks the first row's default panel, and the second row's <label for> resolves to the first row's input (first match on duplicate ids). This is pre-existing, but the new tests make duplicate call IDs a supported rendering scenario without covering the tab interaction.

Suggested fix: Pass the call's loop index into tool_request_view/tool_result_view and include it in the tab key (e.g. reuse tool_run_row_key(call, index, side), or add an index parameter to tool_call_tabs_key) so duplicate-id calls get distinct radio group names and DOM ids.

Validation reported by OpenSeek: moon check --target all: passed (exit 0, 1248 tasks); moon test --target all: passed — wasm 2510/2510, wasm-gc 251/251, js 3199/3199, native 3234/3234. The Playwright spec desktop/e2e/tests/rabbita_dom.spec.js was not run (requires a browser)..

Maintainer assessment: both findings from the first review are addressed. The remaining low-severity tab-key collision is explicitly pre-existing; verified that the relevant tab-key code is unchanged from the PR base. It is non-blocking for this waiting-label fix. All GitHub CI checks passed on the reviewed commit; the focused Playwright regression also passed locally.

PR #1384, opened 2026-09-08
fix(desktop): keep the active dock tab visible

bobzhang:
Opening enough files to overflow the right-panel tab strip could leave the active tab and its close button offscreen. The strip retained its old horizontal position while hiding its scrollbar. Reproduced in Chromium with eight long file tabs: the active-tab geometry assertion fails on the original browser bundle.

Reveal the active tab after dock selection, membership, owner, or panel visibility changes, including explicitly reopening the same active file. Observe strip and active-tab size changes so panel resizing also keeps the close button reachable. Scroll only the strip, preserve deliberate scrolling during unrelated updates, and cap tab width to the available strip width. Disconnect old observations when rendered elements change.

Validation:
- Regression fails before the fix and passes afterward; covers opening, switching, closing, resizing, narrow strips, manual scrolling, and reopening the active file.
- Desktop browser suite: 45 tests passed.
- `just check`, `just test`, `just build`.
- `moon info` and `moon fmt`; no generated public interface changes.

### DIFF
--- a/desktop/e2e/tests/dock_visibility.spec.js
+++ b/desktop/e2e/tests/dock_visibility.spec.js
@@ -1,0 +1,62 @@
+import { test, expect } from '@playwright/test';
+import { DesktopBrowserHarness } from './support/desktop_browser_harness.js';
+
+// Geometry assertions must not click the tab first: Playwright would scroll
+// the target into view and conceal the product bug.
+async function expectActiveVisible(page) {
+  await expect.poll(() => page.locator('.editor-tabs').evaluate(strip => {
+    const active = strip.querySelector('.editor-tab.active');
+    if (!active) return false;
+    const bounds = strip.getBoundingClientRect();
+    const tab = active.getBoundingClientRect();
+    const close = active.querySelector('.tab-close').getBoundingClientRect();
+    return tab.left >= bounds.left - 1 && tab.right <= bounds.right + 1 &&
+      close.left >= bounds.left - 1 && close.right <= bounds.right + 1;
+  })).toBe(true);
+}
+
+test('active dock tab stays visible when opening, switching, closing and resizing', async ({ page }) => {
+  const app = new DesktopBrowserHarness(page);
+  app.searchFiles = Array.from({ length: 8 }, (_, i) => `src/document_${i}_long_name.mbt`);
+  for (const path of app.searchFiles) app.workingFiles[path] = 'fn main {}\n';
+  await app.install();
+  await app.goto();
+  await app.openSession();
+  const openFile = async i => {
+    await app.openQuickOpen();
+    await page.locator('#quick-open-input').fill(`document_${i}`);
+    await page.getByRole('option', { name: new RegExp(`document_${i}`) }).click();
+    await expect(page.locator('.editor-tab.active')).toContainText(`document_${i}`);
+  };
+  for (let i = 0; i < 8; i++) await openFile(i);
+  await expect.poll(() => page.locator('.editor-tabs').evaluate(el => el.scrollWidth > el.clientWidth)).toBe(true);
+  await expectActiveVisible(page);
+  await openFile(0);
+  await expectActiveVisible(page);
+  await openFile(7);
+  await expectActiveVisible(page);
+  await page.getByRole('button', { name: 'Expand panel', exact: true }).click();
+  await expectActiveVisible(page);
+  await page.getByRole('button', { name: 'Restore panel', exact: true }).click();
+  await page.setViewportSize({ width: 1100, height: 900 });
+  await expectActiveVisible(page);
+  // A drag changes CSS geometry without a model message. Also exercise a
+  // strip narrower than the normal 200px chip cap.
+  const strip = page.locator('.editor-tabs');
+  await strip.evaluate(el => { el.style.flex = '0 1 120px'; });
+  await expect.poll(() => strip.evaluate(el => el.clientWidth)).toBeLessThan(200);
+  await expectActiveVisible(page);
+  await strip.evaluate(el => { el.style.removeProperty('flex'); });
+  await expectActiveVisible(page);
+  // Ordinary typing must not undo a user's deliberate horizontal scroll.
+  await strip.evaluate(el => { el.scrollLeft = 0; });
+  await page.locator('#task').fill('keep browsing the earlier tabs');
+  await expect.poll(() => strip.evaluate(el => el.scrollLeft)).toBe(0);
+  // Re-opening even the same active file is an explicit reveal request.
+  await openFile(7);
+  await expectActiveVisible(page);
+  await page.locator('.editor-tab.active .tab-close').click();
+  await expect(page.locator('.editor-tab')).toHaveCount(7);
+  await expectActiveVisible(page);
+  expect(app.pageErrors).toEqual([]);
+});
--- a/desktop/frontend/dock_visibility.mbt
+++ b/desktop/frontend/dock_visibility.mbt
@@ -1,0 +1,53 @@
+// DOM geometry belongs after rendering. Keep the observer with the current
+// strip, disconnecting it when the shell replaces/removes that element.
+
+///|
+extern "js" fn make_dock_visibility_sync() -> (Bool) -> Unit =
+  #| () => {
+  #|   let strip = null;
+  #|   let active = null;
+  #|   const reveal = () => {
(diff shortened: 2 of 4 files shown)


bobzhang (commented):
# Independent Codex CLI review

Reviewed commit: `14e17f935c8cd517326d3fe681dff222aca6add4` (one-commit diff: `git diff 14e17f93^ 14e17f93`). Reviewer: Codex, GPT-6. Reviewed independently without implementation conversation or other reviewers' comments.

**No actionable findings.**

Inspected the actual diff and surrounding dock state transitions, indexed DOM rendering, panel CSS, and Rabbita's after-render scheduling. Checked the potential stale-node and premature-geometry paths in `desktop/frontend/dock_visibility.mbt:9–30` and `desktop/frontend/update.mbt:1847–1854`: the callback runs after DOM patching, refreshes the current nodes, and disconnects/rebinds observations when their identities change. Explicit model comparisons cover tab identity changes even when Rabbita reuses a DOM node.

Validation passed:

- `just check` and `just build` (native and JS; formatting checked without rewriting files).
- `just test`: 3,234 native tests, 3,197 JS tests, 24 cram cases, and both CLI lifecycle checks.
- Fresh Desktop browser bundle; Chromium `dock_visibility.spec.js` (1 test) and `rabbita_dom.spec.js` (39 tests).

The new geometry test (`desktop/e2e/tests/dock_visibility.spec.js:18–62`) verifies opening/reactivating/closing tabs, expansion, viewport and narrow-strip resizing, close-button visibility, and preservation of manual scrolling during typing. Assertions avoid clicking the active tab before checking its visibility.

Validation limits: browser execution used the mocked host transport in Chromium, not the native CEF application or other browser engines. Observer disconnection on actual strip replacement/removal was checked by code inspection, not instrumented lifecycle assertions. The resize test changes CSS directly rather than dragging the real resize handle. No product files were changed.

bobzhang (commented):
# Independent OpenSeek review

- **PR**: https://github.com/moonbitlang/openseek/pull/1384 — `fix(desktop): keep the active dock tab visible`
- **Reviewed SHA**: `14e17f935c8cd517326d3fe681dff222aca6add4` (verified as the PR head via `gh pr view` immediately before writing this review)
- **Reviewer model**: not identified (automated independent reviewer; no access to prior review conversation)
- **Scope**: full diff of `14e17f93^..14e17f93` plus surrounding code (`desktop/frontend/dock/*`, `desktop/frontend/update.mbt`, `desktop/frontend/quick_open/*`, `desktop/styles/*`, `desktop/e2e/*`)

## What the PR does

Adds a `ResizeObserver`-backed reveal for the active dock tab (`desktop/frontend/dock_visibility.mbt`), wired into the root update batch (`desktop/frontend/update.mbt:1847-1854`), changes `.editor-tab` max-width to `min(200px, 100%)` (`desktop/styles/file_panel.css:96`), and adds one Playwright e2e test (`desktop/e2e/tests/dock_visibility.spec.js`).

## Validation performed

- `moon check --target js --deny-warn` (root workspace): clean.
- `moon check --target native --deny-warn` (root workspace): clean.
- `moon fmt --check` (workspace scope incl. `desktop/`): clean.
- `moon test --target js desktop/frontend`: 465/465 pass (frontend is a JS-only package; native is unsupported there by design).
- Built the real browser bundle via the repo's own packager (`moon run package/browser --target native`, i.e. `node package/build.mjs browser`) from this SHA and ran the committed Playwright suites against it: `dock_visibility.spec.js` (1/1), `close_shortcut.spec.js` (5/5), `rabbita_dom.spec.js` (passing, per `.last-run.json`).
- Wrote temporary probe scenarios (removed afterwards; not product code) against the same bundle:
  - natural **narrow layout** (viewport 700px): opening 8 long-named tabs, switching back to the first, and native-menu close (`app.close_focused`) of the last tab all keep the active tab + close button visible — passed;
  - the committed test's own scenarios (expand/restore, viewport resize, 120px squeeze, deliberate-scroll preservation, close of the active tab) — passed.
- CI for this head is already green: check (stable js/native/wasm), desktop e2e, desktop package (Windows ZIP), viz e2e.

## Findings

### Low severity

1. **Redundant force clause in `desktop/frontend/update.mbt:1853`.** `msg is Dock(TabActivated(_))` is a subset of the first condition (`model.dock.active != previous.dock.active`, line 1848): the only effect of `TabActivated` on state is setting `active` (`desktop/frontend/dock/update.mbt:23-28`), and when it does change, the first clause already forces the reveal. The extra clause additionally forces a reveal for *stale* `TabActivated` clicks that change nothing (a tab no longer in the strip), which is harmless but unintended. Suggest dropping it or noting why it is retained.

2. **No unit-level coverage of the reveal decision.** The force-flag computation is inline in `update` and has no white-box test (the only nearby white-box test covers `opens_file_before_panel_reconciliation`, `update.mbt:16495`). All regression pressure sits on the single e2e test. A cheap `_wbtest` over the force-flag cases (open/switch/close/owner/panel-toggle/quick-open-pick vs. composer typing) would lock the behavior independent of Playwright. Minor.

### Info / suggestions

3. **Committed e2e test does not exercise the natural `<760px` narrow layout** (full-screen editor overlay, `desktop/styles/responsive.css:98-113`); its "narrower than 200px" scenario is a synthetic `flex: 0 1 120px` inline style. My probe of the real narrow layout passed, so this is a coverage gap rather than a bug — consider folding it into the suite. Likewise, native-menu close (`app.close_focused`) near the right edge of a scrolled strip is only covered implicitly by `close_shortcut.spec.js` (which never scrolls the strip); my probe passed.

4. **Reveal lands flush against the strip edge** (`desktop/frontend/dock_visibility.mbt:17-18`): `scrollLeft += tab.right - viewport.right` puts the tab's right edge exactly at the viewport's right edge, clipping the 4px flex gap between chips. Cosmetic; an inset (e.g. `+= ... + gap`) would look better. Not a correctness issue.

### Things I checked that are correct

- **`after_render` semantics**: rabbita `EffectKind::AfterRender` runs after DOM patching for the flush (`moonbit-community/rabbita@0.15.4/cmd/commands.mbt:55`), so the reveal reads post-patch geometry; this matches the established `after_render` usage elsewhere (`dom_helpers.mbt:14`, `quick_open/view.mbt:42`).
- **DOM lifecycle**: the sync function re-queries `.editor-tabs` and `.editor-tab.active` on every transition and disconnects/re-observes only when either element identity changes (`dock_visibility.mbt:24-32`). The strip is always rendered by design (`dock/view.mbt:186,232-274`), so the observer is never left watching a permanently detached node, and the `clientWidth === 0` guard covers the hidden/zero-width panel.
- **Scroll correctness**: deltas are computed from `getBoundingClientRect` in the same coordinate space; only `strip.scrollLeft` is mutated (not `scrollIntoView`, which could move ancestors — the comment at `dock_visibility.mbt:15-16` is accurate). Mutating `scrollLeft` fires `scroll`, not `resize`, so there is no observer loop.
- **Force-flag completeness**: open/switch/close (active/tabs diffs), owner swap, panel toggle, and Quick Open pick (`opens_file_before_panel_reconciliation`, incl. `QuickOpen(Pick)` reaching `open_editor_at` in the same transition per `component_quick_open.mbt:53-83`) all force a reveal; expand/restore and window resizes are caught by the `ResizeObserver` on the strip itself.
- **Deliberate-scroll preservation**: composer typing does not force a reveal (not in the force list), which the e2e test asserts explicitly.
- **CSS change**: `max-width: min(200px, 100%)` is sound — the percentage resolves against the flex container, and `flex: 0 0 auto` + `overflow-x: auto` make the strip a horizontal scroller with a hidden scrollbar (`file_panel.css:75-86`), so the tab plus its 16px close button can never exceed the viewport.
- **No interface drift**: `moon info --target js desktop/frontend` shows no `.mbti` change (the new module is entirely private).

## Validation limits

- I could not run the native CEF/proton host; macOS menu key-equivalent handling of Close remains a native-test gap acknowledged in `close_shortcut.spec.js:4-5`.
- Windows/Linux classic-scrollbar rendering is untested here, though the strip hides scrollbars entirely (`scrollbar-width: none`).
- The global `document.querySelector('.editor-tabs')` assumes exactly one strip in the document; true today (single always-rendered dock), worth remembering if a second strip is ever introduced.

## Conclusion

No blockers. The implementation is correct, well-integrated with the framework's render lifecycle, and adequately validated end-to-end. The only actionable items are the redundant force clause (update.mbt:1853) and the small test-coverage suggestions (items 2 and 3), all low priority.

PR #1390, opened 2026-09-08
Add OpenRouter as a DeepSeek backend preset

mizchi:
## Summary

- add OpenRouter to the existing API endpoint selector without changing the host settings protocol
- persist OpenRouter through the existing custom backend fields with its canonical chat-completions URL
- accept an OpenRouter API key and supported DeepSeek model in initial setup
- translate DeepSeek model and reasoning fields at the HTTP transport boundary
- normalize OpenRouter reasoning fields in streaming and non-streaming responses

## Test plan

- `moon test deepseek/client` (36 passed)
- `moon test deepseek` (47 passed)
- `moon test --target js deepseek` (47 passed)
- `moon test --target js desktop/frontend` (471 passed)
- `npm test -- --grep "initial API setup accepts an OpenRouter"` (1 passed)

## Known repository issue

`just check`, `just build`, `just test`, and `moon info` currently stop in unchanged `eval/token_bench/bench.mbt`, which references removed DeepSeek provider/model APIs. The focused packages and OpenRouter browser flow pass as listed above.

### DIFF
--- a/desktop/frontend/model.mbt
+++ b/desktop/frontend/model.mbt
@@ -404,9 +404,13 @@ fn AppFontSize::pixels(self : AppFontSize) -> Double {
 }
 
 ///|
+const OpenRouterChatCompletionsUrl : String = "https://openrouter.ai/api/v1/chat/completions"
+
+///|
 /// Which chat-completions endpoint the engine talks to. SeekMoon runs are
 /// bring-your-own-key only: the hosted proxy is retired. `Deepseek` is the
-/// official endpoint and needs the user's own key; `Custom` is any other
+/// official endpoint and needs the user's own key; `OpenRouter` is a preset
+/// over the existing custom endpoint; `Custom` accepts any other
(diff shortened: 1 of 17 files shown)


tonyfettes:
The mapping of DeepSeek V4 Flash should be deepseek/deepseek-v4-flash-0731

desktop/frontend/model.mbt
```diff
@@ -403,10 +403,14 @@ fn AppFontSize::pixels(self : AppFontSize) -> Double {
   (self.base_pixels() - 1).to_double()
 }
 
+///|
+const OpenRouterChatCompletionsUrl : String = "https://openrouter.ai/api/v1/chat/completions"
```
tonyfettes, line 407: This URL is duplicated here because otherwise we need to add deepseek to the dependencies - it is rather whole a lot more ceremonial than it is now.

PR #1391, opened 2026-09-09
feat(mbtx): save scripts and run bundled workflows

bobzhang:
The `mbtx` tool can now run an existing script with `filename`, or save a reusable workspace script by accepting `source` and `filename` together. Existing `mbtx(source=...)` calls keep their behavior. A successful save returns the filename-only call to use next; saving refuses to overwrite different existing content and respects workspace, read-only, and worker write restrictions.

Bundled workflows use explicit names such as `mbtx(filename="@builtin/check.mbtx")`, resolved under `OPENSEEK_REFERENCES/workflow/`. Ordinary filenames resolve from the workspace root, independently of execution `cwd`. Namespaced scripts are read-only, cannot escape their root, and cannot accompany `source`. Only `@builtin/` is implemented; other namespaces are reserved for future extensions.

The schema, recovery messages, tool description, static system prompt, generated prompt, documentation, and transcript cards describe these call forms consistently. Saved scripts use the same isolated execution and sandbox path as inline source. The bundled resources themselves come from merged #1401.

Validation:
- Self-reviewed path resolution, overwrite protection, worker restrictions, diagnostics, schema, and prompt consistency.
- All five workflows passed through actual `@builtin/...` calls, including execution-directory and failure propagation checks.
- `just test` passed with live API credentials and `OPENSEEK_REFERENCES` unset: 3,265 native tests, 3,229 JS tests, 24 cram cases, and both CLI integration checks.
- `just build`, `moon info`, `moon fmt`, `just check-prompt`, and `git diff --check` passed.
- `just check` remains blocked by 10 pre-existing unused-import errors on the installed Moon nightly; these were also present before this change.

No new A/B evaluation was run.

### DIFF
--- a/agent_tool/mbtx/filename.mbt
+++ b/agent_tool/mbtx/filename.mbt
@@ -1,0 +1,56 @@
+///|
+fn is_bundled_filename(filename : String) -> Bool {
+  filename != "." &&
+  filename != ".." &&
+  !filename.contains("/") &&
+  !filename.contains("\\")
+}
+
+///|
+fn resolve_filename(
+  workspace_root : String,
+  filename : String,
+  references : String?,
+) -> String raise {
+  if is_bundled_filename(filename) {
+    guard references is Some(root) && !root.trim().is_empty() else {
+      fail(
+        "bundled filename \{filename} needs OPENSEEK_REFERENCES. For a workspace script, use ./\{filename}; bundled workflows are unavailable in this environment",
+      )
+    }
+    @workspace_path.resolve(root, "workflow/\{filename}")
+  } else {
+    @workspace_path.resolve(workspace_root, filename)
+  }
+}
+
+///|
+test "bare filenames select bundled workflows without workspace fallback" {
+  assert_eq(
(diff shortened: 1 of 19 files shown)


desktop/frontend/transcript/component/mbtx.mbt
```diff
@@ -18,8 +18,8 @@
 ///|
 /// A recorded `mbtx` call, decoded for display.
 priv struct MbtxCall {
-  /// The `.mbtx` program, exactly as the model sent it.
-  source : String
+  /// The inline program, if supplied. Saved calls show the filename chip.
+  source : String?
```
bobzhang: also display as either Source or Filename?

agent_tool/mbtx/filename.mbt
```diff
@@ -0,0 +1,56 @@
+///|
+fn is_bundled_filename(filename : String) -> Bool {
+  filename != "." &&
```
bobzhang: !finame.starts_with(".") | ".." | "/" && not contains "/" "\\" ?

PR #1403, opened 2026-09-09
refactor: describe static json pattern content as the json it matches

bobzhang:
## What

Two connected cleanups over JSON decodes, driven by one rule: **wherever a
JSON decode's content is static — a fixed command/kind/enum literal, a
`true`/`false` flag — describe it statically**; don't decode it through
dynamic machinery (nested `match` over the same Json, a second
`line is { ... }` boolean test, Option/Map plumbing), and don't spell it
through the enum (`String("prompt")`, `True`) when the pattern can read
like the json itself (`"prompt"`, `true`).

## 1. `Command::parse` / `parse_steer` — one static arm per wire shape

- `prompt` (`submission_id`): tagged / wrongly-tagged / untagged lines are
  three flat arms — the nested `match line { ... }` that restated the wire
  shape is gone.
- `goal` set (`auto`): the armed line (`"auto": true`) and the manual line
  are two static arms; absent and non-true both mean false. The previous
  `auto=line is { "auto": True, .. }` second test over the same Json is
  gone.
- `steer` (`kind`): each kind is its own static arm —
  `{ "kind": "prompt", "text": String(text), .. }` — a present kind that is
  neither is its own refusal, and the no-kind prompt steer is an explicit
  arm instead of a `_` fallback that duplicated the `"prompt"` result.
- `approval` (`allow`): the static `"allow": true` / `"allow": false`
  mirror arms, now spelled like the json.

## 2. Repo sweep — spell static pattern content as json

Static literals inside Json patterns spelled as `true`/`false` and bare
strings where the compiler accepts the json spelling (it does, for
object patterns over Json and over `Map[String, Json]`, and for
`Option[Json]` value matches):

- tool-argument decoders: `mbtx` (`warning` on/off, `escalated`,
  `subrun`), `write`, `multi_edit`, `plan` step statuses
  (`"pending"`/`"in_progress"`/`"completed"`).
- `desktop/internal/remote/serve.mbt`: `Some("2.0")` (JSON-RPC version
  guard); `desktop/internal/api/api.mbt`: `Some("paginated")`.
- `desktop/internal/protocol/engine_command.mbt`: the wire-shape
  assertions of `to_json()` output now read as the json the protocol
  writes.

Places with the same spelling pattern that were left out of this PR (to
keep the diff reviewable): eval harnesses (`eval/bgjobs_capability`,
`eval/file_edit/harness`), `agent_tool/internal/auto_check`,
`jsonrpc` tests, `viz/view.mbt`, host reply classifiers
(`desktop/internal/host`), and `agent_subrun/host/host_test.mbt`. They are
candidates for a follow-up spelling pass.

## Verification

- Spelling/restructure only: no behavior, error-string, or `.mbti`
  interface changes.
- `moon test`: protocol 24/24, mbtx/write/multi_edit decode 8/8/15,
  desktop `internal/remote` 9/9, `internal/protocol` engine-command 1/1.
- codex review sign-off posted on this PR.

Generated with SeekMoon

### DIFF
--- a/agent_tool/mbtx/internal/decode/decode.mbt
+++ b/agent_tool/mbtx/internal/decode/decode.mbt
@@ -82,22 +82,22 @@ pub fn decode(arguments : Json) -> MbtxInput raise {
         _ => None
       }
       let warning = match arguments {
-        { "warning": String("on"), .. } => true
-        { "warning": String("off"), .. } => false
+        { "warning": "on", .. } => true
+        { "warning": "off", .. } => false
         { "warning": Null, .. } => false
         { "warning": _, .. } =>
           fail("arguments.warning to be \"on\" or \"off\"")
         _ => false
       }
       let escalated = match arguments {
-        { "escalated": True, .. } => true
-        { "escalated": False, .. } | { "escalated": Null, .. } => false
+        { "escalated": true, .. } => true
+        { "escalated": false, .. } | { "escalated": Null, .. } => false
         { "escalated": _, .. } => fail("arguments.escalated to be a boolean")
         _ => false
       }
       let subrun = match arguments {
-        { "subrun": True, .. } => true
-        { "subrun": False, .. } | { "subrun": Null, .. } => false
+        { "subrun": true, .. } => true
+        { "subrun": false, .. } | { "subrun": Null, .. } => false
         { "subrun": _, .. } => fail("arguments.subrun to be a boolean")
         _ => false
       }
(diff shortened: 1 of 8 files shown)


bobzhang:
## codex review sign-off

`codex review --base origin/main` (OpenAI Codex v0.153.4) on the current
branch:

> The JSON pattern simplifications and flattened command matching preserve
> existing behavior. All 143 targeted tests passed; the workspace-wide
> native check was blocked by errors in unchanged editor tests.

No actionable feedback — exiting 0.

PR #1404, opened 2026-09-09
chore: make just recipes work on Windows

tonyfettes:
Windows recipes currently depend on POSIX shell behavior and launchers, so normal `just` commands fail before reaching the underlying tools. This change runs all four justfiles with Windows PowerShell on Windows while preserving the default shell on macOS/Linux.

- Resolve editor paths with `absolute_path`, preserve paths containing spaces, and omit an empty optional `--browse-root` argument.
- Invoke Playwright through Node and select `npm.cmd` on Windows.
- Select `python` on Windows for the existing Python lifecycle test, with a `PYTHON` override.
- Add `test-cram` and select Git for Windows' Bash explicitly, avoiding the WSL launcher on PATH. Keep the root test steps as ordered dependencies so overrides propagate.
- Document Python, Git Bash, path overrides, and the NSIS packaging prerequisite; remove the stale jq prerequisite.

### Windows validation

Executed real recipes using Windows PowerShell, just 1.37.0, the repository-pinned MoonBit v0.10.12+1634b282e in an isolated toolchain, the VS 2022 amd64 environment, Node 23.11.0, and Python 3.11.9.

Passed:
- `just build` (native and JS), `just editor-build`, `just cef-setup`, and `just desktop-build-scripts-check`.
- `just test-cram`: 24/24; `just viz-test-browser`: 13/13.
- `just desktop-package`: ZIP and installer generated after adding the installed NSIS directory to the process PATH. The installer was not run.
- Supplementary JS checks and tests: strict check passed, 3217/3217 tests passed.
- Inspector and editor serve/dev startup returned HTTP 200, including editor paths containing spaces. Desktop dev created a window and reported renderer bridge ready; native UI interaction was not tested. All validation services were stopped.
- `moon info`, `moon fmt`, `moon fmt --check`, and `git diff --check`; no generated interface changes.

The full Windows suite is **not green**:
- `just check`: native `--deny-warn` rejects five unused fields/variables in bgjobs and worktree code.
- `just test` and `just test-moon`: native `deepseek/client` diagnostics tests hung and were manually interrupted. Later root test stages were therefore exercised separately.
- `just test-turn-finish`: the run case passes; the serve process does not exit after `agent_finished` and is terminated by the 120-second test timer.
- `just editor-test`: JS 1875/1875; Wasm 1086/1090 and native 1220/1224, with the same four diagnostics-refresh failures; wasm-gc has no test entry.
- `just editor-test-browser`: 105/107; the component-only rerun is 68/70, reproducing the same read_api/viewer_api geometry failures.
- `just desktop-test-browser`: 56/57; the transcript Markdown local-file-link locator fails.

These failures remain unresolved in this PR; they have not been compared against a separate base-branch run. Tests that self-skip for missing online credentials or `sh` are not counted as coverage. Validation above is on Windows; macOS/Linux were not rerun.

### DIFF
--- a/README.mbt.md
+++ b/README.mbt.md
@@ -33,9 +33,9 @@ The root [`moon.work`](moon.work) develops OpenSeek, the desktop app, and the
 smaller `editor/moon.work` as a scoped entry point for editor-only builds and
 browser tests; root Moon commands are the integration gate across both projects.
 
-A fresh checkout needs the MoonBit toolchain and `just`; `just check` also
-requires `jq` to inspect structured compiler diagnostics. The root integration
-gates are:
+A fresh checkout needs the MoonBit toolchain and `just`. Recipes use Windows
+PowerShell on Windows and the default POSIX shell on macOS/Linux. The root
+integration gates are:
 
 ```sh
 just check              # native + JS workspace checks and formatting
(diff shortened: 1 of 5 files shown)


tonyfettes:
Tested by @kokic on Windows machine

PR #1438, opened 2026-09-10
fix(mbtx): simplify outcome briefs at their source

bobzhang:
mbtx briefs repeated failure and exit status already conveyed by the desktop status icon. Generate concise captions in the tool itself: ordinary results omit the optional brief and render as `mbtx`; staged failures identify only the phase (`mbtx build` or `mbtx run`); background handoffs retain their job IDs. Detailed diagnostics, timeouts, and exit codes remain in the output.

Desktop renders the supplied brief directly. Both Desktop and the session viewer recognize the explicit phase captions, preserving failure colors and the build-error filter while removing prefix stripping and target inference. The viewer keeps briefs raw for classification and adds status markers only when rendering. Recorded briefs remain displayed as recorded.

Validation:
- `just check`, `just test`, and `just build` passed: 3,274 Native tests, 3,235 JS tests, 24 Cram cases, and CLI lifecycle checks.
- Updated mbtx producer assertions and documentation snapshots; 23 transcript component tests passed.
- Focused Playwright status/background tests (3) and session-viewer filter tests (3) passed, including call-row styling and visibility in both Raw log and Model view.
- `moon info && moon fmt`; no public interface changes.

A previous broader browser run timed out in the Codex partial-output case at a disabled Send button before exercising tool rendering.


desktop/frontend/transcript/component/view.mbt
```diff
@@ -294,7 +294,13 @@ fn tool_request_view(
     Pending(..) => (false, None)
   }
   let label = if brief is Some(brief) && !brief.is_blank() {
-    brief
+    // The trailing status icon already conveys mbtx's outcome. Keep other
+    // briefs, including background handoffs with their job id.
+    if name == "mbtx" && brief.has_prefix("mbtx (") {
```
bobzhang: is this hackish， why we genreate "mbtx （" still？

PR #1439, opened 2026-09-10
fix(desktop): restore Codex app-server on Windows

tonyfettes:
Windows npm installations expose `codex.cmd`, which native process spawning cannot resolve as an executable. Launch the default command through `cmd.exe /d /c`, and follow MCP stdio's task-owned pipe lifecycle so shutdown cancels pending IO before closing its handles. Preserve startup and handshake errors in the desktop status and logs.

Opening a registered project could also fail with `invalid Codex draft project` because frontend URI paths and native Windows paths differed in separators and drive-letter case. Resolve projects through the registry and use the owning host's path rules for workspace matching. The host converts native Codex cwd/project roots to Desktop resource paths in draft and thread replies, then resolves incoming paths back to native cwd before calling app-server. This uses the same convention as worktree replies and broadcasts; the frontend no longer detects Windows paths or performs platform-specific conversion. The shared host path encoder emits lowercase Windows drive letters to match frontend URI serialization, so sidebar grouping recognizes the registered workspace.

Validation:
- Codex frontend JS tests: 72 passed, including Windows project sidebar grouping.
- Native path, API and Codex tests: 47 passed, including thread catalog/reply resource-path conversion and workspace grant roundtrips.
- Host workspace authorization regression: passed.
- Live npm-installed Codex 0.140.0: initialize/initialized handshake and shutdown passed.
- Native desktop and JS frontend builds passed using a separate build directory.

Full repository gates remain limited by existing Windows issues: `just check` reports unused declarations in unchanged bgjobs/worktree code, and `just test` hangs in unrelated native tests. A repeat of `just build` hit a locked Windows packaging executable; the separate desktop/frontend builds above passed.

### DIFF
--- a/desktop/internal/api/api.mbt
+++ b/desktop/internal/api/api.mbt
@@ -813,10 +813,9 @@ async fn ApiState::codex_file_search(
   self : ApiState,
   payload : @protocol.CodexFileSearchPayload,
 ) -> @protocol.CodexFileSearchReply {
-  guard @work_dir.resolve_in_workspace(
+  guard @work_dir.resolve_host_workspace_authorization(
       payload.thread_id,
-      "",
-      workspace=payload.cwd,
+      payload.cwd,
     )
     is Some(root) else {
     raise @host.HostError("Codex workspace is not authorized")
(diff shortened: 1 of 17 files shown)


desktop/internal/api/api.mbt
```diff
@@ -234,8 +241,14 @@ fn CodexWorkspaceBinding::decode_reply(value : Json) -> CodexWorkspaceBinding? {
 /// Terminal request caused by that render can resolve the cwd immediately.
 async fn CodexWorkspaceBinding::authorize(self : CodexWorkspaceBinding) -> Unit {
   match self.workspace {
-    Some(workspace) if !workspace.trim().is_empty() =>
-      @work_dir.authorize_host_workspace(self.session, workspace)
+    Some(workspace) => {
+      guard @pathx.Absolute::from_uri_path(workspace) is Some(path) else {
+        raise @host.HostError(
+          "invalid Codex workspace resource path: " + workspace,
```
tonyfettes, line 247: use string interpolation

PR #1445, opened 2026-09-10
refactor: split fields with split_once/lexmatch instead of offset slicing

bobzhang:
## What

`parse_goal_baseline` located its `dirty` field with `find` and then cut
the sha out as `rest[:cut]` / `rest[cut + 7:]` — that `+ 7` being the
width of the separator it had just searched for. The same
find-then-slice-by-offset shape, with its hand-rolled separator widths
and `view(start_offset=...)` arithmetic, appeared in a dozen more parsers
across `agent_session`, `mcp`, `desktop` and `editor`.

Each site now calls the library function that already means "split here":

- `agent_session/goal.mbt` — `find(" dirty=")` plus a `match` on
  `"true"`/`"false"` becomes an anchored `lexmatch` whose arms carry the
  sha as the unmatched `before=` prefix.
- `editor/server/host/{language_provider,moon_check}.mbt` — four
  `line:col` parsers use `split_once(":")` / `split_once("-")`.
- `desktop/internal/host/git_ops.mbt` and
  `desktop/internal/worktree/submodules.mbt` — tab-separated git records
  drop `view(start_offset=tab + 1)` for `split_once("\t")`.
- `desktop/internal/skillmarket/{library,registry}.mbt`,
  `desktop/frontend/{skills/document,interop/query}.mbt`,
  `desktop/internal/auth/pkce.mbt`, `desktop/internal/update/update.mbt` —
  `:`, `=`, `?` and `-` field splits.
- `mcp/streamhttp/transport.mbt` — the two-hop URL split
  (`find("://")`, `scheme_end + 3`, `host_start + slash`) rebuilds the
  origin as `"{scheme}://{host}"`, byte-identical to the slice it
  replaced.
- `editor/internal/viewer/markdown/sections.mbt` and
  `.../browser/markdown_document/markdown_document.mbt` — `+ 1` / `+ 2`
  separator cuts.

## Behavior

Unchanged. The goal-baseline `lexmatch` keeps the strict shape its doc
comment promises: the sha is the unmatched prefix and `$` keeps the value
the last field, so a line with trailing text is still not a baseline.
`agent_session/goal_test.mbt` pins that case, which the old code got from
requiring the value to be exactly `true`/`false`.

## Verification

- `moon check --deny-warn` — clean.
- `moon fmt --check` — clean.
- `moon test` for every touched package — `agent_session`,
  `mcp/streamhttp`, `editor/server/host`,
  `editor/internal/viewer/markdown`,
  `editor/internal/viewer/browser/markdown_document` (144 tests) and
  `desktop/internal/{host,worktree,skillmarket,auth,update}`,
  `desktop/frontend/{skills,interop}` (184 tests) — all pass.

Left alone deliberately: `editor/base/common/extpath.mbt` (a scan loop
ported from Monaco), `split_tokenized_code_lines` (its whole point is
keeping the trailing empty item), `desktop/internal/shellenv` (`find` plus
`rev_find` for text between the FIRST and LAST marker) and `agent_tool/edit`
(offsets needed for error reporting).

Generated with SeekMoon

### DIFF
--- a/agent_session/goal.mbt
+++ b/agent_session/goal.mbt
@@ -221,7 +221,9 @@ pub fn parse_goal_baseline(content : String) -> GoalBaseline? {
     (re" dirty=false$", before=sha, after=_) => (sha, false)
     _ => return None
   }
-  if sha.is_empty() {
+  // The FIRST ` dirty=` is the field separator, so a prefix that still holds
+  // one means the line repeated the field — mangled, not a sha containing it.
+  if sha.is_empty() || sha.contains(" dirty=") {
     return None
   }
   Some({ sha: sha.to_owned(), dirty, })
--- a/agent_session/goal.mbt
+++ b/agent_session/goal.mbt
@@ -213,17 +213,18 @@ pub fn parse_goal_baseline(content : String) -> GoalBaseline? {
   guard content.strip_prefix("\{GoalBaselinePrefix} sha=") is Some(rest) else {
     return None
   }
-  guard rest.find(" dirty=") is Some(cut) else { return None }
-  let sha = rest[:cut].to_owned()
-  let dirty = match rest[cut + 7:].to_owned() {
-    "true" => true
-    "false" => false
+  // Either value literal ends the line, and the sha is the unmatched prefix:
+  // the `$` keeps the value the LAST field, so a line with trailing junk
+  // degrades to a plain notice instead of a half-parsed baseline.
+  let (sha, dirty) = lexmatch rest {
+    (re" dirty=true$", before=sha, after=_) => (sha, true)
+    (re" dirty=false$", before=sha, after=_) => (sha, false)
     _ => return None
   }
   if sha.is_empty() {
     return None
   }
-  Some({ sha, dirty, })
+  Some({ sha: sha.to_owned(), dirty, })
 }
 
 ///|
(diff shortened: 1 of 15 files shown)


bobzhang:
### `codex review --base origin/main` — sign-off

Reviewing this branch twice:

**First pass** (`4d146f339`) raised one regression:

> **[P3] Preserve rejection of repeated dirty fields** — `agent_session/goal.mbt:219-221`
> For a malformed notice such as `[goal baseline] sha=abc dirty=true dirty=false`, the suffix search now returns `sha="abc dirty=true"` and `dirty=false`; previously, parsing returned `None`. This violates the parser's strict-shape contract and causes `current_goal_baseline()` to pass an invalid commit and misleading cleanliness flag into the audit prompt.

Addressed in `a90f54fe9`: the FIRST ` dirty=` is still the field separator,
so a captured prefix that holds another one is rejected, and the case is
pinned in `agent_session/goal_test.mbt` next to the trailing-text case.
The review's own suggestion — "explicitly reject another delimiter in the
captured prefix" — is what the fix does.

**Second pass** (`a90f54fe9`):

> The parsing refactors preserve existing behavior, including rejection of malformed goal baselines. `just check` and targeted tests passed; socket-dependent validation was limited by sandbox restrictions.

The caveat is the reviewer's own: its sandbox cannot bind a local socket,
so socket-dependent tests were skipped there. CI runs them (see the checks
on this pull request).

PR #1454, opened 2026-09-11
fix(multi_edit): explain a rising error count and let the agent admit the batch

### LINKED ISSUE #1452: multi_edit revert guardrail counts pre-existing errors as newly introduced, rolling back correct edits
## Summary

`multi_edit`'s auto-revert guardrail decides "did this batch introduce errors?" by subtracting two whole-project error **counts**:

```moonbit
// agent_tool/multi_edit/multi_edit.mbt:573-593
let after = count_errors_or_none(check_path)        // batch applied
...
write_files(writes, w => w.original_content, ...)   // roll the originals back
let before = count_errors_or_none(check_path)       // "pre-batch"
let introduced = after.error_count - before_count
if introduced > threshold { /* stay reverted */ }
```

That would be sound if the count were "the number of errors in the project". It is not: a `moon check` run reports only the part of the project it manages to compile. The build stops at the first failing package level, so the count is a **lower bound that moves as the build gets further**. Fixing errors therefore *raises* the reported number, by un-hiding the errors that the earlier run never reached.

## Reproduction of the bogus delta

Tree: a worktree holding a partly applied mechanical refactor (many sites where `.to_owned()` was deleted, roughly 97 real errors in total). Verified measurements on it:

- `moon check --output-json` -> reports exactly **10** errors (the first failing cluster: `cmd/openseek/*` 9 plus `desktop/internal/engine/ops.mbt` 1).
- `moon check --output-json --diagnostic-limit 500` -> still **10**, so this is not the diagnostic limit; the build simply stops there.
- `moon check --output-json --target all` -> **50** lines / 37 unique sites (another partial view).

Now apply a batch whose only effect is to add `.to_owned()` **back** on 10 lines of `cmd/openseek/{approval,baseline,concurrent,main}.mbt` and `desktop/internal/engine/ops.mbt`. Those edits take values from `StringView` to the *wider* `String`; they cannot create "needs a `String`" errors. Result:

```
reverted: applied 10 edit(s) across 5 file(s), but moon check then reported 87 error(s)
(77 newly introduced > revert_when_errors_above=10); all 5 file(s) were rolled back ...
```

Arithmetic of what actually happened:

```
before:  reports 10   (build stopped early; about 87 more errors never reached)
after:   reports 87   (first 10 fixed -> build goes further -> next errors surface)
actual:  about 97 errors -> about 87 errors   (10 fixed, none introduced)
tool:    87 - 10 = 77 -> "77 newly introduced" -> roll back
```

Two independent arguments that none of the 87 were introduced by the batch:

1. **Direction.** Every listed error is `value has type StringView, wanted String`. The batch only *adds* `String`-ness, so it cannot cause that class of error. (A restore that broke something would show the opposite, e.g. `String wanted StringView`; none do.)
2. **Locality, with its own cause.** The listed errors are all in `desktop/internal/host/*` (`browse_ops.mbt:49,:58,:139`, `file_search.mbt:299,:358`, `fs_watch.mbt:102`, `git_ops.mbt:115,:318,:498,:571`), whose own `.to_owned()` sites are removed in that tree: for example `browse_ops.mbt:49` fails because `requested` on line 36 of the *same file* is the swept one. They are not downstream of the edited packages.

## Impact

This makes a large mechanical sweep unfixable once the tree holds more errors than the first failing cluster: every *corrective* batch is reported as introducing the errors it is fixing, so it is rolled back and the error count never comes down. The standard workflow for a project-wide refactor - delete a redundant call everywhere, then let the compiler name the sites to restore - becomes impossible. It also silently discards correct work, and the "reverted" report actively misleads the model about what it broke.

## Related observations while reading

1. `check_path = first.path` (`multi_edit.mbt:381`): the check's cwd/module comes from whichever file is first in the batch (`nearest_project_dir`, `agent_tool/internal/auto_check/auto_check.mbt:35`). A batch mixing a file with its own `moon.mod` (e.g. `scripts/`) with main-workspace files measures a different project entirely.
2. `if after.error_count <= threshold { keep }` (`multi_edit.mbt:577`): the fast path compares an *absolute* post-batch count, not a delta. In a tree that already has more than `threshold` errors every batch takes the expensive restore, re-check and re-apply path, and that path is the unsound one above.
3. The partial view is reachable from the CLI too, which is why an out-of-band enumeration cannot dodge this: the same tree measured 10 (default/native), 40 (`--target js`), 50 (`--target all`), and 87 once the build got further.

## Suggested direction

- Judge the batch by errors **located where the batch edited** (its own changed lines/files), not by a whole-project count. An over-match shows up as errors inside the edited spans; pre-existing failures elsewhere cannot be attributed to the batch.
- If counts must be used, make both runs comparable: compare error *identities* (`path:line:code`) within the same compiled window, and treat "the count went down" as a keep.
- `moon check` could expose whether the build stopped early (a partial result), so callers never treat a truncated count as a total; `CheckErrors::truncated` already hints at that distinction internally.

bobzhang:
Fixes #1452.

## Problem

`moon check` skips the dependents of a failing package, so in a tree that already has errors a batch that **fixes** them lets the build reach further and report errors that were always there. The revert guard subtracted two such counts, read that as "N newly introduced", rolled the correct batch back, and told the model to narrow it. For a large mechanical sweep that is a dead loop: every corrective batch is reported as introducing the errors it fixes.

## Approach (no new option)

The guard still decides by count, but the revert report now classifies every new error with moon's own package graph (`moon tree --package --json`, ~0.2 s on this workspace) and the existing `revert_when_errors_above` is the explicit override:

| verdict | evidence | advice |
|---|---|---|
| **over-match** | new error inside an edited file | narrow `old_string`s / split the batch |
| **breakage** | new error in code that compiled cleanly before and depends on (or shares a package with) the edit | fix the batch (changed signature or type?) |
| **certified reach** | every new error is in a package that depends on nothing edited | re-issue unchanged with the exact `revert_when_errors_above` named |
| **plausible reach** | new errors in packages that depend on a package the batch *fixed* (moon had skipped them) | the sites to vet are listed; re-issue with the named value only if they are the pre-existing class |
| **inconclusive** | no baseline / clean baseline / truncated output / no graph / an edited file outside the checked project / a module-manifest edit / an unplaceable error | stays reverted, narrow or split |

Facts verified along the way: moon hides only dependents of a failing package (independent packages are still reported, even with `-j 1`); in this workspace `desktop/internal/host` transitively imports `desktop/internal/engine`, so the issue's own case is a plausible reach, not a certifiable one.

Other changes:

- `auto_check` keeps every error site (`path`, `loc`, `error_code`, `message`); identity for comparing runs is `path:loc#code` when moon supplies a code. New `package_graph.mbt`: fetch, decode, module `source` roots (with lexical normalization), longest-prefix package lookup, transitive dependency.
- Attribution handles: canonical paths (`/tmp` vs `/private/tmp`), edited symlinks (link name for diagnostics, compiled target's package for dependencies), a fixed package's own other files (plausible, its tests were skipped), legacy `moon.pkg.json`/`moon.mod.json`, and non-source files a `dev_build` rule names as inputs.
- `revert_when_errors_above` default 5 → 10, host cap 10 → 200; schema text and the **default prompt** (the one every supported model receives) describe the verdicts and say to raise the value only to what the report names.
- The unused base prompt (`base_prompt.mbt.md`, never selected) is removed in its own commit; drop it if you want it split out.

## Tests

- Graph decode / source roots / normalization / placement / dependency; every verdict, including the mixed-module and partial-fix cases from the review; symlink target, build-input, fixed-package-sibling, error-code identity, path matching.
- End-to-end through real `moon check` and `moon tree` on a two-package fixture where the edited file is a symlink under `/tmp`: fixing `a` surfaces twelve errors in `b` (1 → 12, "11 introduced", none real) → plausible-reach report with the sites and the value; re-issue kept; a 25-error break of the edited file → over-match, reverted. Passes on native and wasm.
- `moon check --deny-warn` clean on native, js and wasm; 91 tests across `auto_check` and `multi_edit`.

Not addressed here (from the issue): the `check_path = first.path` module choice for mixed-module batches (now made safe: inconclusive, never reach) and the absolute-count fast path; and moon marking an incomplete build, which would let the tool certify the plausible class too.

https://claude.ai/code/session_01F9pUVVTAwr6QykqEf9sFwg


bobzhang:
Reviewed the diff against the #1452 repro. **Verdict: yes, this addresses the dead loop we hit**, and it does so where it matters: the guard stays exactly as strict when `before_count == 0` (the only case where a count comparison is sound), and for `before > 0` it now returns the evidence plus an exact override. In the session that produced the issue, `before` reported 10, the batch fixed all 10, `after` reported 87 — your `comparability recognises the build reaching further after a fix` test encodes exactly those numbers. With this patch the report would have said "re-issue it UNCHANGED with revert_when_errors_above=77", the batch would have been admitted, and the loop would have ended. Keeping `revert_when_errors_above` as the explicit override (no new flag) is the right call, and the e2e test proving the second call is *kept* is the part I most wanted to see.

Residual risks I'd want addressed or explicitly accepted before merge:

**1. `looks_like_reach()` cannot see the one breakage shape the guard exists for: a signature/type change that breaks only unedited callers.**
You document this honestly ("a batch that changes a signature can leave every old error gone and break only unedited callers"), and the advice asks the model to check for it. But that turns the safety net into a prompt-level promise in exactly the case where the model is most confident and most wrong — and the new advice *actively instructs* it to override the guard with a large threshold (up to 200). Concrete way to make it mechanical: use the package graph. A new error in a package that transitively imports an edited package *can* be caused by the batch; a new error in a package that does not, cannot. In our case `desktop/internal/host` does not depend on `cmd/openseek` or `desktop/internal/engine`, so reach still classifies correctly — while a signature-change over-match would be caught, because its callers are dependents. The repo already has graph machinery (`desktop/internal/host/package_graph_ops.mbt`); even a cheap `import { ... }` parse of the `moon.pkg` files would do. I'd much rather have that than the prompt sentence.

**2. Reach can be concluded from a comparison that never saw the edited files.**
`in_edited_file` compares moon-reported paths against the canonical write targets. If the check ran in a different module than the batch edited — which `check_path = first.path` permits, and which this PR lists as follow-up — then no diagnostic can ever match an edited path, so `new_in_edited` is empty for a reason that has nothing to do with the batch, and `looks_like_reach()` will cheerfully conclude "reach". Minimum guard: record the cwd the check ran in (`moon_check_cwd` already computes it) and require every edited path to live inside it; otherwise mark the comparison inconclusive and never claim reach. That would make the `check_path = first.path` follow-up a correctness prerequisite of this change rather than a nice-to-have.

**3. `remaining == 0` is too strict for the advice it gates.**
A batch that fixes most but not all of the errors the pre-batch run stopped at — very common in a sweep — fails `looks_like_reach()` and falls through to the unconditional "re-issue with more precise old_strings …", telling the model to narrow a batch that may be perfectly fine. The evidence line already prints "N of M pre-batch error(s) remain"; I'd make the closing advice conditional on the other two signals (`new_in_edited.is_empty() && unlocated == 0`) instead of dropping straight back to the old wording.

**4. The cap going 10 → 200 changes the property `definition()` advertises.**
The schema text still says the model "cannot raise it past this cap, so an overconfident call cannot silently disable the safety net". With the cap at 200 and the report handing over the exact value, the net is in practice opt-out for deltas up to 200 — which is defensible (I would have needed it), but it is only safe if the admit/deny signal is mechanical (see #1). If the prompt plea stays, consider at least bounding the auto-suggested value, or requiring positive evidence that the pre-batch run was at its stop point rather than merely the absence of a negative signal.

**5. Identity is the message text; `error_code` would be sturdier.**
`ErrorSite::line()` (`path:loc: message`) is used as the identity behind `remaining` and `new_in_edited`. A reworded diagnostic (moon release, or a message that embeds a count) turns a still-present pre-batch error into a "new" one; if it also sits outside the edited files, it counts toward the reach evidence. The JSON already carries `error_code` — capturing it in `ErrorSite` and using `(path, loc, error_code)` for identity, with `message` kept for display only, would be more stable.

**Tests.** The e2e fixture is the right shape: a real two-package `moon check` where fixing `a` surfaces `b`, a negative control (25 errors inside the edited file → over-match, never reach), and the symlink//`private/tmp` canonical-path case. Unit coverage for the clean, missing, truncated and unlocated baselines is good. Two gaps I'd add: (a) a mixed-module batch (edits in a module the check did not run in) asserting the report does *not* claim reach — finding #2; (b) a partial fix (some pre-batch errors remain, new errors elsewhere) asserting the advice is not the unconditional narrow-your-old_strings — finding #3.

**Nit.** `truncated` means "capture budget overflow", while `looks_like_reach`'s doc talks about the build reaching further; the early-stop case is not representable in `CheckErrors` today. A comment saying so — and, longer term, asking `moon check` to mark an incomplete build — would stop a future reader from assuming the tool can detect early stop. That is the third point from #1452 and is still open.

bobzhang:
Thanks — all five addressed in 478d39f0d, with two findings that change the picture a little.

**Two facts I verified first.**

- moon hides only the *dependents* of a failing package. On a four-package fixture (`b` imports broken `a`; `c` is independent and has its own errors), `moon check` reports `a` and `c` and skips `b` — also with `-j 1`. So reach through an independent package should not happen from dependency order.
- In this workspace, `moon tree --package --json` shows `desktop/internal/host` transitively importing `desktop/internal/engine`, and `engine/ops.mbt` held one of the ten pre-batch errors in the issue's session. So the issue's own case was a reach *through a dependent of a fixed package* — exactly the class the graph cannot certify.

**1. Mechanical attribution — done with the graph, honestly scoped.** The guard now runs `moon tree --package --json` (0.2 s here, 700 nodes / 3 300 edges) and sorts every new error into:

- *over-match*: inside an edited file;
- *breakage*: in another file of an edited package, or in a package that depends on an edited package which had **no** pre-batch errors — that code compiled cleanly before, so only the batch can have changed it. This is the signature-change case, caught mechanically whenever the changed package was clean before;
- *certified reach*: in a package independent of everything edited — cannot be caused by the batch. The report names the value with no caveat;
- *plausible reach*: in a package that depends on a package the batch **fixed**. moon had skipped that package while its dependency was broken, so the error is either surfacing or a caller broken by the fix, and no single check can tell (the package's pre-batch state against a broken dependency does not exist). The report lists those exact sites and asks the model to vet them before re-issuing with the named value.

So the prompt plea is gone for every class except the one where it is genuinely the only signal left, and that class is named site by site rather than waved at.

**2. Check ran elsewhere — done.** Every edited file must map to a package of the checked project; otherwise the verdict is *inconclusive* and no reach is claimed. Test added.

**3. `remaining == 0` — done.** A partial fix stays on the reach path (“4 of 10 pre-batch error(s) remain (the batch fixed the rest)”); the narrow-your-old_strings advice is now given only for over-match and inconclusive verdicts. Test added.

**4. Cap and schema text — done.** The description now describes the four verdicts and says to raise the value only to what the `comparability:` line names; the default prompt says the same. The suggested value is only ever emitted for the two reach verdicts.

**5. Identity — done.** `ErrorSite` carries `error_code`; identity is `path:loc#code` when present, message text only as a fallback and for display.

**Nit — done.** `CheckErrors` now documents that early build stop is not representable there and has to be reasoned about from the graph.

Still open (from the issue, not this PR): `check_path = first.path` module choice for mixed-module batches — now made safe by point 2 (inconclusive), but not resolved; the absolute-count fast path; and moon marking an incomplete build, which would let the tool certify the plausible class too.

PR #1459, opened 2026-09-11
refactor: drop 97 redundant view-to-owned conversions

bobzhang:
## What

Removes 97 redundant view-to-owned conversions - `.to_owned()` on a `StringView`, plus a couple of `BytesView`/`ArrayView` copies - across 64 files. No file gains one.

## Why

`StringView`/`BytesView`/`ArrayView` share their API with the owned types, so most `.to_owned()` calls at a use site only allocate a copy at hand: the value is read locally, or handed to something that already takes a view. Examples from the diff:

- `@json.parse(line.to_owned())` -> `@json.parse(line)` (`parse` takes a `StringView`)
- `(first.to_owned(), rest.to_owned())` -> `(first, rest)`
- `retained.write_bytes(chunk[:length].to_owned())` -> `retained.write_bytes(chunk[:length])`

## How it was verified

Not a text-level replace. Each candidate site's call was deleted, the compiler was asked what broke, and every site where an owned value is genuinely required (a struct field, a `String?` return, an `Array[String]` push, a `String`-taking callee, ...) was restored. Roughly 40% of the candidates needed their `.to_owned()` back.

- `moon check --target native|js|wasm|all --deny-warn` - clean
- `moon fmt --check` - clean (the formatter collapsed a few now-shorter calls)
- `moon info` - no interface change
- `moon test --target native` - 3071/3074; the three failures were two environment-sensitive snapshots in `cmd/openseek` (`environment_prompt_section` appends `OPENSEEK_REFERENCES` when set, and this sandbox sets it; they pass with the variable emptied) and one inline snapshot in `editor/viewer/common/languages/token_pipeline_test.mbt`, which is why that file keeps its two conversions.

## Notes for review

- Only `.to_owned()` calls were touched. A per-file check confirms none was added.
- A few leftovers of the mechanical sweep were tidied in the same commit: `map(field => field)` lost its identity `map`, and three `let x = x` shadows left behind by removing the copy were deleted.
- Behaviour is unchanged: a view borrows the same backing value, so the only difference is the copy that is no longer allocated (where a view is stored, the parent stays alive).

Generated with SeekMoon

### DIFF
--- a/agent_session/store/store.mbt
+++ b/agent_session/store/store.mbt
@@ -195,7 +195,7 @@ fn parse_session_file(
   text : String,
 ) -> ParsedSessionFile raise {
   let (first, rest) = match text.split_once("\n") {
-    Some((first, rest)) => (first.to_owned(), rest.to_owned())
+    Some((first, rest)) => (first, rest)
     None => (text, "")
   }
   let header_json = @json.parse(first) catch {
@@ -221,7 +221,7 @@ fn parse_session_file(
     )
   }
   let ends_with_newline = text.has_suffix("\n")
-  let segments = rest.split("\n").map(line => line.to_owned()).collect()
+  let segments = rest.split("\n").collect()
   let decoded : Array[@agent_session.SessionEvent] = []
   for index, line in segments {
     if line.is_blank() {
--- a/agent_subtask/lifecycle_test.mbt
+++ b/agent_subtask/lifecycle_test.mbt
@@ -448,7 +448,7 @@ async test "capture is re-entrant: resumed work stacks on the same branch" {
     }
     assert_true(commit1 != commit2)
     match @agent_subtask.git_require(["rev-parse", "\{commit2}^"], cwd=repo) {
-      Ok(parent) => assert_eq(parent.trim().to_owned(), commit1)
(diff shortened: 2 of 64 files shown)


bobzhang:
## Codex CLI review - clean

Reviewed commit `935132d23` against the branch base `b856e7228` with `codex review --base <base>` (Codex CLI v0.153.4, model `gpt-6-astra`, reasoning effort high, session `01a08fd2-fcdd-7d20-9250-7b0df595d8ec`). The head has not changed since.

> No actionable regressions found in the ownership-conversion cleanup. Native and JS checks passed, as did 623 targeted JS tests; native tests reported 352 passes and one failure caused by sandbox restrictions on TCP binding.

The single failure it hit was its own sandbox refusing a loopback `bind()`, not a test failure.

For completeness: CI is green on this commit (13/13 - `check` on native/js/wasm, `browser-correctness`, Editor `moon` on stable/nightly x macOS/Ubuntu, `desktop e2e`, `desktop package (Windows app)`, `viz e2e`, GitGuardian), and `Devin Review` reported "pass" but with *full review skipped (trial expired)*, so it is not an independent review - this Codex run is.

Sign-off: no changes requested.

Generated with SeekMoon

PR #1464, opened 2026-09-11
refactor(agent_subrun): remove the unused SubrunBudget

bobzhang:
## What

`moon ide analyze agent_subrun` reports `SubrunBudget` — the constructor and
`reserve`/`reset_turn` — with `usage: 0 (0 in test)`: no consumer anywhere,
not even tests. Per `shrink_package.md`, a zero-usage block with no live
owner is vestigial, so delete it:

- remove `agent_subrun/budget.mbt` (struct, methods, and its four tests)
- drop the README bullet that documented it
- regenerate `agent_subrun/pkg.generated.mbti` via `moon info`

The README previously described it as a reusable per-turn call allowance for
direct consumers, but the CLI's model-initiated delegation goes through
hosted workflows and their reserved child blocks, and the goal-met gate has
its own allowance — nothing withdrew from the budget.

## Verification

- `moon check -d --target native` (workspace): clean
- `moon test agent_subrun --target native`: 19 passed / 0 failed (was 23;
  the four removed tests were budget's own)
- `moon fmt --check agent_subrun`: clean
- `moon ide analyze agent_subrun`: `SubrunBudget` no longer appears
- `.mbti` diff: the `SubrunBudget` type and its three methods are gone

## Caveat

This removes a public API from a published module, so it is a breaking change
for any external consumer that used `SubrunBudget`. In-repo and analyzer usage
is the yardstick applied here.

Generated with SeekMoon


bobzhang:
Independent review (bundled review workflow, subrun): **1 finding, 0 blockers.**

Finding (medium): `docs/plans/subtask-worker-subagents.md` still described
`SubrunBudget` as live design — line 60's "SubrunBudget slot" and the
"Unchanged from v2 (still in force)" bullet — so the "no dangling references"
criterion was not fully met.

Fixed in f46f38f93: the backstop is now noted as removed in #1464 and the
reservation dropped from the precondition list; `rg SubrunBudget` over the repo
now returns nothing.

Everything else verified clean: the deleted API had no consumer (checked
against the parent tree, since the analyzer can no longer re-derive usage for a
deleted symbol), the README matches the remaining code, the regenerated `.mbti`
is byte-identical, `moon check -d --target native` exit 0, `moon test agent_subrun`
19/19 with no skips, `moon fmt --check agent_subrun` clean, and the diff is
exactly the intended three files plus this doc fix.

PR #1465, opened 2026-09-11
Plan durable background job logs and Desktop visibility

bobzhang:
Background jobs previously exposed lifecycle mainly through prose, created output files only above an inline limit, and removed those files at session teardown. This plan defines files available before background handoff, durable lifecycle metadata and structured control, bounded host reads, and a Desktop Jobs panel.

The six-part implementation is complete. The document preserves the reviewed proposal and records final scope decisions and limits. Independent `openseek review` found no blocking/design issue; three citation nits were corrected. This slice changes no executable behavior; its GitHub checks pass.

Stack: #1465 → #1466 → #1467 → #1468 → #1469 → #1470. Review and merge in this order; aggregate validation and review disposition are recorded in #1470.

### DIFF
--- a/docs/plans/background-job-observability.md
+++ b/docs/plans/background-job-observability.md
@@ -14,14 +14,14 @@ Durable logs survive engine shutdown. Processes keep their current session-owned
 
 | Area | Current behavior | Consequence |
 | --- | --- | --- |
-| `agent_tool/shell_exec/sink.mbt:100` | Memory-first sink creates a spill file only after the inline cap; seeds it with prior output. | A quiet or short job has no file to tail. The sink default is 12,000 characters, but `mbtx` passes 48,000. |
-| `agent/tool_definition.mbt:69` | Creates `openseek-jobs-*` in a temporary directory; removes the directory after the session task group ends. Creation failure silently selects memory-only mode. | Logs cannot serve as durable user artifacts. Snippet scratch space and output retention currently share ownership. |
-| `agent_tool/shell_exec/sink.mbt:100` | Swallows write failures and still advances `file_bytes`; read failures fall back to the head. | A log can be incomplete while reads look successful. This needs correction before advertising a reliable output file. |
+| `agent_tool/shell_exec/sink.mbt:121` | Memory-first sink creates a spill file only after the inline cap; seeds it with prior output. | A quiet or short job has no file to tail. The sink default is 12,000 characters, but `mbtx` passes 48,000. |
+| `agent/tool_definition.mbt:69–90` | Creates `openseek-jobs-*` in a temporary directory; removes the directory after the session task group ends. Creation failure silently selects memory-only mode. | Logs cannot serve as durable user artifacts. Snippet scratch space and output retention currently share ownership. |
+| `agent_tool/shell_exec/sink.mbt:128` | Swallows write failures and still advances `file_bytes`; read failures fall back to the head. | A log can be incomplete while reads look successful. This needs correction before advertising a reliable output file. |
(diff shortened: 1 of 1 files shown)


bobzhang:
## Architecture review of this plan, read against the implementation in #1466 → #1468 → #1469

Verdict: **the architecture is solid and the stack should proceed.** This is a better-than-usual plan — it does the two things that decide whether an observability feature survives contact with reality, and it does them before any code:

- It names the **identity tuple** `(store, session, runtime_generation, job_id)` and explicitly forbids deriving a log path from a displayed `bg-N`. That single decision is why a restarted engine can't overwrite an old log and why two sessions with a `bg-1` stay separate. The implementation honours it end to end (`JobRecord.generation`, `valid_job_component`, `WrongRuntime`).
- It separates **committed** from **observed** bytes and insists on honest I/O failure instead of the previous swallow-and-advance. `sink.mbt` now refuses to advance `file_bytes` after a failed write and propagates the failure into a `capture_failed` terminal state. That is the right invariant and it is the one most implementations get wrong.

Other choices that hold up: scratch/log ownership split (`@fs.tmpdir` for snippets, session store for logs), relative artifact references resolved at serve time so archive/unarchive works, `owner.lock` as the ownership primitive rather than a PID, and byte cursors with explicit UTF-8 boundary handling. The phased gates were real gates — the integration fixture in `turn_finish.py` and the native host tests in `jobs_wbtest.mbt` actually test the things the gates named.

### Four places where the plan itself has gaps, each of which produced a finding downstream

1. **`stop` has no time bound and no escalation.** The plan says "keep stopping asynchronous relative to command intake" and "emit its final result when the tracked execution settles", but never says how long settling may take or what to do when it doesn't. It also states plainly that cancellation reaches only the direct child. The implementation took this literally: the engine emits `JobStopResult` only after the process is terminal, and #1468 wraps that reply in a 5 s timeout. A process that ignores SIGTERM therefore reports a *failed* stop while the stop is in fact still pending, with the 30-minute reaper as the only backstop. The contract needed a typed third outcome (`Stopping`/`StopTimedOut`) or a SIGTERM→SIGKILL escalation; details in #1468.

2. **Revision-based merging was specified, then quietly replaced.** The plan asks for "merge the snapshot with incoming events by identity and revision, so a late snapshot cannot roll back an exit". The implementation instead makes disk authoritative and re-lists on every `job_changed` plus a 1 Hz timer; the event's `job` payload is discarded and `JobRecord.revision` has no consumer anywhere in `desktop/`. The simpler scheme is defensible — but then `revision` and the event payload are dead protocol surface, and the freshness ordering now depends on poll timing rather than a stated rule. Pick one: use the revision, or drop it and shrink the event to an identity ping.

3. **The generation directory has no lifecycle.** "Age/quota cleanup can follow with an explicit policy" covers *log* growth. It doesn't cover the fact that a directory-per-runtime scheme creates a directory unconditionally — `spawn_foreground_retained` calls `prepare_output`, so any engine run that executes a single snippet allocates `jobs/<generation>/owner.lock` even when no job is ever adopted, and nothing ever removes an empty one. #1468 then readdir/stats every one of them on every list. That's a design hole, not an implementation nit.

4. **The headline UX affordance was dropped.** "Provide a persistent `Jobs · 2 running` entry in the conversation UI" is the sentence that makes this feature discoverable; the shipped UI puts the count inside the panel, so a running job is invisible unless the user already knows to open the tab. #1469's doc update acknowledges the change, but it converts the feature from ambient to opt-in. Worth an explicit decision rather than a footnote.

Also: the plan's own numbers drifted in implementation without being restated — 500 ms polling became 1 s, `tail -n 200` became `tail -n 100`. Harmless, but the plan is the spec people will read later.

None of the four is a reason to hold the stack. (1) is a real bug to fix in #1468, (3) wants a four-line prune, (2) and (4) are decisions to record. Per-PR reviews follow on #1466, #1468 and #1469.

PR #1466, opened 2026-09-11
Retain background job logs and durable lifecycle controls

bobzhang:
Background commands now expose a retained output file before their handoff or Started event. Output produced before background adoption is seeded into the same file, and subsequent output remains available for external tailing and later inspection. Small foreground commands remain memory-first.

This PR includes the complete runtime contract: generation-scoped metadata and ownership, structured lifecycle events, correlated snapshot/stop commands outside model turns, and retention with the owning session. Storage failures remain explicit. Failure to create required scratch/log storage stops setup or adoption with an error; production does not silently degrade to memory-only retention. Runtime review fixes and real CLI regressions are included here, so correctness does not depend on a later hardening PR.

Two small Desktop exhaustive-match adaptations accompany the new wire variants so this PR builds on its own. Desktop job access and the Jobs panel follow in #1468 and #1469.

## Follow-up hardening

- Remove an unused generation after foreground-only execution, only when no job was adopted and no unexpected files remain. Missing metadata is never a reason to delete a retained log.
- Snapshot the committed output prefix under the sink mutex and read the file outside the writer lock. Preserve the captured memory head after a failed initial seed while retaining the partial path for inspection.
- Add foreground-only cleanup, failed-metadata retention, concurrent append/read, and failed-seed regressions.

The review's proposed cleanup based only on absent JSON was narrowed to protect logs after metadata failure. The suggested terminal-event race was not established under the existing cooperative scheduling and serialized publication; no speculative ordering rewrite is included.

## Validation

Current-head `47543fb72` [CI](https://github.com/moonbitlang/openseek/actions/runs/34663737012) and [Desktop/Windows](https://github.com/moonbitlang/openseek/actions/runs/34663737043) passed. This PR remains one commit and is mergeable.

- The complete final stack also passed 86 browser E2E scenarios and eight Windows job-host regressions; final aggregate evidence is recorded in #1469.

## Stack and review history

### DIFF
--- a/agent/job_lifecycle_wbtest.mbt
+++ b/agent/job_lifecycle_wbtest.mbt
@@ -1,0 +1,113 @@
+///|
+#cfg(not(platform="windows"))
+async test "job lifecycle publishes ordered durable start and failure with retained UTF8" {
+  let root = @fs.tmpdir(prefix="openseek-job-events-")
+  defer @fs.rmdir(root, recursive=true)
+  let events : Array[(@protocol.JobEventKind, @protocol.JobRecord)] = []
+  @async.with_timeout(5000, () => {
+    @async.with_task_group(group => {
+      let jobs = @bgjobs.BgJobRuntime(
+        AgentTaskScope(group),
+        spill_dir=root,
+        persistent_logs=true,
+        on_job_event=(kind, job) => events.push((kind, job)),
+      )
+      let id = jobs.start(
+        program="/bin/sh",
+        args=["-c", "printf '你好'; exit 7"],
+        command="failure",
+      )
+      ignore(jobs.wait_any([id]))
+      assert_eq(events.length(), 2)
+      assert_true(events[0].0 is Started)
+      assert_true(events[1].0 is Failed)
+      assert_eq(events[1].1.revision, 2)
+      assert_true(events[1].1.state is Exited(7))
+      assert_eq(events[1].1.output_bytes, 6L)
+      assert_true(events[1].1.finished_at_ms is Some(_))
+      let saved : @protocol.JobRecord = @json.from_json(
+        @json.parse(@fs.read_file("\{root}/\{id}.json").text()),
+      )
+      assert_eq(saved, events[1].1)
+      assert_eq(@fs.read_file("\{root}/\{id}.out").text(), "你好")
+      jobs.finalize()
+      assert_eq(events.length(), 2)
+    })
+  })
+}
+
+///|
+#cfg(not(platform="windows"))
+async test "stop emits stopping before one settled terminal event" {
+  let events : Array[(@protocol.JobEventKind, @protocol.JobRecord)] = []
+  @async.with_timeout(5000, () => {
+    @async.with_task_group(group => {
+      let jobs = @bgjobs.BgJobRuntime(AgentTaskScope(group), on_job_event=(
+        kind,
+        job,
+      ) => events.push((kind, job)))
+      let id = jobs.start(program="/bin/sleep", args=["30"], command="sleep")
+      assert_true(jobs.stop(id))
+      assert_eq(events.length(), 3)
+      assert_true(events[0].0 is Started)
+      assert_true(events[1].0 is Updated)
+      assert_true(events[2].0 is Finished)
+      assert_true(events[2].1.state is Stopped("user"))
+      assert_true(jobs.stop(id))
+      assert_eq(events.length(), 3)
+    })
+  })
+}
+
+///|
+#cfg(not(platform="windows"))
+async test "session cleanup persists a terminal record after cancellation" {
+  let root = @fs.tmpdir(prefix="openseek-job-finalize-")
+  defer @fs.rmdir(root, recursive=true)
+  let path = @async.with_timeout(5000, () => {
+    @async.with_task_group(group => {
+      let tools = build_tools(
+        AgentRuntime(),
+        AgentTaskScope(group),
+        job_log_dir=root,
+      )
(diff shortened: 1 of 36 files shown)


bobzhang:
## Review — runtime slice (retained logs, durable records, controls)

Design is sound and matches the plan closely. The things I checked and found correct:

- `open_spill`/`append_locked` no longer swallow I/O errors, and `file_bytes` is not advanced past a failed write — the committed-vs-observed distinction the plan asked for is real.
- Adoption is serialized against the pipe reader through the sink mutex, and `materialize()` seeds the file before `register` publishes, so the `output_file` in a `job_changed(started)` always exists. `turn_finish.py` asserts exactly that, including the pre-adoption UTF-8 prefix.
- `register` publishes `Started` under `protect_from_cancel` before spawning the watcher, so a process that exits during adoption still gets `Started` then one terminal event. Group defers run FILO and `with_task_group` asserts `children.is_empty()` before running them, so `finalize()` cannot race the watcher's `finish_job` and the `scratch` rmdir cannot outrun `finalize`. Ordering is right.
- Lazy `prepare_output` is a good call — it is why an idle restart allocates no generation directory, which the integration test also checks.
- `owner.lock` under a `CreateNew` in an exclusively-created directory is the right ownership primitive.

### 1. Nothing ever prunes an empty generation directory (medium)

`spawn_foreground_retained` calls `prepare_output()`, so **any** engine run that executes a single snippet allocates `<session>/jobs/<generation>/` plus `owner.lock` — even when no job is ever adopted and `discard()` removes the `fg-N.out`. `finalize()` writes metadata but never removes a generation that ended up with no `*.json`.

A conversation reopened N times accumulates N directories containing nothing but a lock file, forever. #1468's `job_history_page` then does `kind()` + `mtime(owner.lock)` + `readdir()` on every one of them, and #1469 calls it once per second while the panel is open. The two multiply.

Suggest: at the end of `finalize()`, under `protect_from_cancel`, `rmdir` the generation directory if it contains no `*.json` (after closing/removing `owner.lock`). Cheap, and it keeps the listing cost proportional to jobs rather than to session age.

### 2. `read_all` now holds the sink mutex across a full-file read (medium)

`read_all` acquires `self.lock`, then opens the log and reads `file_bytes` in one shot. `append_chunk` needs the same lock. So a `job_output` call with an `offset` (which goes through `read_output` → `read_all`) blocks the execution's pipe reader for the entire read+`decode_lossy` of up to `hard_cap` = 20,000,000 characters — ~80 MB worst case. The child's pipe fills and the child stalls.

This is new: before this PR `read_all` touched `self.file` without a lock. The plan's "network slowness must never backpressure the execution's pipe reader" applies to a local reader too.

The lock is only needed for the cursor. Suggest snapshotting `file_bytes` (and the head) under the lock, releasing, then opening and reading outside it — `read_tail` is already bounded and fine either way.

### 3. `stop` can publish `Updated` after the watcher published `Finished` (low)

```
job.requested_stop = true
job.exec.request_stop()
self.publish(id, Updated)
```

`publish` contends on `publish_lock` with the watcher's `finish_job`. If the child dies immediately, `finish_job` can win the lock, emit `Finished` at revision 2, and then this `Updated` lands at revision 3 carrying a terminal state — so `Updated` is the last event for a terminal job, with the highest revision.

Harmless today because #1469 discards event payloads and re-polls, but it breaks the "one terminal event, last" property the `register` comment is careful to establish. Skip the `Updated` publish when the job is already terminal, or publish it before `request_stop()`.

### 4. A failed seed write hides output that was already captured (low)

`open_spill` sets `self.file_created = true` *before* `file.write_at(seed, position=0)`. If the seed write fails, `file_created` stays true and `file_bytes` stays 0, so `spill_path()` returns `Some` and `read_all` reads zero bytes — returning `""` rather than the head that is sitting in memory. The field comment says the path is retained "for inspecting partial output", but the read path can't reach the memory copy any more.

`failure` is set and the execution is killed, so this is narrow. Still, honest reporting would be to fall back to the head when `file_bytes == 0 && failure is Some(_)`, or to set `file_created` only after the seed succeeds and keep a separate `path_exists` flag if the path really needs retaining.

### 5. Smaller things

- `ShellExecution::output_bytes` falls back to `@utf8.encode(self.head()).length()` when no file was created. For a memory-only sink whose output passed `inline_cap`, that understates the real total (the head is capped). Fine for the no-spill-path case where the rest is dropped; worth a comment saying so, since the value is surfaced as "N B" in the UI.
- `generation: "memory-\{@env.now()}"` collides for two memory-only runtimes constructed in the same millisecond. Not persisted, so cosmetic, but a counter would cost nothing.
- `sync_job_directories` fsyncs `dir`, `dir/..` and `dir/../..` on **every** publish. Only the first publish in a generation needs the parent-directory syncs; subsequent ones only need `dir` (the rename target). Three opens + three fsyncs per lifecycle event is more than the plan's "match the session store's best-effort policy" implies.
- `@fs.tmpdir` failure is now a hard `fail` out of `build_tools` rather than a silent memory-only degrade. That is what the plan asked for ("not a silent production fallback") and the error text is actionable — but it changes the engine's startup failure mode, so it belongs in the PR description.
- `start` now calls `exec.background()`, which also clears `kill_when_full`. `spawn_exec` doesn't set that flag, so this is a no-op today; the state change `Running → Backgrounded` is the real effect. Worth a one-line comment on why `start` needs it, since `adopt`'s reason (the foreground waiter) doesn't apply.

Nothing here blocks. (1) and (2) are the two I'd fix before merge; (3) and (4) are cheap; the rest are notes.

PR #1468, opened 2026-09-11
Expose retained job history and bounded Desktop controls

bobzhang:
Desktop can list retained background jobs, read bounded log slices, and stop a live job through scoped host commands. The host overlays generation-specific live state on retained history, allowing controls to work independently of the model turn.

The PR includes ownership and recovery rules, 100-record history pages, 64 KiB UTF-8-safe byte reads, explicit missing/truncated/storage errors, and preservation of live jobs when metadata or directories are damaged. POSIX, Windows, and UNC path handling and all related review fixes live in this layer.

## Follow-up hardening

- Separate Stop delivery (5 s) from settlement (30 s). A settlement timeout explicitly reports unconfirmed completion; delivery alone never becomes Stopped.
- Consume correlated snapshot/stop replies after routing, including late replies, instead of broadcasting them to every browser.
- Distinguish a failed live refresh from an external owner while preserving generation-specific control checks.
- Add actual stream-consumer isolation and settlement-timeout tests, and extend owner diagnostics coverage.

The review's proposed clamp to persisted output_bytes is not applied: metadata changes on lifecycle events and would freeze reads of growing logs. The runtime already hard-cancels the direct process. Empty-generation cleanup belongs in #1466; generation caching and scan quotas remain separate optimizations.

## Validation

Current-head `4b254ce31` [CI](https://github.com/moonbitlang/openseek/actions/runs/34663737032) and [Desktop/Windows](https://github.com/moonbitlang/openseek/actions/runs/34663737041) passed. This PR remains one commit and is mergeable.

- The complete final stack also passed 86 browser E2E scenarios and eight Windows job-host regressions; final aggregate evidence is recorded in #1469.

## Stack and review history

### DIFF
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -102,6 +102,18 @@ jobs:
           CL: /bigobj
         run: moon run ./desktop/package/windows --target native -- --release --target app
 
+      - name: Test Windows background job paths and retained reads
+        env:
+          CL: /bigobj
+        run: |
+          # The test executable lives outside the packaged app, so Windows
+          # cannot find its linked Proton/CEF DLLs beside the executable.
+          $runtimeDirs = Get-ChildItem desktop/dist/seekmoon -Recurse -Filter *.dll |
+            Select-Object -ExpandProperty DirectoryName -Unique
+          if (-not $runtimeDirs) { throw 'Packaged app contains no runtime DLLs' }
+          $env:PATH = ($runtimeDirs -join ';') + ';' + $env:PATH
+          moon test desktop/internal/engine/jobs_wbtest.mbt --target native
+
       - name: Upload Windows app
         uses: actions/upload-artifact@v4
         with:
--- a/desktop/internal/api/api.mbt
+++ b/desktop/internal/api/api.mbt
@@ -443,6 +443,15 @@ pub fn endpoints(
     @endpoint.Endpoint::remote(@commands.agent_compact, async fn(payload) {
       @engine.compact_run(state.manager, payload)
     }),
+    @endpoint.Endpoint::remote(@commands.jobs_list, payload => {
+      @engine.list_jobs(state.manager, payload)
+    }),
+    @endpoint.Endpoint::remote(@commands.jobs_read, payload => {
+      @engine.read_job(state.manager, payload)
+    }),
+    @endpoint.Endpoint::remote(@commands.jobs_stop, payload => {
+      @engine.stop_job(state.manager, payload)
+    }),
     @endpoint.Endpoint::remote(@commands.agent_goal, async fn(payload) {
       @engine.goal_run(state.manager, payload)
     }),
--- a/desktop/internal/commands/commands.mbt
+++ b/desktop/internal/commands/commands.mbt
@@ -751,3 +751,18 @@ pub let browser_close : Command[
   @protocol.BrowserIdPayload,
   @protocol.EmptyReply,
 ] = Command("browser.close")
+
+///|
+pub let jobs_list : Command[@protocol.JobsListPayload, @protocol.JobsListReply] = Command(
+  "jobs.list",
+)
+
+///|
+pub let jobs_read : Command[@protocol.JobReadPayload, @protocol.JobReadReply] = Command(
+  "jobs.read",
+)
+
+///|
+pub let jobs_stop : Command[@protocol.JobTarget, @protocol.JobStopReply] = Command(
+  "jobs.stop",
+)
--- a/desktop/internal/commands/pkg.generated.mbti
+++ b/desktop/internal/commands/pkg.generated.mbti
@@ -165,6 +165,12 @@ pub let host_open_path : Command[@protocol.OpenPathPayload, @protocol.OpenPathRe
 
 pub let host_reveal_path : Command[@protocol.OpenPathPayload, @protocol.RevealPathReply]
 
+pub let jobs_list : Command[@protocol.JobsListPayload, @protocol.JobsListReply]
+
+pub let jobs_read : Command[@protocol.JobReadPayload, @protocol.JobReadReply]
+
+pub let jobs_stop : Command[@protocol.JobTarget, @protocol.JobStopReply]
+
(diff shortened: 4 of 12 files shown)


bobzhang:
## Review — host slice (list / read / stop bridge)

The byte-cursor work is the strongest part of the stack. `job_utf8_window` correctly skips continuation bytes only on a genuine mid-file tail, rejects a cursor that isn't on a boundary, trims an incomplete trailing sequence by lead-byte width, and validates lead bytes (`0xc2..0xf4`, so overlongs and `0xf5+` are refused). `read_job_record` checks that the decoded record's `generation`/`id` match the directory it came from. `jobs_store_path` normalizing both sides of the ownership comparison, and the Windows CI job that exercises it, are the right amount of paranoia. The `owner.lock` `try_lock` projection to `Interrupted` is exactly the plan's owner-verification contract.

### 1. The 5 s `job_request` timeout is applied to `JobStop`, whose reply arrives only after the process settles (medium–high)

`job_request` wraps `reply.get()` in `@async.with_timeout(5000, …)`. On the engine side, `JobStopResult` is emitted only after `jobs.stop(job_id)` returns, and that awaits `exec.wait()` (process actually terminal) plus `wait_any` (completion published). `request_stop` sends one cancel to the direct child; there is no SIGTERM→SIGKILL escalation anywhere, so the ceiling is the 30-minute reaper.

Consequence: any process that doesn't die within ~5 s of the cancel makes `jobs.stop` raise `EngineError`, which #1469 renders as an error under the Stop button and clears `stopping`. Meanwhile the stop is genuinely in flight and `defer self.job_requests.remove(request_id)` has already torn down the waiter, so the real `JobStopResult` is dropped on arrival. The UI only recovers a second later via the next `jobs.list` poll, if the process died by then. A well-behaved process with a large final drain can hit this too.

Two ways out, either is fine:
- Split the timeouts — keep 5 s on the `ack` (delivery really should be fast) and give the reply a much longer budget, or none, relying on the request's own cancellation.
- Or change the contract: have the engine reply `Stopped` as soon as the stop is *requested* and let `job_changed` report settlement. That also matches "keep stopping asynchronous relative to command intake" better than the current blocking `BgJobRuntime::stop`.

Either way, a `Stopping`/`StopTimedOut` outcome would let the UI say "stop requested, still running" instead of showing a transport error.

### 2. `read_job_page` reads to `file.size()`, ignoring the record's committed `output_bytes` (medium)

#1466 goes to some trouble never to advance `file_bytes` past a failed write, precisely so that a committed cursor exists. The reader then ignores it and uses the raw `file.size()`. So bytes from a torn/partial write are inside the page, and `job_utf8_window` will raise `invalid UTF-8 in job output` for the whole read rather than returning the committed prefix — turning a degraded log into an unreadable one. The same applies to a mid-`write_at` tail observed while the writer is active (the width trim handles a truncated-but-valid sequence; it hard-fails on an invalid lead).

Suggest clamping `size` to `min(file.size(), job.output_bytes)` when `job.output_bytes > 0`, and reporting the gap as log health rather than as a read failure. The record already carries `output_error` for the "log is incomplete" story.

### 3. A failed or slow live snapshot mislabels a live job as owned elsewhere (medium)

In `list_jobs`, if `job_request(JobsSnapshot)` raises (including the 5 s timeout, or `session_root is None`, or the directory-mismatch guard), the page is built from disk with `live = false`. For a still-running record that sends `job_view` down the owner-lock path, where `try_lock` fails *because this very engine holds the lock* — so the row reports:

> The owner is running outside this Desktop connection; controls are unavailable.

and `controllable: live && …` is false, disabling Stop. That's the wrong diagnosis for the most likely cause. Worth distinguishing "we know the live engine for this scope exists but the refresh failed" (keep the row controllable, show the refresh error) from "no live engine claims this scope" (current message).

### 4. History listing cost is proportional to session age, not to job count (medium, given #1469's 1 Hz poll)

`job_history_page` does `readdir(jobs/)`, then per generation `@fs.kind` + `@fs.mtime(owner.lock)` + `readdir(jobs/<gen>)`, accumulating ids until `offset + 100` — every call. #1469 calls it once per second while the panel is visible, and also on every `job_changed`. Combined with #1466's unpruned empty generation directories (one per engine run that executed anything), that is unbounded growth in per-second syscalls for a long-lived conversation. Deep paging also re-scans from zero each time, so `Load older jobs` is O(n²) in pages.

Fixing the prune in #1466 removes most of this. Beyond that, caching the generation list keyed on the `jobs/` directory mtime would make the steady-state poll nearly free.

### 5. Correlated replies are also broadcast to every client (low–medium)

`ServeEngine::consume` routes `JobsSnapshot`/`JobStopResult` to `job_requests[request_id]` but does not consume them — they continue to the sink and reach every attached browser, where `decode_engine_event` returns `None` and discards them. So every `jobs.list` poll (1/s) pushes the full `JobRecord` array to all clients for nothing. Suppress these two after routing, or route them out of band.

### 6. The host re-derives the store layout instead of using `SessionStore::job_logs_dir` (low)

The plan asked for "a session-store artifact-path API rather than duplicating private path construction across packages". #1466 added `SessionStore::job_logs_dir`, `cmd/openseek` uses it — and then `jobs_store_path` here rebuilds `root/sessions/<id>/jobs` by hand, in two functions, with the `ArchivedStore` variant on top. I understand why (the host works from `@pathx.Absolute` roots, not a `SessionStore`), but the layout now lives in two places and a change to one breaks the other silently. Either expose a path-level helper from the store package that the host can call, or add a test that pins the two constructions to the same string.

### 7. Nits

- `job_utf8_window` ends with `ignore(@utf8.decode(bytes[start:end]))` and `read_job_page` then decodes the same slice again. Return the decoded string from the window function.
- `errors : Array[String]` is unbounded — one entry per unreadable generation or job. #1469 renders each as its own alert div. Cap it and add an "and N more" tail.
- `read_job_record`'s 1 MiB metadata cap and `read_job_page`'s `limit` bounds are good; consider the same explicit cap on the number of generations scanned per call, so a pathological directory can't make one request unbounded.

(1) and (2) are the two I'd want fixed before merge; (3) is a small conditional; the rest can follow.

PR #1469, opened 2026-09-11
Visualize background jobs with live logs and browser coverage

bobzhang:
Add a Jobs tab to the existing Desktop panel launcher. Users can inspect active and retained jobs, follow or pause output, copy a log path or tail command, reload the latest output, load older jobs, and stop the tracked execution.

The view polls only while visible, bounds reads and displayed text, preserves Unicode boundaries, and rejects stale replies after session or visibility changes. History refresh and storage failures remain visible even when output following is paused. UI review fixes and the complete browser regressions are included in this PR.

## Follow-up hardening

- Drain at most four 64 KiB pages per refresh cycle while following, display catching-up status, and defer a partial UTF-8 suffix when its cursor cannot advance.
- Serialize pause/resume and queued manual reload behind the outstanding selected-log read; cancel pending refresh intent when pausing or hiding the panel.
- Keep Stop available without an output path, keep the toolbar compact, and show readable stop reasons.
- Add state regressions and three production-browser scenarios: a large completed output burst, paused reload with a held response, and stopping a controllable job without a log path.

The existing opt-in Jobs launcher remains the product scope. A persistent conversation-level running-count entry and history-cache work are separate improvements.

## Validation

Current-head `c64c70dba` [CI](https://github.com/moonbitlang/openseek/actions/runs/34664735968) and [Desktop/Windows](https://github.com/moonbitlang/openseek/actions/runs/34664736056) passed. This PR remains one commit and is mergeable.

- Final-head GitHub tests passed: **3,095 native, 3,198 JS, 2,345 Wasm**, 38 cram cases, and offline CLI/workflow integration.
- The rebuilt production browser passed **86/86 E2E tests without retries locally**, and **86/86 in GitHub CI**. All eight Jobs scenarios passed, including the final read-error/no-log assertions. Windows packaging and all **8 native job-host regressions** passed. Light/dark/narrow screenshots were inspected.
- Browser tests use production frontend code with a transport fixture backed by real subprocesses and retained files. Native host and real CLI tests cover the actual backend separately.

An earlier UI-head native attempt failed with process-spawn EAGAIN and a child-output assertion. Identical backend code passed runtime/host CI, and all 31 relevant tests passed locally. Its retry was superseded by the final UI-only correction. The final native run passed all 3,095 tests with no backend change or weakened test.

## Limits

The panel retains bounded recent output and exposes the complete retained file. Stop controls the tracked execution/direct child. Retention ends with session deletion; streaming does not fsync each chunk. Same-path external replacement with equal-or-larger content is outside the byte-cursor contract.

## Stack and review history

### DIFF
--- a/desktop/e2e/tests/background_jobs.spec.js
+++ b/desktop/e2e/tests/background_jobs.spec.js
@@ -1,0 +1,188 @@
+import { test, expect } from '@playwright/test';
+import { readFileSync } from 'node:fs';
+import { BackgroundJobsHarness } from './support/background_jobs_harness.js';
+
+let app;
+test.afterEach(async () => { if (app) await app.cleanup(); app = null; });
+
+async function mount(page) {
+  app = new BackgroundJobsHarness(page);
+  await app.install(); await app.goto(); await app.openSession(); await app.openJobs();
+  await expect(page.getByRole('region', { name: 'Job output' })).toContainText('ready 你好🙂');
+}
+
+test('real background output follows, pauses, resumes, copies its file and stops', async ({ page, context }, testInfo) => {
+  await context.grantPermissions(['clipboard-read', 'clipboard-write']);
+  await mount(page);
+  const output = page.getByRole('region', { name: 'Job output' });
+  app.write('stdout without newline'); app.write('\nstderr captured\n', true);
+  await expect(output).toContainText('stdout without newline');
+  await expect(output).toContainText('stderr captured');
+  await page.getByRole('button', { name: 'Pause', exact: true }).click();
+  app.write('while paused\n');
+  await expect.poll(() => readFileSync(app.livePath, 'utf8')).toContain('while paused');
+  await expect(output).not.toContainText('while paused');
+  await page.getByRole('button', { name: 'Follow output', exact: true }).click();
+  await expect(output).toContainText('while paused');
+  await page.getByRole('button', { name: 'Copy path', exact: true }).click();
+  await expect.poll(() => page.evaluate(() => navigator.clipboard.readText())).toBe(app.livePath);
+  await page.getByRole('button', { name: 'Copy tail command', exact: true }).click();
+  await expect.poll(() => page.evaluate(() => navigator.clipboard.readText())).toContain("'\\''");
+  app.write(Array.from({ length: 150 }, (_, i) => `build step ${i} passed\n`).join(''));
+  await expect(output).toContainText('build step 149 passed');
+  await expect.poll(() => output.evaluate(el => el.scrollHeight - el.scrollTop - el.clientHeight)).toBeLessThan(25);
+  await output.evaluate(el => { el.scrollTop = 0; });
+  await expect(page.getByRole('button', { name: 'Follow output', exact: true })).toBeVisible();
+  const top = await output.evaluate(el => el.scrollTop);
+  app.write('new output while reading earlier lines\n');
+  await expect.poll(() => readFileSync(app.livePath, 'utf8')).toContain('new output while reading');
+  await expect.poll(() => output.evaluate(el => el.scrollTop)).toBe(top);
+  await page.screenshot({ path: testInfo.outputPath('jobs-light.png'), fullPage: true });
+  await page.emulateMedia({ colorScheme: 'dark' });
+  await page.screenshot({ path: testInfo.outputPath('jobs-dark.png'), fullPage: true });
+  await page.getByRole('button', { name: 'Stop job', exact: true }).click();
+  await expect(page.locator('.jobs-detail-title')).toContainText('Stopped');
+  await expect(page.getByRole('button', { name: 'Stop job', exact: true })).toBeDisabled();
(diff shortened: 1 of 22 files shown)


bobzhang:
## Review — Jobs UI slice

The browser coverage here is the best part of the stack: a real subprocess writing to a real file behind the transport seam, with the production bundle in front of it. `background_jobs.spec.js` covers stdout/stderr interleaving, output with no trailing newline, multibyte text, pause → write → resume, manual-scroll auto-pause with a scrollTop assertion, clipboard for both path and tail command, stop with a `signalCode` check, a deliberately delayed obsolete `jobs.read` reply, conversation switching, page reload, and a selected job beyond the first 100 rows. The epoch/serial invalidation in `state.mbt` is correct and the tests actually prove it.

Also right: `subscriptions` returns `@sub.none` when the panel isn't active, so there is no hidden-view polling; the 262,144-UTF-16-unit buffer cap with the trailing-surrogate guard; and `tail_command`'s quoting — `'\''` for POSIX and `''` for PowerShell are both correct, and the UNC/drive-letter detection is reasonable.

### 1. Stop, Pause and Reload are nested inside `if view.output_path is Some(path)` (medium)

In `view.mbt` the whole `jobs-toolbar` — Copy path, Copy tail command, Pause/Follow, Reload tail **and Stop job** — lives inside that branch. A job whose record has `output_file: None` therefore has no Stop button at all, and no way to pause or reload. That is reachable: memory-only retention, and any job whose log creation failed but which still registered.

Only Copy path and Copy tail command actually depend on the path. Split the toolbar so the lifecycle controls render unconditionally.

### 2. `page.more` is never read — no burst continuation, no final drain (low–medium)

`more` is returned by `jobs.read` and never consulted anywhere in `desktop/frontend`. Following therefore advances at most one 64 KiB page per 1 s tick, while the note above the pane says "Following output". A job writing faster than ~64 KiB/s falls permanently behind and the UI doesn't say so. The plan called for "a continuation for bursts and a final drain after terminal status".

Minimal fix: when a page comes back with `more: true` and `follow` is on, immediately issue the next read instead of waiting for the tick (bounded by a small per-tick page budget so it can't become a tight loop), and show "catching up" while `more` holds.

### 3. Every `Poll` re-lists from disk, and `job_changed` already tells you when to (low)

`Poll` fires `request_list(offset=0)` *and* `request_read` every second, and `JobsChanged` dispatches another `Poll`. Since the event is the exact signal, the timer only needs to be a backstop. Listing is the expensive side — see #1468 (4), where each list walks every generation directory. Suggest keeping the read at 1 s and dropping the list to a 5–10 s backstop, letting `job_changed` drive it otherwise. That also removes the list request from the steady state where nothing is happening.

### 4. Raw reason identifiers reach the user (low)

`status_label` special-cases `user` and `session_ended`, then falls through to `"Stopped · \{reason}"`. So the failure reasons the runtime actually produces render as **Stopped · output_limit**, **Stopped · time_limit**, **Stopped · capture_failed**. `status_class` already enumerates exactly those three — give `status_label` the same arms with sentences ("Stopped · hit the output cap", "Stopped · hit the 30-minute limit", "Stopped · output capture failed"). The plan asked to "display the reason when they stop a job", which I read as prose, not the wire token.

### 5. The plan's persistent `Jobs · N running` conversation entry isn't here (product decision)

The plan's UX section opens with it, and it's the thing that makes the feature ambient rather than opt-in. What shipped puts "N active · M loaded" inside the panel, so a user with a job running sees nothing until they open the Jobs tab — and the tab is reachable only through the launcher row. The doc update in this PR records the change, but I'd flag it as a decision to take explicitly rather than absorb: right now a backgrounded `mbtx` is still discovered the old way, through the model's prose handoff.

### 6. Nits

- `tail -n 100` / `-Tail 100`; the plan said 200. Pick one and make the doc match.
- `Pause` sets `reading: false` while a read may still be outstanding, so `ToggleFollow` can put a second read in flight. `read_serial` makes this correct, but it breaks the plan's "one request in flight". Leave `reading` alone and let the serial guard drop the reply.
- `state.warnings` renders one `jobs-error` div per host error with no cap; with #1468's unbounded `errors` array a directory with many unreadable entries becomes a wall of alerts above the list.
- `Elapsed Ns` / `Last output Ns ago` are computed from `@interop.now_ms()` at render time, so they freeze whenever polling stops (panel hidden, disconnected, paused list). Fine, but the "Disconnected" banner is the only hint that the numbers are stale.
- The `jobs.read` UTF-8 windowing is reimplemented in `background_jobs_harness.js`, so a divergence from `job_utf8_window` wouldn't surface in the browser suite. The native tests cover the real one — worth a comment in the harness saying the two must be kept in step.
- This PR is where the plan doc flips to "implemented". #1466 and #1468 each merge carrying a plan that still says "Status: proposed … No runtime or UI changes implemented". Moving the status line and the decisions section to #1466 would make each PR self-consistent.

(1) is the one I'd fix before merge; (2) and (3) are worth doing while the code is fresh; (5) is a call for you rather than a defect.

PR #1474, opened 2026-09-12
fix(mbtx): avoid oversized sandbox arguments in review agents

bobzhang:
Review and explore agents can fail every `mbtx` call, including `fn main { println(1) }`, with `OSError("@process.run(): Argument list too long")`. Their source-write sandbox profile grows with the repository and was passed inline to `sandbox-exec -p`, exhausting the OS argument limit before the compiler or snippet starts. Scanning nested agent worktrees amplified that growth. Changing the backend or execution directory does not avoid the wrapper.

The source-write sandbox now:

- Writes its profile to a caller-owned temporary file and launches with `sandbox-exec -f`. `mbtx` retains the file in its per-call build directory through compilation and background execution, then uses the existing cleanup. Write errors propagate instead of disabling protection.
- Appends a final deny rule protecting the canonical profile path, even inside a writable scratch lab. The availability probe exercises the same file-based mechanism.
- Skips `.worktrees` and `.claude/worktrees` during directory enumeration and mtime tracking. A workspace rooted inside either location is still scanned normally; ordinary source directories named `worktrees` are preserved. The source-file regex remains in force, and literal directory rules remain necessary to prevent moving source directories outside the workspace.
- Runs a dedicated macOS native CI job for the sandbox package and a reviewer print/read smoke test. `OPENSEEK_REQUIRE_SANDBOX_TESTS=1` makes unavailable enforcement fail the job rather than silently skipping the kernel tests. Portable tests also check the file-based argv contract and worktree exclusions on Linux/Wasm.

Regression coverage includes a profile exceeding 2 MB, successful reads with denied source writes, blocked profile overwrite/removal inside a writable lab, and a blocked source-directory escape. The bounded worker profile remains inline, with its different growth bound documented. Constructor and lifecycle documentation describe profile creation and loading explicitly.

Validation:

- Reproduced errno 7 / `Argument list too long` with the original inline-profile launch.
- `OPENSEEK_REQUIRE_SANDBOX_TESTS=1 moon test agent_tool/internal/sandbox --target native` — 26 passed, enforcement required.
- `moon test agent_tool/internal/sandbox --target wasm` — 26 passed, including portable transport and scan checks.
- `just check`
- `just test` — 3,068 native tests, 3,186 JS tests, 38 cram cases, CLI lifecycle checks, and offline workflow cases passed.
- `just build`
- `moon info && moon fmt`
- CI workflow YAML parsed successfully.

The separately reported missing `OPENSEEK_REFERENCES` environment variable is not addressed by this change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,67 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
+  sandbox-macos:
+    name: sandbox (macOS, native)
+    runs-on: macos-latest
+    timeout-minutes: 20
+    permissions:
+      contents: read
+    env:
+      # Fail if sandbox enforcement disappears; guarded tests must not silently
+      # turn this platform-specific regression job green without running.
+      OPENSEEK_REQUIRE_SANDBOX_TESTS: "1"
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Set up MoonBit
+        run: |
+          curl -fsSL https://cli.moonbitlang.com/install/unix.sh | bash
+          echo "$HOME/.moon/bin" >> "$GITHUB_PATH"
+          "$HOME/.moon/bin/moon" version --all
+
+      - name: Update MoonBit package registry
+        run: moon update
+
+      # Even package-scoped tests run the workspace's module prebuilds, so
+      # Proton needs its runtime although these tests never open Desktop.
+      - name: Resolve pinned Proton version
+        id: proton
+        run: |
+          version="$(sed -n 's/.*"moonbit-community\/proton@\([^"]*\)".*/\1/p' desktop/moon.mod)"
+          test -n "$version"
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      - name: Restore Proton runtime
+        id: cef-cache
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            ~/.proton/store
+            ~/.proton/helpers
+          key: proton-runtime-${{ runner.os }}-${{ runner.arch }}-${{ steps.proton.outputs.version }}
+
+      - name: Assemble Proton CEF runtime
+        run: >-
+          moonx "moonbit-community/proton_cli@${{ steps.proton.outputs.version }}"
+          cef setup
+
+      - name: Save Proton runtime
+        if: github.event_name == 'push' && steps.cef-cache.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v4
+        with:
+          path: |
+            ~/.proton/store
+            ~/.proton/helpers
+          key: ${{ steps.cef-cache.outputs.cache-primary-key }}
+
+      - name: Test sandbox enforcement and oversized profiles
+        run: moon test agent_tool/internal/sandbox --target native
+
+      - name: Test reviewer printing and file reads
+        run: moon test agent_tool/mbtx --target native --filter 'read-only snippets can print and read workspace source'
+
   check:
     name: check (${{ matrix.moonbit-channel }}, ${{ matrix.target }})
     runs-on: ubuntu-latest
--- a/agent_tool/internal/sandbox/README.mbt.md
+++ b/agent_tool/internal/sandbox/README.mbt.md
@@ -22,10 +22,13 @@ Prepare, run, classify:
 ```mbt check
 ///|
 async test "prepare, run, and classify a shell command" {
+  let dir = @fs.tmpdir(prefix="sandbox-example-")
+  defer (@fs.rmdir(dir, recursive=true) catch { _ => () })
   let shell_text = "echo sandbox-ready"
   let prepared = @sandbox.SandboxedCommand::create_if_available(
     ".",
     Shell(shell_text),
+    profile_path="\{dir}/profile.sb",
   )
   let (program, args) = match prepared {
     Some(command) => (command.program(), command.args())
@@ -52,7 +55,10 @@ constructor, because the kernel matches profile rules against real paths.
 
 The profile is built from the tree as it exists at preparation; renaming the
 root or subtree afterwards makes the rules stale. Prepare close to where the
-command runs.
+command runs. The caller supplies `profile_path` in a temporary directory and
+keeps it until the command exits. The profile travels through `sandbox-exec -f`,
+so repository size does not consume the OS command-line argument limit. `mbtx`
+keeps this file in its per-call build directory, including background runs.
 
 ## Commands
 
--- a/agent_tool/internal/sandbox/README.mbt.md
+++ b/agent_tool/internal/sandbox/README.mbt.md
@@ -56,9 +56,15 @@ constructor, because the kernel matches profile rules against real paths.
 The profile is built from the tree as it exists at preparation; renaming the
 root or subtree afterwards makes the rules stale. Prepare close to where the
 command runs. The caller supplies `profile_path` in a temporary directory and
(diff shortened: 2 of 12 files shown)


bobzhang:
cc @tonyfettes

bobzhang:
## Review

The fix is the right one and the mechanism is correct. Things I checked and they hold:

- `-f <path>` is the correct replacement for `-p <profile>`, and the path is `realpath`'d before it goes into argv, so a relative CWD at spawn time can't misresolve it.
- The availability probe still runs *before* the write, so off macOS `create_if_available` returns `None` without touching the filesystem — no new failure mode on Linux/Windows.
- A write failure raises instead of silently dropping enforcement, which is the same "misconfiguration cannot disable protection" discipline the existing input validation follows.
- `confine()` is called immediately before each of the two spawn sites (`build_snippet` for the wasm build phase, and the single-shot/staged run), so the file is always freshly written for the spawn that consumes it. The build phase is synchronous, so the run phase's `CreateOrTruncate` can't pull the profile out from under a live build.
- Cleanup checks out on both exit paths: `cleanup()` removes `dir`, and after background adoption `terminal_cleanup` removes `build_dir` (or `dir`) at terminal status. The PR body's "reclaimed by the existing cleanup" is accurate.

### 1. The regression test never runs in CI (medium–high)

Every test in `agent_tool/internal/sandbox` sits behind `guard sandbox_exec_available() else { return }`, and the `capability_test.mbt` argv-shape assertions early-return off macOS because `create_if_available` returns `None` there. No workflow runs `moon test` on macOS: `ci.yml` is `ubuntu-latest` × three targets, `desktop.yml` is windows + ubuntu, and the only `macos-latest` jobs are `editor.yml` and the release-gated `desktop-release.yml`.

So the new 2 MB-profile test, the `assert_eq(args[0], "-f")` assertion, and the `mbtx_test.mbt` read-only case all no-op in CI. A bug that broke *every* `mbtx` call for the review/explore roles is protected only by tests that run on developer machines — a revert to `-p` would go green.

Worth a narrow macOS job; `moon test agent_tool/internal/sandbox --target native` on `macos-latest` is a couple of minutes and would cover the whole sandbox package, not just this fix. If that's not wanted, please say so in the PR body so nobody reads green CI as coverage here.

There's also a platform-independent assertion available cheaply: "the profile never enters argv" is checkable on Linux if the argv construction is reachable without the availability probe (or by asserting on `source_write_readonly_profile_data` plus a pure argv builder). Right now nothing outside macOS can catch the regression.

### 2. Argv length was the only thing observing the profile's growth, and the growth is untouched (medium)

`source_write_readonly_profile_base` emits one `(deny file-write* (literal …))` per source-containing directory, and `should_prune_profile_scan_path` prunes `.git`, `.hg`, `.svn`, `.jj`, `node_modules` — but **not `.worktrees`**. This checkout has 22 worktrees, each a full copy of the repo, so the scan walks ~23× the tree and the profile carries ~23× the directory list. That is the dominant term behind the 2 MB+ profile this PR routes around.

After the fix, that cost is still paid, just silently: a multi-MB file written on every `mbtx` call, and `sandbox-exec` compiling a multi-MB SBPL profile on every spawn. The hard failure is gone; so is the signal. (Same root cause makes `cached_source_write_profile_valid` stat every scanned directory per call.)

Adding `".worktrees"` to the prune list is a one-word change and removes the dominant factor. It's arguably in scope here — it's the same defect, and this repo is the one that reported it.

Separately worth deciding out loud: the `(deny file-write* (regex …))` rule already covers every protected source *file* anywhere under the root, so the per-directory literal denies only add directory-level protection (creating/removing entries, `rm -rf` of a source dir). If that's the intent, a comment saying so keeps the next reader from deleting them as redundant. If it isn't, the list can go entirely and the profile becomes O(1) — which would make this whole class of bug unreachable rather than worked around.

### 3. The profile file lives inside the region the confined process may write (low, hardening)

`writable_subtree` is `dir` (or the enclosing lab), `build_dir = "{dir}/build"`, and the profile lands at `build_dir/source-write.sb`. So a read-only subagent's snippet can rewrite its own sandbox profile.

Not exploitable as written — `sandbox-exec` compiles the profile at exec time, and `confine()` rewrites it immediately before each spawn — but the safety rests on an invariant that is stated nowhere and enforced nowhere: *confine is called once per spawn, immediately before it*. A future retry, a re-spawn on transient failure, or a second phase that reuses a prepared `SandboxedCommand` would break it silently, and the failure mode is "the sandbox was disabled", which no test would notice.

Cheapest fix: SBPL is last-match-wins and the code already appends the lab allow last, so one more line after it — `(deny file-write* (literal "<profile_path>"))` — closes it permanently. Otherwise place the file outside `writable_subtree` while keeping it under the cleanup root.

### 4. `create_worker_if_available` still passes its profile inline (low)

It stays on `-p`. Its profile is bounded today — `(allow default)`, one deny per `deny_roots` entry, two allows and one deny — so there's no bug. But the two constructors now differ in launch mechanism for a reason that isn't visible from the worker call site, and anything that grows `deny_roots` reintroduces exactly this failure. Either move it to `-f` for uniformity, or leave a one-line comment there pointing at `create_if_available` for why the source-write profile can't use `-p`.

### 5. Smaller things

- `create_if_available` now writes a file and gained a raise path, but the name still reads as a pure probe. The doc comment covers it; `prepare_if_available` would carry it in the signature.
- "Keep the file unchanged until the command exits" is stronger than the actual contract — the profile is compiled into the kernel at `exec`, so it only has to be correct *until the spawn*. Nothing depends on the stronger wording (`build_dir` survives to terminal status in the background path anyway), and over-stating it makes the contract look harder to satisfy than it is.
- The tests and the README example put `profile_path` inside the workspace root being sandboxed (`"{dir}/profile.sb"`, `"{root}/profile.sb"`) — exactly what the new README text tells callers not to do ("a caller-owned file in an existing temporary directory"). Harmless (the profile is computed before the file is written) but it also perturbs the `watched_dirs` mtime cache for that root, and the examples are what people copy.
- `sandbox_exec_available()` still probes with `-p`, so the probe doesn't confirm that `-f` works on the host it just declared available. Fine in practice; noting it because the probe is the thing standing between "unavailable" and "enforced".

None of this blocks the fix — it addresses a real outage and does so in the right place. (1) is the one I'd want resolved before merge, since without it the fix has no guard; (2) is the underlying defect and is a one-word change away.

bobzhang:
Correction to finding (2) — I named the wrong directory, and understated the effect.

A full scan finished after I posted. Source-containing directories in this checkout, by the profile scan's own exclusions (`.git`, `node_modules`, `_build`, `.mooncakes`):

| | dirs | share |
| --- | --- | --- |
| `.claude/worktrees` (52 worktrees) | 14,036 | 65% |
| `.worktrees` (22 worktrees) | 6,607 | 30% |
| everything else | 1,114 | 5% |
| **total** | **21,757** | |

So pruning `.worktrees` alone cuts ~30%, not the 23× I claimed — and it **misses the bigger one entirely**. `should_prune_profile_scan_path` matches whole path components, so `".worktrees"` does not match `.claude/worktrees`, whose component is `worktrees`.

The prune list needs both spellings. Between them they are 95% of the scan, which is the real reason the profile passed ARG_MAX.

The rest of finding (2) stands, and the tail is worth a look too — `.moonagent/eval_runs` (245), `desktop/target` (240), `desktop/dist` (161), `Desktop.app/Contents` (80), `.repos/*` vendored clones (~40). Those are build outputs and vendored copies carrying source that the source-write profile has no reason to protect; the genuine first-party tree is roughly 200 directories. That also strengthens the alternative I raised: if the per-directory literal denies are not load-bearing beyond the regex rule, dropping them makes the profile O(1) and none of this enumeration matters.

PR #1475, opened 2026-09-12
refactor(mbtx): render the spawn allowlist instead of restating it

bobzhang:
## The duplication

`spawnable_commands` (`agent_tool/mbtx/wasm_policy.mbt`) builds the moonrun
process policy, and the same vocabulary was written out twice more by hand —
in the `mbtx` tool description and in the default system prompt. The table's
own doc comment named the hazard:

> The mbtx tool description and the default system prompt both carry this list
> in prose. Admitting a command here without adding it to both costs a turn
> steps: the model never learns it may use it.

**Both copies had already drifted:**

| Entry | In the table | Tool description | System prompt |
| --- | --- | --- | --- |
| `moon --version` / `--help` | ✅ | ❌ | ❌ |
| `moonx` | ✅ | ❌ | ✅ |

The `moon --version` entry is the pointed one: it was added to the table
*because* a live run burned three attempts on the refusal — plain, then with a
rebuilt `PATH`, then by absolute path — reading each one as a broken toolchain
rather than as the allowlist. The fix landed in the table and in neither place
the model reads.

## The change

`spawn_list_markdown()` renders the bullet list from the table itself. Verbs
come from the prefixes; multi-token prefixes group under their shared first
token (`worktree list|add|prune`); a new `spawn_notes` map carries **only** the
caveats no prefix can express (`Not \`config\``, the refused pre-subcommand
globals). Admitting a command is now one edit to one table.

Three copies became one:

- `agent_tool/mbtx/mbtx.mbt` interpolates the render into the tool description.
- `prompt/default_prompt.mbt.md` cites that description by heading instead of
  restating it — which also stops the model paying for the same ~25 lines
  twice in one context window. The MoonBit-alternatives table and the
  `moongrep` cross-reference stay in the prompt, where the teaching belongs.

Both drifts correct themselves as output. **The table is untouched, so the
policy moonrun enforces is byte-for-byte unchanged.**

## Tests

- `the rendered spawn list covers every admitted command` — every program and
  verb in the table must appear in the render, checked **inside that program's
  own bullet**. The loose version of this passed with `moon add` deleted,
  because `add` also occurs in `worktree list|add|prune`; `spawn_bullet`
  isolates one bullet and `bullet_lists_verb` splits the verb region so a
  grouped prefix or a note's commas cannot satisfy it. Mutation-tested:
  dropping a verb from the render fails the test, restoring it passes.
- `the rendered spawn list groups multi-token prefixes under one verb`
- `the mbtx description carries the rendered spawn list`
- `the description advertises the Proton CLI command surface` (existing) now
  flattens whitespace before matching — it was pinning exact line breaks that
  a generated bullet reflows whenever a verb is admitted.

This replaces the review-time discipline the old comment settled for:

> Keeping the list and the prompts in step is a review-time job, not a
> test-time one.

## Verification

`moon check --deny-warn` (root) clean · mbtx suite 106/106 · prompt 5/5 ·
cram 38/38 · `just check-prompt` fresh · `moon fmt` stable.

https://claude.ai/code/session_01CcSFBB6Pd7Pkj5Xw6Ups9J

### DIFF
--- a/agent_tool/mbtx/mbtx_wbtest.mbt
+++ b/agent_tool/mbtx/mbtx_wbtest.mbt
@@ -139,10 +139,126 @@ test "the spawn policy admits Proton subcommands but not global cwd options" {
   assert_false(proton_prefixes.contains(["--cwd"]))
 }
 
-// The spawn allowlist is taught by each role's system prompt now, and that
-// prose is NOT pinned by a test: asserting a prompt contains particular words
-// only breaks on rewrites. Keeping the list and the prompts in step is a
-// review-time job, not a test-time one.
+///|
+/// The prose that teaches the spawn allowlist is RENDERED from it, so a
+/// command admitted to `spawnable_commands` reaches the model without a
+/// second edit. This pins that: every program and every verb in the table has
+/// to appear in the rendered list, which is the property the old hand-written
+/// prose could not hold — `moon --version` and `moonx` had both drifted out
+/// of a copy by the time it was replaced.
+///
+/// The wording around the list is deliberately not asserted; that only breaks
+/// on rewrites.
+test "the rendered spawn list covers every admitted command" {
+  for entry in spawnable_commands {
+    let (program, args_prefix) = entry
+    let bullet = spawn_bullet(program)
+    match args_prefix {
+      // A program admitted with no prefix takes any arguments; there is no
+      // verb to look for.
+      [] => assert_true(bullet.contains("any arguments"))
+      // Checked inside this program's OWN bullet and delimited, so a dropped
+      // `moon add` cannot be covered by the `add` in `worktree list|add|prune`.
+      [single] => assert_true(bullet_lists_verb(bullet, single))
+      // Longer prefixes are grouped under their first token, so the whole
+      // prefix is never one substring: check the group the parts land in.
+      [head, .. rest] =>
+        assert_true(
+          bullet.contains("`\{head} ") && bullet.contains(rest.join("|")),
+        )
+    }
+  }
+}
+
+///|
+/// The rendered bullet for `program`, whitespace flattened so a wrap cannot
+/// split a verb list, and with the following bullets cut off so a check reads
+/// only this program's grant. Fails the test when the program has no bullet
+/// at all, which is the coverage gap worth reporting by name.
+fn spawn_bullet(program : String) -> String raise {
+  let flattened = {
+    let builder = StringBuilder()
+    let mut in_space = false
+    for char in spawn_list_markdown() {
+      if char is (' ' | '\n') {
+        if !in_space {
+          builder.write_char(' ')
+          in_space = true
+        }
+      } else {
+        builder.write_char(char)
+        in_space = false
+      }
+    }
+    builder.to_string()
+  }
+  let opening = "- `\{program}` — "
+  guard flattened.find(opening) is Some(start) else {
+    fail("the rendered spawn list has no bullet for \{program}")
+  }
+  let body = flattened[start + opening.length():]
+  match body.find(" - `") {
+    Some(next) => body[:next].to_owned()
+    None => body.to_owned()
+  }
+}
+
+///|
+/// Whether `bullet` offers `verb` as a whole entry of its comma-separated verb
+/// list, rather than as a fragment of a longer word or of a grouped prefix.
+/// Flags render as code, so both spellings count.
+fn bullet_lists_verb(bullet : String, verb : String) -> Bool {
+  let verbs = spawn_verbs(bullet)
+  verbs.contains(verb) || verbs.contains("`\{verb}`")
+}
+
+///|
+/// The single-token verbs a bullet lists, split back out of the rendered
+/// prose. The verb list runs to whichever comes first: the `, plus ` that
+/// introduces the grouped multi-token prefixes, or the `. ` that ends the
+/// sentence and starts the note — so neither a grouped prefix nor a note's
+/// own commas can be mistaken for a verb.
+fn spawn_verbs(bullet : String) -> Array[String] {
+  let mut cut = bullet.length()
+  for marker in [", plus ", ". "] {
+    if bullet.find(marker) is Some(index) && index < cut {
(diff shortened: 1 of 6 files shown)


agent_tool/mbtx/mbtx_wbtest.mbt
```diff
@@ -139,10 +139,126 @@ test "the spawn policy admits Proton subcommands but not global cwd options" {
...
+/// The prose that teaches the spawn allowlist is RENDERED from it, so a
+/// command admitted to `spawnable_commands` reaches the model without a
+/// second edit. This pins that: every program and every verb in the table has
+/// to appear in the rendered list, which is the property the old hand-written
+/// prose could not hold — `moon --version` and `moonx` had both drifted out
+/// of a copy by the time it was replaced.
+///
+/// The wording around the list is deliberately not asserted; that only breaks
+/// on rewrites.
+test "the rendered spawn list covers every admitted command" {
```
bobzhang, line 152: **[LOW] Coverage test pins table ⊆ render, never render ⊆ table** — criterion 2's "invents none" half is untested.

`the rendered spawn list covers every admitted command` iterates `spawnable_commands` and asserts each entry appears in its program's bullet; nothing asserts the converse, so a render that emitted an **extra** verb or an extra grouped tail still passes the whole file. The only negative assertion about content is one hardcoded string (`assert_false(rendered.contains("stash list|show|pop"))`, mbtx_wbtest.mbt:243) — a spot check, not a property.

The render cannot invent anything today (`spawn_list_markdown`, wasm_policy.mbt:284-321, only copies table tokens), so this is a coverage gap rather than a live defect — but it is the half of the criterion no test holds.

agent_tool/mbtx/wasm_policy.mbt
```diff
@@ -218,6 +219,149 @@ let spawnable_commands : Array[(String, Array[String])] = [
...
+/// The caveats about the spawn allowlist that no `args_prefix` can express:
+/// which forms are deliberately absent, and which options are refused for
+/// sitting before the subcommand where no prefix reaches them. Keyed by
+/// program; a program named here renders its note after its verbs, and one
+/// absent renders its verbs alone.
+///
+/// Only the caveats live here. The verbs do NOT: `spawn_list_markdown`
+/// derives those from `spawnable_commands` itself, so admitting a command is
+/// one edit to one table and the prose that teaches it cannot fall behind.
+let spawn_notes : Map[String, String] = {
```
bobzhang, line 232: **[LOW] `spawn_notes` keys are unvalidated: a renamed program silently loses its caveat from the model-facing text.**

`spawn_notes` is keyed by free-form program name and nothing pins the keys against `spawnable_commands`. If a program is renamed in the table (or a key is typo'd), the note is silently dropped from `spawn_list_markdown`'s output and the model loses the caveats that carry the real enforcement limits — ``Not `config` `` / `-C`/`--cwd` refused / not `exec`, `install`, or a bare `codex`.

That is the same silent-drift class this refactor exists to remove, and precisely the caveats the removed hand-written copies *did* carry. No test in `mbtx_wbtest.mbt` or `mbtx_test.mbt` mentions `spawn_notes`.

bobzhang (commented):
## `openseek subrun review` — verdict: **goal MET**, no blockers

This PR was audited by OpenSeek's own review subagent (`openseek subrun review`,
`deepseek-v4-flash`, base `1fd3052ee`, 6 files, ~45 steps) against the eight
acceptance criteria in the PR description. Posting its findings verbatim.

It did not take the change on trust — it recompiled `wasm_policy.mbt` lines
1-363 in its own lab, printed `spawn_list_markdown()`, and diffed the output
against the table entry by entry:

- **Render completeness:** moon 25/25 (including the `--version`/`--help` the
  hand-written copy had lost), proton_cli 9/9, git 38 singles + all three
  grouped prefixes, gh 6/6, codex, and `diff`/`rg`/`moonx`. Nothing invented,
  grouping neither widens nor narrows a grant.
- **Determinism:** two separate `moon run` processes produced byte-identical
  1385-char output; `by_program`/`tails`/`spawn_notes` are only ever `get`-ed,
  never iterated, so no output depends on `Map` order.
- **Policy unchanged:** every table line and every line of
  `wasm_policy_document` is untouched context in the diff.
- **The tests bite** (mutation-tested against the real render): dropping the
  last single verb of each program fails the coverage test; dropping grouped
  tails fails it; and — the case that matters — dropping `add` and `` `--help` ``
  from *moon's* bullet alone still fails, even though `add` survives in git's
  list **and** inside git's `worktree list|add|prune`, proving `spawn_bullet`
  isolates one bullet and `bullet_lists_verb` cannot be satisfied cross-bullet.
- **`wrap_bullet`:** no loop or panic, word count preserved, a word longer than
  the 78-column width kept intact.

Gates it ran: `mbtx_wbtest` 10/10 on wasm · `moon fmt --check` clean ·
`moon check agent_tool/mbtx prompt --target wasm --deny-warn` exit 0 ·
`check-prompt` exit 0. It did not run the black-box `mbtx_test.mbt` suite or
the native/js deny-warn gate (those spawn real moon builds) — both are green
locally and in CI.

### Findings: 2 low, 1 nit

Two are inline below. The third is not on a changed line:

**[NIT] `agent_tool/mbtx/mbtx.mbt:97` — the last hand-written description of
what the allowlist holds, and it has already drifted.** The tool doc comment
still says the policy decides which programs may start, "an allowlist holding
the moon/git/gh toolchain and inert utilities". It is a doc comment, not
rendered to the model, so criterion 1 holds — but it is the one remaining
hand-written sentence about the list's contents, and it omits `codex` and names
none of `diff`/`rg`/`moonx`, the "inert utilities" it alludes to.

One non-issue it flagged and dismissed: the `gh` bullet now lists `api` before
the grouped `repo view`/`auth status`, a cosmetic reorder with no effect on what
the model believes it may run.

---
🤖 Review run by [Claude Code](https://claude.com/claude-code) via `openseek subrun review`

https://claude.ai/code/session_01CcSFBB6Pd7Pkj5Xw6Ups9J

PR #1483, opened 2026-09-12
refactor(agent): type session control state and outcomes

bobzhang:
Goal changes, autonomous input, and context yields currently depend on reserved message text, while goal baselines are paired by event adjacency. Replace these conventions with shared `GoalChange`, `InputOrigin`, and `TurnTerminal` values across the agent loop, serve scheduling, session projection, fleet evaluation, and desktop consumers.

A goal set now stores its captured or explicitly unavailable baseline in the same durable event. Successful completion and context yield are separate outcomes, so classification no longer examines answer text. Existing goal completion, blocking/resumption, automatic continuation, review boundaries, and model-facing wording are preserved. Regression coverage checks that ordinary text resembling control markers cannot change state or misclassify a successful answer.

The durable session format moves to version 2, with CLI and desktop readers updated together. Version 1 logs are deliberately unsupported; there is no legacy decoder or compatibility path. Browser fixtures, generated interfaces, and session documentation are updated to match.

Validation:

- `just check`: native and JS checks with warnings denied, plus formatting.
- `just test`: 3,111 native and 3,197 JS tests; 38 CLI documentation cases; offline turn-finish and workflow integration tests.
- `just build`: native and JS builds.
- Session viewer browser suite: 14 passed.
- Desktop transcript browser regressions: 16 passed.
- Merge check against current `main`: no conflicts.


desktop/internal/protocol/desktop_replies.mbt
```diff
@@ -408,7 +412,12 @@ pub impl FromJson for SessionUserMessage with fn from_json(json, path) {
   } else {
     None
   }
-  { content, submission_id, }
+  let origin : InputOrigin = match fields.get("origin") {
+    Some(value) => @json.from_json(value, path=path.add_key("origin"))
+    None =>
+      raise JsonDecodeError((path.add_key("origin"), "expected input origin"))
```
bobzhang, line 418: **A missing `origin` now silently deletes the user's prompt from the transcript.**

The doc comment 20 lines above this one warns against exactly this: *"Refusing there would drop a user's prompt out of the transcript over an annotation nothing renders, and the enclosing `SessionItem` decoder would swallow the refusal as `Unknown` rather than surface it."* The new `origin` branch raises, and `SessionItem::from_json` has `"user" => User(...) catch { _ => Unknown(json) }`, and `apply_session_item` ends with `Unknown(_) => ()`.

Scenario: a desktop frontend at this build talks to an older host (a remote `openseek` over the remote protocol, or a stale native host after a partial self-update). `session.load` / `session.load_archived` returns v1 user payloads with no `origin`. Every user prompt decodes to `Unknown` and vanishes; the transcript renders assistant turns answering nothing, with no error anywhere. `SessionDocument.version` is carried in the reply but the frontend never reads it (grep for `.version` under `desktop/frontend` finds only the file-editor's own counter), so the version bump this PR introduces buys no diagnosis on this path.

Either gate on `document.version` before projecting, or keep `origin` optional in the decoder with an explicit default so an old record degrades to a rendered prompt rather than a hole.

protocol/session_state.mbt
```diff
@@ -0,0 +1,285 @@
+///|
+/// Durable session snapshots and JSONL headers use this schema.
+pub const SessionFormatVersion : Int = 2
```
bobzhang, line 3: **`openseek_protocol` gains a whole public API surface but `protocol/moon.mod` still says `version = "0.1.1"`.**

This file adds `SessionFormatVersion`, `InputOrigin`, `GoalBaseline`, `GoalBaselineState`, `GoalChange` and `TurnTerminal` as public API, and the root `moon.mod`, `desktop/moon.mod` and `cmd/viz_app/moon.mod` all still import `bobzhang/openseek_protocol@0.1.1`.

Inside `moon.work` the local member wins (the protocol README says so outright), so `just check` / `just test` / `just build` all pass while resolving the working tree. Any build that resolves from the registry instead — a publish tarball, `moonx bobzhang/openseek/cmd/openseek`, or an external consumer of `bobzhang/openseek@0.3.1` — gets the published 0.1.1, which has none of these symbols, and fails to compile.

This is the exact failure commit c3ebe5a72 ("release: publish openseek 0.3.1 and openseek_protocol 0.1.1 for wasm") was written to fix: *"`protocol/` had drifted from the published `openseek_protocol@0.1.0` without a version bump."* Bump protocol to 0.1.2 and follow both importers.

agent_session/types.mbt
```diff
@@ -403,6 +372,8 @@ pub(all) enum SessionItem {
   /// Runtime-injected context, such as a plan reminder. JSON tag:
   /// `runtime_notice`. Projection sends it as a user-role message even though it
   /// did not originate from the user's prompt.
+  /// Durable goal transition, including the set operation's baseline.
+  Goal(GoalChange)
```
bobzhang, line 376: **The new `Goal` variant was inserted between `Runtime`'s doc comment and `Runtime` itself, so both variants are now documented wrongly.**

As written, `Goal(GoalChange)` carries *"Runtime-injected context, such as a plan reminder. JSON tag: `runtime_notice`. Projection sends it as a user-role message even though it did not originate from the user's prompt."* followed by its own one-liner, and `Runtime(RuntimeNotice)` on the next line has no doc at all. `missing_doc` is on for this module but does not fire, because the misplaced paragraph satisfies it.

While here: the `Terminal` variant's doc a few lines down still reads *"`Finished(text)` can project as a final assistant message; aborted, interrupted, and failed terminals project assistant status messages"* — it never mentions `ContextYield`, which this PR adds and which projects nothing at all.

agent/auto_compact.mbt
```diff
@@ -184,32 +184,28 @@ async fn AgentLoop::yield_at_context_ceiling(
...
-  // Built FROM the shared prefix so the durable-log contract
-  // (@agent_session.ContextYieldAnswerPrefix) can never drift from the
-  // answer actually persisted — and honest about whether a checkpoint
-  // actually exists (fail-open must not promise a summary that was never
-  // appended).
+  // Keep the displayed answer honest about whether a checkpoint exists.
+  // Its wording carries no state; the terminal variant does.
   let answer = if compacted {
-    @agent_session.ContextYieldAnswerPrefix +
+    "[context ceiling]" +
```
bobzhang, line 198: **`yield_at_context_ceiling`'s doc (line 156, just above the calls changed here) still asserts the opposite of what this PR does.**

It reads:

> The terminal is `Finished`: a distinct terminal kind would break old readers (the session decoder is strict), and continuation machinery chains finished turns. The `context_yield` event is the discriminator for consumers that must not read a yield as task success.

After this change the terminal is `ContextYield`, a distinct kind *is* what the decoder writes, and the stream event is no longer the discriminator — the terminal variant is. `finish_context_yield`'s own doc was updated; its only caller's was not, so the next reader gets the removed contract stated as fact.

agent/auto_compact.mbt
```diff
@@ -184,32 +184,28 @@ async fn AgentLoop::yield_at_context_ceiling(
...
     " This turn reached the model's context window and was checkpointed. " +
     "The work is not necessarily complete — continue it in a new turn, " +
     "which starts from the checkpoint summary."
   } else {
-    @agent_session.ContextYieldAnswerPrefix +
+    "[context ceiling]" +
```
bobzhang, line 203: **Reuse: removing `ContextYieldAnswerPrefix` left the literal `"[context ceiling]"` open-coded here, in the branch above, and in ~12 assertions across `agent/auto_compact_test.mbt`, `agent_session/session_test.mbt`, `cmd/openseek/serve.mbt` and `eval/session_analyzer/analyzer_test.mbt`.**

The constant no longer needs to be a *state* contract — that is the point of the PR — but it is still a shared presentation string, exactly like `GoalChange::model_content`, which this PR kept centralized in `protocol/session_state.mbt`. A private `const ContextCeilingLabel` in `agent` (or a `TurnTerminal::yield_label()` beside `model_content`) would keep the wording in one place instead of thirteen.

desktop/e2e/tests/support/desktop_browser_harness.js
```diff
@@ -98,7 +98,7 @@ export class DesktopBrowserHarness {
     // UML filter keeps branching nodes, so the fixture has one package with
     // two source dependencies and exercises the real renderer path.
     this.packageGraphSource = JSON.stringify({
-      version: 2,
+      version: 1,
```
bobzhang, line 101: **Collateral edit: this is the `moon.package_graph` fixture, not a session record, and its version was moved *backwards* (2 → 1).**

`moon tree --package --json` emits `"version":2` — see the real captures in `agent_tool/internal/auto_check/package_graph.mbt:287` and `:373`. Every other `version: 1 → 2` change in this PR is a session-format bump; this one flips an unrelated schema the wrong way, so the fixture no longer matches what the host actually receives.

It does not fail today only because `is_moon_package_graph_source` (`desktop/internal/host/package_graph_ops.mbt:85`) validates `nodes`/`edges`/`status`/`error` and ignores `version` entirely. Restore `version: 2`.

desktop/internal/protocol/desktop_replies.mbt
```diff
@@ -378,7 +379,10 @@ pub(all) struct SessionUserMessage {
 /// itself writes: an untagged prompt is byte for byte the line every record
 /// held before the field existed.
 pub impl ToJson for SessionUserMessage with fn to_json(self) {
-  let fields = { "content": Json(self.content) }
+  let fields = {
+    "content": Json(self.content),
+    "origin": ToJson::to_json(self.origin),
```
bobzhang, line 384: **The doc directly above is now false, and the key order diverges from the native writer.**

Lines 378-380 still claim *"Omit an absent tag rather than writing `null`, matching what the store itself writes: an untagged prompt is byte for byte the line every record held before the field existed."* With `origin` unconditionally emitted, an untagged prompt is no longer byte-for-byte anything that was written before.

Separately, this writes `content` then `origin`, while `agent_session`'s derived `ToJson` for `UserMessage` (field order `origin, content, submission_id`) writes `{"origin":"human","content":...}` — as the updated fixtures in `agent_session/session_test.mbt`, `tests/cram/cli.md` and `desktop/internal/engine/follower.mbt` all show. Two writers for one durable line now produce different bytes for the same value; if that is intended, the "matching what the store itself writes" sentence needs to go.

viz/view.mbt
```diff
@@ -382,12 +386,14 @@ priv enum UserInput {
...
 
 ///|
-/// Classify one user-role message. `delegated` is a property of the session
-/// being rendered, not of the text, so it decides first.
-fn user_input(content : String, delegated~ : Bool) -> UserInput {
-  if delegated {
+/// Classify typed provenance, retaining the delegated-session caption.
+fn user_input(
+  origin : @agent_session.InputOrigin,
+  delegated~ : Bool,
```
bobzhang, line 392: **`viz/README.mbt.md` still documents the text-matching rule this function just replaced — including the rationale for it.**

Line 38 of that README:

> The auto-continue text is matched against the constant the engine submits, so rewording that prompt stops the match instead of quietly captioning engine text as the user's.

and the classification table at line 34 keys the **Auto-continue** caption off `@agent_session.GoalContinuePrompt`. Both describe the mechanism this PR deleted; classification is now `origin is AutoContinue`, and rewording the prompt no longer affects it at all. The deleted test `"user cards name who actually wrote the message"` was the executable half of that contract, so nothing catches the drift.

`desktop/internal/engine/README.mbt.md:161` ("`goal_run` appends or steers a `[goal]` marker") and `protocol/README.mbt.md` (whose package table still lists only `Event, Usage, Command, SteerKind, to_json, parse`, and whose module description calls the module "the openseek run/serve stdout wire contract" though it now also owns the durable session schema) have the same problem.

cmd/openseek/concurrent.mbt
```diff
@@ -236,15 +226,16 @@ async fn run_one_attempt(
 ///|
 /// The terminal status and message of a finished run, read from its last
 /// `Terminal` event. `("(no terminal)", "")` when the run recorded none.
```
bobzhang, line 228: **Stale doc: this now returns a 3-tuple, and the doc still describes the 2-tuple.**

*"`("(no terminal)", "")` when the run recorded none"* — it is `("(no terminal)", "", false)` now, and nothing here says what the new `Bool` means or that `context_yield` is a status this can return.

`AttemptOutcome`'s doc at line 101 has the matching gap: *"`ok` is true only for a `finished` terminal; everything else (aborted/interrupted/failed/no terminal, or a thrown error) is a non-success"* — the new `context_yield` terminal belongs in that list by name, since it is precisely the case this PR was written to stop mis-scoring.

agent_runtime/runtime.mbt
```diff
@@ -144,18 +145,20 @@ pub fn[X] AgentTaskScope::group(
...
 /// deduplicate, or drop blank strings. The agent loop may apply its own
-/// filtering when it drains. `wake=true` marks a user-originated notice (for
-/// example a goal change) as actionable during job_wait without changing its
-/// persisted Notice representation. Ordinary notices do not wake waiting.
+/// filtering when it drains. Goals wake job_wait directly. `wake=true` can
+/// make a notice actionable too; ordinary background notices do not wake it.
 pub fn AgentRuntime::queue_steer(
   self : AgentRuntime,
   input : SteerInput,
   wake? : Bool = false,
```
bobzhang, line 153: **`wake` no longer has a production caller, and the one test that covers the behavior it existed for still fakes it.**

The only place that passed `wake=true` was serve's goal-marker path (`runtime.queue_steer(Notice(marker), wake=true)`), now `runtime.queue_steer(Goal(marker))`. Grepping the tree, `wake=true` survives only in `agent/job_wait_test.mbt:103` and `agent_runtime/input_wait_test.mbt:34,43,46`.

That first one matters: `agent/job_wait_test.mbt`'s `GoalChange` wake case is `runtime.queue_steer(Notice("NEW-GOAL"), wake=true)` — a shape serve no longer produces. So after this PR nothing verifies that a real `Goal(_)` steer interrupts a `job_wait`, which is the actual production path (`Goal(_) => true` two lines below). Switch that test to `queue_steer(Goal(GoalSet(...)))`; then decide whether `wake` still earns its place in the public API.

viz/view.mbt
```diff
@@ -411,7 +417,7 @@ fn UserInput::card_style(self : UserInput) -> (String, String) {
 /// summary keeps the class the raw log gives it, so the same event reads the
 /// same way in both views.
 fn user_role_card_style(
-  content : String,
+  _content : String,
```
bobzhang, line 420: **Dead parameter: `_content` is now unused but still threaded through both call sites.**

Every branch of this `match` reads only `origin` and `delegated`, yet `viz/view.mbt:2024` and `:2411` still compute and pass `message.content`. The underscore silences the warning rather than removing the argument. Drop the parameter and the two arguments.

desktop/frontend/model.mbt
```diff
@@ -4057,8 +4057,8 @@ fn Conversation::apply_commit(
   let changed = @transcript.apply_session_item(messages, item, ts?, sequence~)
   // A `[goal]` marker commit is the durable confirmation of a goal
   // set/clear/block — the only writer of the mirror the banner renders.
-  let goal_marker = if item is @protocol.RuntimeNotice(notice) {
-    @transcript.goal_marker(notice.content)
+  let goal_marker = if item is @protocol.Goal(change) {
+    Some(@transcript.goal_marker(change))
```
bobzhang, line 4061: **Simplification: `goal_marker` is total now, so wrapping its result in `Some(...)` just to unwrap it twice below is redundant state.**

The `Option` existed because the old `goal_marker(content : String) -> GoalMarker?` could fail to parse. It cannot fail any more. This reads more directly as a single match on the item:

```suggestion
  let goal_marker : @transcript.GoalMarker? = match item {
    @protocol.Goal(change) => Some(@transcript.goal_marker(change))
    _ => None
  }
```

…or better, collapse `goal`, `goal_marker` and the `goal_pending` test into one `match item { @protocol.Goal(change) => { … } _ => self }` arm, which removes both `is Some(marker)` re-tests.

The comment just above ("A `[goal]` marker commit…") should lose the literal marker text too — the commit is a typed `Goal` item now.

desktop/frontend/transcript/goal.mbt
```diff
@@ -16,17 +8,12 @@ pub(all) enum GoalMarker {
...
-    Some(GoalCleared)
-  } else if content == "[goal unblocked]" {
-    Some(GoalUnblocked)
-  } else if content.strip_prefix("[goal blocked]\n") is Some(reason) {
-    Some(GoalBlocked(reason.to_owned()))
-  } else if content.strip_prefix("[goal]\n") is Some(text) {
-    Some(GoalSet(text.to_owned()))
-  } else {
-    None
+pub fn goal_marker(change : @desktop_protocol.GoalChange) -> GoalMarker {
```
bobzhang, line 11: **Reuse: `GoalMarker` is now a structural duplicate of `@openseek_protocol.GoalChange`, and this function is a 1:1 remap that only drops the baseline.**

The file header that justified the duplication — *"this js-only package cannot depend on that native package, so the marker shapes are duplicated here"* — was deleted in this same diff, because the package now imports `bobzhang/openseek_protocol` directly (see `desktop/frontend/transcript/moon.pkg`). With that reason gone, `GoalMarker` is four variants mirroring four variants, and `apply_goal_marker`, `SessionLoadView.goal_markers`, `Conversation.markers`, `GoalRequest::settled_by` and both `sessions.mbt` match sites all carry the shim.

Using `@protocol.GoalChange` throughout and matching `GoalSet(text, ..)` where the baseline is irrelevant deletes this type, this function, and the `Some(...)` wrapper in `model.mbt`.

desktop/frontend/transcript/sessions.mbt
```diff
@@ -426,77 +426,69 @@ pub fn apply_session_item(
...
-                  is_error=false,
-                  body=Some(payload.content),
-                ),
-                ts,
-                sequence,
-              })
-            }
-        }
+      if !payload.content.is_blank() &&
+        !is_model_directed_notice(payload.content) {
```
bobzhang, line 478: **Altitude: the goal *state* is typed now, but this arm still decides what to render by string prefix — and it silently eats notices.**

`is_model_directed_notice` (`desktop/frontend/transcript/runtime_update.mbt:15`) is `content.has_prefix("[goal") || content.has_prefix("[plan reminder]")`. Any runtime notice whose text happens to start with `[goal` — a background-job completion quoting a goal, a tool result echoed into a notice — is dropped from the transcript with no card and no trace. That is the same defect class this PR set out to remove ("ordinary text resembling control markers cannot change state"), left standing one level down in the same `match` being rewritten here.

The deep fix is the same move applied to the remaining model-directed notices: give `[goal reminder]` / `[goal check]` / `[plan reminder]` a typed `SessionItem` (or a tagged `RuntimeNotice` kind) so "is this for the model or for the reader" is a field, not a prefix, and this predicate disappears.

protocol/session_state.mbt
```diff
@@ -0,0 +1,285 @@
...
+///|
+#doc(hidden)
+#deprecated
+pub extend TurnTerminal with FromJson::{from_json}
+
+///|
+pub impl ToJson for GoalBaselineState with fn to_json(self) {
+  match self {
+    Captured(value) =>
+      { "kind": "captured", "sha": value.sha, "dirty": value.dirty }
```
bobzhang, line 272: **`GoalBaselineState` flattens its payload next to the discriminator while `GoalChange` nests its own — two encodings for one value, now frozen by a format version.**

Here `Captured` writes `{"kind":"captured","sha":…,"dirty":…}`, and `from_json` (line 282) recovers it by re-decoding the *same* object as a `GoalBaseline`, which only works because derived `FromJson` ignores the extra `kind` key. `GoalBaseline`'s own derived `ToJson` produces `{"sha":…,"dirty":…}`, so the struct has two on-wire shapes depending on who serializes it, and adding any field named `kind` to `GoalBaseline` silently corrupts the tagged one.

`GoalChange::to_json` two functions up already shows the shape that does not have this problem: `{"kind":"set","text":…,"baseline":<nested>}`. Nesting here too (`{"kind":"captured","baseline":{…}}`) makes the payload one value with one encoding. Worth settling before v2 records exist in the wild, since changing it later costs another format version.

PR #1491, opened 2026-09-12
feat(desktop): add one-click Chats with persistent workspaces

bobzhang:
Users can now start a standalone conversation with one click on the new **Chats → +** action. No folder picker or confirmation is needed. Chats appears above Projects, starts expanded, and lists the three most recent conversation titles with Show more for older chats. The active conversation remains visible even outside the recent three. Projects continues to list user-selected directories. Managed workspace folders are excluded from Projects and the new-chat project chooser; their conversations remain in Chats after reload.

Each creation allocates a persistent directory under `SeekMoon/workspaces` in macOS Application Support, Windows LocalAppData, or the Linux user data directory. Development builds use the same stable location independently of the checkout. Exclusive directory creation safely allocates `Unnamed workspace`, `Unnamed workspace 2`, etc., preserving existing files across concurrent requests. The host supplies the canonical managed root with workspace snapshots so classification follows directory ownership rather than display names.

Chat-storage resolution errors are carried separately from the project registry and displayed in the sidebar, so ordinary project operations remain usable. A newer project choice invalidates outstanding chat allocation results to prevent stale focus changes.

The frontend reuses the existing workspace registration and agent selection flow, blocks duplicate creation, ignores stale connection/request results, and allows retry after failure. OpenSeek and Codex conversations share the flat Chats section. The New Chat shortcut allocates another managed workspace when used from a managed chat. Allocated folders are retained if registration fails; there is no automatic deletion.

Validation:
- `moon info`, `moon fmt`, and generated interface review
- `just check`, `just test`, `just build`
- Host tests: concurrent allocation, existing-file preservation, storage failures, stable data path
- Frontend/protocol tests: one-click creation, duplicate/stale requests, failure recovery, flat grouping, similarly named user projects, snapshot ordering, Codex draft visibility, JSON metadata validation
- Chromium tests: direct creation and reload, failure/retry, user project classification, three-row preview and Show more, existing project picker
- Dark-mode Chats screenshot visually inspected
- Native desktop launched with `just desktop-dev`; startup verified from window/renderer logs. Native automated clicking is unavailable because the computer-use client/server versions differ; interactive coverage uses the browser build.


bobzhang:
Fresh Codex CLI review — findings addressed

Ran `codex review --base origin/main` in two new CLI sessions (Codex CLI 0.153.4, gpt-6-astra), without resuming the implementation conversation. The first reviewed ec40e69a2774662b90adf99edfc6c21942c346d7; the follow-up reviewed 025afadaf.

Both actionable P2 findings are fixed:

- **Chat-storage failures blocked unrelated project operations.** Fixed in 025afadaf. Chat-root resolution now returns a typed success/failure value independently of the registry snapshot. Projects continue loading and accepting registry updates while the sidebar explicitly displays the storage error. Cancellation still propagates. Added host regression coverage for broken path components/non-directories, protocol round-trip validation, frontend snapshot/recovery tests, and a browser test that adds an ordinary project while Chats storage reports an error.
- **An older chat allocation could overwrite a newer project selection.** Fixed in 2b7e78d20. Opening the project picker or choosing another project invalidates the outstanding chat-creation request. A late result can no longer replace the newer pending adoption or steal focus. Added deterministic regression coverage for both newer project confirmation and direct project selection.

Validation on the final revision:
- `just check`, `just test`, `just build` passed.
- Frontend package: 472 tests passed.
- Five Chromium tests passed for the storage-error fix, including ordinary project loading/addition during failure and error recovery.

PR #1494, opened 2026-09-12
refactor(mcp): share configured header filtering

bobzhang:
The Streamable HTTP transport applied the same configured-header filtering loop to POST requests and the closing session DELETE. Both paths now call one private helper, keeping extras last and preserving the reserved-header filter and existing header-map mutation.

The existing integration test now sends Authorization and a mixed-case forged session header. It verifies Authorization on both POST and DELETE while preserving the server-assigned session id; new observations use optional values. Validation: focused transport tests 5/5, unchanged public interfaces, `moon info`, `moon fmt`, and root `just check`, `just test`, `just build`.

CI snapshot at `8616131ab`: 1 pending.

### DIFF
--- a/mcp/streamhttp/streamhttp_test.mbt
+++ b/mcp/streamhttp/streamhttp_test.mbt
@@ -25,6 +25,9 @@ struct SeenHeaders {
   mut protocol_version : String
   mut accept : String
   mut deleted_session : String
+  // A configured non-reserved extra, recorded on both request paths.
+  mut post_authorization : String?
+  mut delete_authorization : String?
 }
 
 ///|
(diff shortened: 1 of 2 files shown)


bobzhang:
Review sign-off for 8616131abafe54386b56425e20095992e647b531: no actionable findings.

The extracted helper preserves header mutation, ordering, and reserved-header filtering for both POST and session DELETE. The extended integration test covers Authorization forwarding on both paths and mixed-case session-header protection.

Validation: `moon test mcp/streamhttp --target native` passed all 5 tests. Full repository gates were not rerun.

Posting as a comment because the authenticated account is the PR author, so GitHub does not allow a formal self-approval.

PR #1510, opened 2026-09-14
refactor(grep-search): repair invalid patterns in one click

Betterlol:
## Summary

- replace the Pattern Repair composer/preview flow with a one-click repair action
- report API error when API not configure
- allow agent to use validate tool: `dump --expr <pattern> --strict-check --exit-code`

## Validation

- `moon info --frozen desktop/internal/pattern_repair desktop/internal/host`
- `moon check --target native desktop/internal/pattern_repair desktop/internal/host`


tonyfettes (commented):
Can we split the timeout error-reporting as a separate PR?

agent_pattern_repair/validation.mbt
```diff
@@ -0,0 +1,157 @@
...
+    Ok(validator)
+  }
+}
+
+///|
+async fn run_pattern_repair_validator(
+  validator : PatternRepairValidator,
+  pattern : String,
+  workspace_root : String,
+) -> Result[String, String] {
```
tonyfettes: pattern / workspace_root are both parameter of type `String`, using labelled parameter might be better.

tonyfettes: Also I'd recommend using error-handling over `Result[String, String]`.

PR #1515, opened 2026-09-14
feat(desktop): show edit and multi_edit captions with emojis

bobzhang:
## What

`mbtx` calls already render with a 🐇 caption. This extends the same treatment to the `edit` and `multi_edit` tools: ✏️ and 📝.

The caption mapping (`tool_display_label`) previously lived in `mbtx.mbt` and handled only `mbtx`. It moves to its own `tool_label.mbt` and becomes a small tool→emoji table. Only the **leading tool word** is replaced, so a brief keeps its own words (the file, the line, the count):

| brief | caption |
| --- | --- |
| `edited lib/parser.mbt:12 (1×)` | `✏️ lib/parser.mbt:12 (1×)` |
| `edit preview lib/parser.mbt (3 site(s))` | `✏️ preview lib/parser.mbt (3 site(s))` |
| `multi_edit lib/parser.mbt (3×)` | `📝 lib/parser.mbt (3×)` |
| `multi_edit reverted (2 new error(s))` | `📝 reverted (2 new error(s))` |

`edit`'s success brief opens with `edited`, which the tool name prefixes only as a substring, so that spelling is matched as its own case (a naive `strip_prefix("edit ")` would produce `✏️ ed ...`). A label that does not open with the tool's name is left untouched, so a phrase that merely mentions "edit" is never rewritten. Recorded tool names and payloads stay intact.

## Verification

- `moon test --target js desktop/frontend/transcript/component` → 27 passed, 0 failed (new `tool_label_wbtest.mbt` covers every brief spelling plus non-matching labels).
- `moon fmt --check desktop/frontend/transcript/component` clean.
- `moon check` at the module root: 0 errors.

Not run: the desktop Playwright e2e suite (its runner needs `npm`/`playwright`, unavailable here). No existing e2e test pins an edit/multi_edit caption, so nothing regresses; an e2e assertion could be added later for parity with the mbtx case.

Generated with SeekMoon


bobzhang:
## Self-review

Author's review of this change, backed by an independent worktree audit (`@builtin/review.mbtx`). Net result: **0 blockers, 0 high**; three low/nit findings — two fixed here, one left as-is.

### What I checked
- **Prefix completeness.** Enumerated every brief `edit`/`multi_edit` emit and confirmed each is name-prefixed or uses the `edited` spelling: `edit <basename>` (edit.mbt:156,171); `edited <basename>:<span> (append|insert)` (edit.mbt:224,229); `edited <basename>:<line> (1×)` (edit.mbt:284); `edit preview …` (edit.mbt:521); `edit rejected …` (edit.mbt:773). And `multi_edit` exact (multi_edit.mbt:181,264); `multi_edit <file> (<n>×)` / `multi_edit <n> files (<m>×)` (multi_edit.mbt:1280,1282); plus the rejected/reverted/partial briefs. The prefix match fires on every real label and never on an unrelated one (`editor notes`, `multipled` are left alone).
- **`mbtx` unchanged.** The `mbtx` arm reproduces the previous body exactly; a differential run over 29 labels × 8 tool names gave 0 differences for `mbtx` and 11 intended differences for `edit`/`multi_edit`.
- **Call sites.** Exactly four, all in `view.mbt` (307, 313, 478, 481), each threading `name`.
- **No regressions.** Repo-wide search found no test pinning an `edit`/`multi_edit` caption; the only caption assertions target `mbtx`/`read` and are unaffected.

### Findings
1. **Coverage gap (low) — fixed.** The tests fed hand-written labels, so two shapes the tools really emit were unpinned: `edit`'s append/insert briefs and `multi_edit`'s multi-file brief. Added both to `tool_label_wbtest.mbt`.
2. **Header wording (nit) — fixed.** The file header claimed a label is left alone unless it opens with the tool's name, which the `edited` branch contradicts. Corrected to match the code.
3. **Untracked `scripts/review-pr.mbtx` (nit) — left as-is.** Not part of this change and not authored here; it cannot trip CI (only tracked sources are gated). Flagged rather than deleted.

### Verification
- `moon test --target js desktop/frontend/transcript/component` → 27 passed, 0 failed
- `moon fmt --check desktop/frontend/transcript/component` → clean
- CI on head `852e1a628`: 10/10 checks pass, including `desktop e2e`

Caveats: Playwright could not run locally (`npm` unavailable), so local verification is unit-level; the `desktop e2e` gate passed in CI. `Devin Review` reports success but actually **skipped** its review ("trial expired and no credits remaining"), so it is not corroboration.

PR #1531, opened 2026-09-15
refactor(cli): keep openseek main small with internal packages

bobzhang:
## Summary

`cmd/openseek` had grown into the CLI implementation, with command dispatch, session setup, execution policies, and tests sharing one package. Reduce it to a 12-line executable entry point and move the implementation into focused internal packages.

- Separate options, setup, execution, run/fleet, serve, and the smaller command handlers under `internal/openseek`.
- Keep dependencies flowing from handlers to shared helpers; preserve the command interface, executable path, and JSONL output lifecycle.
- Move all 87 existing tests into the owning packages without changing their assertions, and document the package boundaries.

## Validation

- All 87 relocated tests passed; verified function bodies and tests changed only for package qualification, visibility, documentation, and formatting.
- `just check`
- `just test` — 3,134 native tests, 3,263 JS tests, 38 cram cases, and the turn-lifecycle and workflow regressions passed.
- `just build`
- `moon info && moon fmt` — existing public interfaces unchanged; new interfaces are internal.


bobzhang:
Review of the split (inventoried 216 top-level blocks old↔new: byte-identical modulo `moon fmt` wrapping, 87 tests moved by name, only the two required visibility widenings). No correctness bugs. The duplications and stale docs it introduced are fixed in the follow-up commit. Six layering/test-placement observations left as-is for a later pass:

1. **`ApprovalPolicy` lives in `options`**, so `execution` imports the argparse layer for one enum + `mcp_config_path`. The policy is an execution concept the flag tree merely selects; moving it (and the `approval_policy(matches, interactive~)` reader) into `execution` removes the edge.
2. **`setup` now carries `testkit/filesystem` in its production imports** (`prompt.mbt` → `@vfs.dir_outline`), so `execution`/`run`/`serve`/`commands` all transitively link a package documented as a test fixture. Previously only the leaf binary did. `dir_outline` wants a production home.
3. **`new_scratch_lab`/`remove_scratch_lab`** are defined in `commands/subrun.mbt` but also called from `commands/review.mbt` — the README's own "shared behavior goes below the handlers" rule points at `execution` or `setup`.
4. **`run_matches`** is duplicated verbatim in four `helpers_wbtest.mbt` files (`run`, `setup`, `execution`, `options`) and `sessions_leaf_matches` in two. A `for "wbtest"` helper package keeps one copy out of the `.mbti`s.
5. **`setup/session_wbtest.mbt`** asserts mostly `@options` behavior (`agent_settings`, `max_steps`, `task_text`, `retry_overrides`); those cases belong in `options`, which would also drop `setup`'s wbtest import of `options` + `cli`.
6. **The old `main.mbt` "no logger in this binary" invariant** (stdout carries only the event stream) is no longer written down anywhere; with six `moon.pkg`s it is easier to violate. Worth a line in `internal/openseek/README.md`.

PR #1541, opened 2026-09-15
feat(tools): carry structured result data and a program-callable opt-in

bobzhang:
First slice of #1532 (SDK-based programmatic tool calls), split for review. This PR is pure plumbing with no behavior change for the model or the tools.

## What changes

- `ToolOutput`, `ToolResult` (session log), `SessionToolResult` (desktop wire), and `JobRecord` gain an optional `data : Json?` field. It is persisted and forwarded to the desktop, and never enters model messages (test: `agent_session/program_data_test.mbt`).
- `web_search` attaches its normalized sources as structured `data` (`{sources: [{url, title?, snippet?, published_at?}], truncated}`), omitting absent fields rather than writing empty strings.
- `AgentToolDefinition` gains a host-owned `program_callable` opt-in (default false) with `is_program_callable()`, which also excludes loop-control tools. `edit`, `multi_edit`, and `web_search` opt in. Nothing reads the flag yet.
- `AgentToolCall::from_arguments` constructs a call from already-decoded JSON.
- Fixture updates for the new field shape in desktop tests and doc examples.

## Split plan

A (this) structured results → B shared file gate → C job lifecycle → D desktop nested cards → E PTC service package → F mbtx/agent integration → G prompts and docs. B and C are independent of A.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1532, opened 2026-09-15
Support SDK-based programmatic tool calls across mbtx and background jobs

bobzhang:
## What changes

`mbtx` scripts can call existing tools through the published `bobzhang/openseek_tools@0.1.0` SDK, for example `@tools.call("edit", {...})` or `@tools.call("web_search", {...})`, using the same JSON arguments as direct calls. PTC defaults on for supported wasm executions and survives the normal five-second background handoff. Desktop shows nested calls, edit diffs, and retained job traces.

This is one standalone PR against `main`; superseded stack PRs #1519, #1520, and #1522 are now closed. SDK #1518 is already merged and published. SDK 0.2.0 cleanup is now isolated in **#1535**, a separate PR against main. This host PR has **no changes under `tools_sdk/` in its final diff**. Its examples use the dynamic `call` API already available in published 0.1.0, so the two PRs can merge independently. Neither is merged or newly published yet. The complete reviewed feature is preserved on `codex/ptc-complete-backup` locally. No merge is requested yet.

## Review commits in order

| Commit | Review focus |
|---|---|
| 9b8a502ed | Tool callable opt-in, JSON metadata, shared file serialization and cancellation guard |
| 0aaa38fba | Process ending hook, bounded job publication, frozen terminal data, session/wire codecs |
| dbf6526ce | Loopback RPC admission, capabilities, deadlines, trace budgets and cancellation tests |
| d1d27c67d | SDK/mbtx integration, background ownership, source snapshots, agent trace recording |
| 552d36dce | Desktop full/minimal nested calls, live jobs, historical snapshots and browser tests |
| d06e511e7 | System prompt, SDK examples, architecture invariants and review closure |
| ce6f21152 | Evaluation harness and results; read last, these are not runtime inputs |
| dbd0db7c4 | Canonical dynamic call in prompts and host guidance; SDK portions subsequently extracted to #1535 |
| 7b2753fc6 | Complete deprecation-migration script, fixture, run instructions, native/wasm host integration |
| 2449ba9da | Extract SDK changes to #1535; move the example under `agent_tool/ptc/examples` |
| 4977d2270 | Parameterize the example by OLD/NEW; compiler-span matching and native/wasm regressions |

[...]

PR #1542, opened 2026-09-15
feat(agent_tool): serialize file operations through one session gate

bobzhang:
Second slice of #1532 (SDK-based programmatic tool calls), split for review. Independent of #1541.

## What changes

- `FileStateMap` gains a session-wide async mutex and `with_access(op)`, which serializes a complete file operation (validation, write, check, rollback) across async suspension points. After the lock is granted it rechecks cancellation: a release can hand the lock to a waiter that was cancelled before it resumed (regression test in `serialization_wbtest.mbt`).
- `FileStateMap::serialize(definition)` wraps a tool executor in that gate. `build_tools` wraps `edit`, `multi_edit`, `write`, `remove`, and every non-concurrent-safe extra tool.
- mbtx saves a script and reads a saved script under the same gate, so a saved-source read captures a consistent snapshot rather than a half-applied edit or rollback (test in `save_wbtest.mbt`). The gate is released before compilation and before waiting on the child.

## Why now

Today the agent loop runs non-concurrent-safe tools one at a time, so nothing interleaves and the gate is a no-op in practice. The later PTC slice adds a second dispatcher (a loopback listener invoking the same executors) that can run between a direct tool's suspension points; both dispatchers must use these same wrapped definitions.

## Split plan

A #1541 structured results → B (this) shared file gate → C job lifecycle → D desktop nested cards → E PTC service package → F mbtx/agent integration → G prompts and docs.


bobzhang:

### Review (`review_b`)

## Blocking defects

- **[P2] Cancellation becomes a task-group failure** — [serialization.mbt:12–13](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review/agent_tool/serialization.mbt:12).

  The post-acquisition guard calls `fail(...)`, producing an ordinary error. The pinned async runtime exempts only its `Cancelled` error from cancelling sibling tasks ([task_group.mbt:75](/Users/dii/.moon/cache/deps/v1/sources/moonbitlang/async/0.21.1/src/task_group.mbt:75)).

  **Concrete scenario:** use the regression test’s grant-then-cancel sequence with the default `group.spawn(...)` options. When the waiter resumes, this guard fails the entire group and cancels unrelated siblings, although only that waiter was cancelled. Catching `task.wait()` cannot undo the group failure. The test’s [explicit `allow_failure=true`](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review/agent_tool/serialization_wbtest.mbt:13) masks this defect.

  Preserve the runtime’s cancellation error and extend the regression to verify that an independently cancelled waiter leaves its group and siblings running.

## Non-blocking suggestions

None.

Reviewed the full diff and surrounding code. No files modified; no builds or tests run.

VERDICT: REQUEST_CHANGES
### Re-review after fixes (`review_b2`)

## Blocking defects

None found in the full diff.

The cancellation fix is correct: [serialization.mbt:8–17](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_b2/agent_tool/serialization.mbt:8) registers lock release before checking cancellation. In the pinned async runtime, `pause()` raises the runtime’s cancellation error, which avoids failing the waiter’s task group. The sibling-survival test covers that distinction.

The gate covers each wrapped executor through completion, and mbtx releases it before compilation or child execution.

## Non-blocking suggestions

Add a registry-level serialization test for [tool_definition.mbt:174](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_b2/agent/tool_definition.mbt:174). The new tests exercise the gate and snapshot helper; they would still pass if registry wrapping were accidentally removed.

VERDICT: APPROVE

PR #1543, opened 2026-09-15
feat(bgjobs): explicit job finish sequence, ending hook, bounded publication

bobzhang:
Third slice of #1532 (SDK-based programmatic tool calls), split for review. Independent of #1541 and #1542. This is the one slice that changes behavior for every background job, so it is isolated here.

## What changes

- `ShellExecution::on_ending(callback)`: a synchronous hook fired at stop initiation or the moment the direct child's exit is observed, before output draining. A late registration fires immediately. The later PTC slice uses it to revoke a script's capability before any suspension.
- Background jobs get an explicit finish sequence, `JobPhase` (Live, Finishing, CleanedUp, Published), replacing three booleans. `finish_job` runs launcher cleanup exactly once and then publishes one terminal record.
- **Behavior change:** a job whose process has exited reports `Running` (snapshot) / `Stopping` (record) until launcher cleanup has completed, so the terminal record is genuinely final and a cleanup failure is part of it. Only mbtx registers cleanup today (deleting the run's build directory), so the window is that deletion; shell jobs have no cleanup and no visible window. `job_wait` and the turn's finish check are unchanged (they already keyed on publication).
- Cleanup failures become visible: `BgJobSnapshot.cleanup_error`, the background notice says "cleanup failed: …", and `job_output` marks the result as an error and prints the reason. Note that the only cleanup callback registered today (mbtx deleting its build directory) still swallows its own errors, so nothing populates this yet; the PTC slice adds the first callback that reports.
- **Behavior change:** job storage writes for Started/Updated events get a five-second deadline so a slow disk cannot stall every job's publication behind the lock; a timeout releases the lock and surfaces as `persistence_error`. The terminal Finished/Failed write is exempt: it happens once and a restart reads it, so it waits out a slow disk instead of being lost (tests cover both).
- `publish` refuses to proceed for a caller that was cancelled while waiting on the publication lock, raising the runtime's own cancellation error so the caller's task group sees a cancellation, not a failure. Updated events after completion are dropped.
- `BgJobRuntime::metadata_changed(id)` publishes an Updated event on request (used by the later slice for live trace updates).

## Split plan

A #1541 structured results → B #1542 shared file gate → C (this) job lifecycle → D desktop nested cards → E PTC service package → F mbtx/agent integration → G prompts and docs.


bobzhang:

### Review (`review_c`)

## Blocking defects

- **[P2] Session cancellation bypasses the early ending hook.** In [execution.mbt:220](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review/agent_tool/shell_exec/execution.mbt:220), cancellation raises before `begin_ending()`. If the session shuts down while the output reader is awaiting a spill-file write, the process is cancelled independently, but the monitor’s task group waits for that reader before reaching its outer `set_terminal` cleanup. The reader’s `request_stop` defer also waits for that write to settle. Consequently, `on_ending` can remain unfired throughout output cleanup, violating its new before-draining guarantee. Invoke it on the cancellation path before joining the reader, and cover this ordering in a regression test.

## Non-blocking suggestions

- [mbtx.mbt:869](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review/agent_tool/mbtx/mbtx.mbt:869) still discards directory-deletion errors. Although unchanged here, this means the only production cleanup callback cannot populate the newly exposed `cleanup_error`. Propagate those errors or narrow the PR’s cleanup-reporting claim.

VERDICT: REQUEST_CHANGES
### Re-review after fixes (`review_c2`)

### Blocking defects

None found in the full diff. The prior cancellation-path defect is fixed: [execution.mbt:224](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_c2/agent_tool/shell_exec/execution.mbt:224) calls `begin_ending()` before closing or joining the reader.

### Non-blocking suggestions

- Strengthen the regression test at [lifecycle_wbtest.mbt:69](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_c2/agent_tool/shell_exec/lifecycle_wbtest.mbt:69). Observing `Running` does not prove the hook preceded reader cleanup: the old `set_terminal` fallback also invokes the hook before changing status. Gate reader cleanup and assert the hook fires before releasing that gate.

Read-only review; no builds or tests run.

VERDICT: APPROVE

PR #1545, opened 2026-09-15
feat(ptc): bounded session-owned loopback service for programmatic tool calls

bobzhang:
Fifth slice of #1532 (SDK-based programmatic tool calls), split for review. Depends on #1541 and #1542 (both merged). This adds the `agent_tool/ptc` package and its tests only; nothing wires it into a host yet, so behavior on main is unchanged.

## What changes

- **`Service`**: one session-owned loopback HTTP listener (`127.0.0.1:0`) started on the session task group. It serves only tools that opted in with `program_callable`, filtered through `Tools::program_tools()`, and invokes their executors directly. It never dispatches through the agent loop, which is what rules out the nested-dispatch deadlock.
- **`Program`**: one registration per script with a random 32-byte bearer capability handed to the child as `OPENSEEK_PTC`, its own active-call set, and a bounded trace. `stop` revokes admission and cancels active calls synchronously; `close` joins them and publishes the final trace. `handoff(job_id, observer)` always detaches the foreground observer, installing a no-op if the registration is already closed (regression test: a process that exits during adoption must not publish into the foreground slot).
- **Admission** reads one bounded HTTP/1.1 POST directly from the socket: 8 KiB request line plus headers, 64 KiB body, 128 KiB total, Content-Length or plain chunked, no duplicate headers, no browser origin, within five seconds of accept. Nothing is reserved until authentication and framing succeed. The async library's server parser reads headers unbounded, which is why this is hand-rolled.
- **Budgets**: 64 registrations per session, 16 connections, 64 calls and 4 active per script, 120 s executor deadline (revoked at expiry before the waiter unwinds), 64K-character replies, 512 KiB retained trace with per-entry reservation before execution. All bounds are named constants.
- Stateful tools serialize across scripts through a service mutex; every executor is also expected to be gate-wrapped by the host (#1542), so both dispatchers share one file gate.
- The tool description and README use the SDK's `@tools.call(name, arguments)` API, which is the only entry point in SDK 0.2.0 (unreleased on main) and also works with published 0.1.0.

## Split plan

A #1541 → B #1542 → C #1543 → D #1544 → E (this) → F mbtx/agent integration → G prompts and docs.


bobzhang:

### Review (`review_e`)

## Blocking defects

1. **[P2] Timeout publication can outlive `Program.close` and overwrite the final trace.**  
   [server.mbt:213](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_e/agent_tool/ptc/server.mbt:213) publishes from the HTTP handler after the executor has terminated. [server.mbt:345](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_e/agent_tool/ptc/server.mbt:345) only joins executor tasks. Concrete sequence: call A times out and its observer suspends with a snapshot showing B running; `close()` cancels B, publishes B interrupted, and returns; A’s suspended observer then saves its older snapshot. Clearing `on_trace` does not cancel an already-running callback. Finalization must join pending publications and ensure the final snapshot is published last.

2. **[P2] `Program.result` silently discards existing structured output.**  
   [server.mbt:409](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_e/agent_tool/ptc/server.mbt:409) replaces `output.data` with either `None` or the PTC trace. For example, passing `respond("ok", data={"artifact": "report.txt"})` through a registration that made zero calls loses the artifact metadata. Preserve existing data when no trace exists, and define how trace metadata combines with existing output when calls occur.

## Non-blocking suggestions

- [README.mbt.md:3](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_e/agent_tool/ptc/README.mbt.md:3) describes default activation, host integration, and Desktop/job behavior as already available. Mark those sections as planned integration so this slice’s documentation remains accurate if subsequent slices never land.

VERDICT: REQUEST_CHANGES
### Re-review after fixes (`review_e2`)

## Blocking defects

None found in the full diff.

Both previous fixes look correct:

- Publication snapshots are taken under the mutex. `close` joins executors **before** acquiring that mutex, then publishes and installs the no-op observer. This prevents stale delivery without creating the executor/close deadlock.
- `attach_trace` preserves existing data when no calls occurred, merges object fields without mutating the original map, and retains non-object data under `output`.

The slice introduces no host wiring.

## Non-blocking suggestions

- Add focused regression tests for [`Program::attach_trace`](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_e2/agent_tool/ptc/server.mbt:427): no-call passthrough, object merging, non-object preservation, and trace-key collisions. The structured-data fix currently lacks direct coverage.

VERDICT: APPROVE

PR #1547, opened 2026-09-15
test(eval): add the PTC capability and YAML parser evaluation harness

bobzhang:
Evaluation slice of #1532 (SDK-based programmatic tool calls). Independent of the runtime slices; nothing here is a production input.

## What changes

- `eval/ptc_prompt/`: a small standard-library Python runner that stages byte-exact fixtures, runs the real OpenSeek CLI, and scores nested PTC metadata; a YAML-parser A/B benchmark; the baseline and candidate prompt variants; and 13 unit tests for the harness (`python3 -m unittest test_run test_yaml_benchmark`).
- The markdown summaries of the 2026-09-15 capability and YAML runs. The raw per-trial JSON reports (about 2,000 lines) are deliberately not committed; the summaries carry the numbers and the harness regenerates the JSON.
- `.gitignore` whitelist entry for the new eval directory, and a line in `eval/README.md`.

The recorded runs used SDK 0.1.0's `@tools.multi_edit` wrapper; SDK 0.2.0 on main exposes only `@tools.call`, and the summaries describe what happened rather than what to write.


bobzhang:

### Review (`review_h`)

## Blocking defects

1. **[P1] Grading executes the candidate’s build configuration with credentials.**  
   [yaml_benchmark.py:154](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_h/eval/ptc_prompt/yaml_benchmark.py:154) copies `moon.pkg` unchanged, then runs `moon test` with `os.environ.copy()`. Package-level `rule`/`dev_build` entries execute shell commands before testing. A candidate manifest containing an environment-dumping build rule can therefore write `DEEPSEEK` into the grading log; a rule can also overwrite the oracle. This contradicts the claimed grading isolation. Generate a trusted package manifest from validated core imports and exclude credentials from the grading environment.

2. **[P2] The documented historical comparison crashes on this checkout.**  
   [run.py:210](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_h/eval/ptc_prompt/run.py:210) requires `### Programmatic tool calls`, but the current `prompt/generated_default_prompt.mbt` contains no such heading. The documented `--prompt-ab --engine …` invocation raises `ValueError` before launching trials, even with the correct historical binary: prompt extraction reads this checkout, not that binary. Supply an explicit historical base prompt or otherwise make this mode self-contained.

3. **[P2] A successful trial with a review child is scored as failed.**  
   [run.py:76](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_h/eval/ptc_prompt/run.py:76) calls `session_items(workspace)` without the parent session name. That reader requires exactly one matching journal. If the model invokes a review subrun and subsequently finishes with correct files, both parent and child journals match; analysis raises `ValueError`, and `trial()` records failure. `--analyze-only` instead aborts on the same exception. Pass the known parent session name, as the YAML runner already does.

4. **[P2] YAML fixture preservation is not byte-exact.**  
   [yaml_benchmark.py:167](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_h/eval/ptc_prompt/yaml_benchmark.py:167) compares protected files using `read_text()`, which normalizes CRLF to LF. A candidate that changes only their line endings receives `preserved_fixture: true` and can pass the overall oracle despite changing protected bytes. Use `read_bytes()` against encoded expectations to support the report’s “protected bytes preserved” claim.

## Non-blocking suggestions

- Update references to the “accompanying JSON ledger” and “committed ledger” to reflect that those artifacts are intentionally uncommitted, and identify the actual regeneration outputs.

VERDICT: REQUEST_CHANGES

bobzhang:
## Codex CLI re-reviews after fixes (`codex exec`, high reasoning, read-only sandbox)

### Re-review after the first fix (`review_h2`)

## Blocking defects

1. **[P2] Manifest reconstruction drops valid core imports.**  
   [yaml_benchmark.py:159](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_h2/eval/ptc_prompt/yaml_benchmark.py:159) matches entire lines, so valid syntax such as `import { "moonbitlang/core/strconv" @num }` produces an empty grading manifest. An import with a trailing `// comment` is also discarded. A correct parser using `@num` therefore fails compilation only in grading. Parse the import syntax while continuing to exclude executable manifest settings.

2. **[P2] Deleting a protected fixture aborts the benchmark instead of recording failure.**  
   [yaml_benchmark.py:196](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_h2/eval/ptc_prompt/yaml_benchmark.py:196) reads protected files without handling absence. If a candidate deletes `visible_test.mbt` while leaving `moon.mod` intact, grading raises `FileNotFoundError`. That escapes through `analyze_trial` and `future.result()`, preventing the trial summary and aggregate `results.json` from being written; reanalysis fails identically. Record missing protected files as preservation failures, and retain other grading errors explicitly.

## Non-blocking suggestions

- Update the historical README command to include `--base-prompt`. The new diagnostic fixes the crash, but the documented command still fails on this checkout.
- Correct the ledger references: hashes are written to `manifest.json`; the summaries still direct readers to `results.json`.

The named-parent selection and byte-comparison fixes are present. Grading now excludes candidate build rules and strips the specified credential-like environment variables.

The permitted Python tests were attempted with bytecode writes disabled, but all were blocked during setup by the sandbox’s prohibition on temporary-directory creation. No builds or MoonBit tests were run.

VERDICT: REQUEST_CHANGES
### Re-review after the second fix (`review_h3`)

## Blocking defects

None found. Both previous fixes are correct: imports are collected across one-line blocks after stripping `//` comments, and deleted protected fixtures fail preservation checks without raising `FileNotFoundError`.

## Non-blocking suggestions

- `eval/ptc_prompt/results-2026-09-15.md:42` still claims the runner regenerates `results-2026-09-15.json`. Update this remaining reference to `manifest.json` and `results.json`.

Reviewed the full diff and relevant surrounding code. No files modified; no builds or tests run.

VERDICT: APPROVE

PR #1549, opened 2026-09-15
feat(mbtx): enable SDK tool calls through the session PTC service

bobzhang:
Sixth slice of #1532 (SDK-based programmatic tool calls): the host integration. Depends on #1541, #1542, #1543, and #1545 (all merged). This is the slice that turns the feature on.

## What changes

- **mbtx `ptc` argument.** Advertised only when the registry wires a PTC service. Defaults to true for a plain wasm run; explicit `true` with `escalated`, `subrun`, or another target is an error; `false` opts out. A saved script's retry hint carries the choice.
- **Registration lifetime.** mbtx registers a `Program` per run and passes its `OPENSEEK_PTC` capability into the moonrun policy environment; runs without PTC get `{"available": false}` so a child never inherits a parent's capability. The execution's `on_ending` hook stops the program synchronously at exit or stop; every exit path closes it. The foreground path also stops it from the process cancel handler and on timeout.
- **Background handoff.** The adopted job takes over the registration: trace updates publish as job metadata through `metadata_changed`, `job_output` returns the trace as `data`, and the job's terminal cleanup closes the program before deleting scratch space, then fails the job if the trace publication failed or the process exited 0 with unfinished calls. Completed jobs freeze their final data.
- **Agent loop.** `build_tools` starts the session service over the gate-wrapped file tools and stateful extras, and keeps a per-session trace slot that the loop drains after each call; on cancellation it records the partial trace as an error result so mutations are never lost from the transcript. `Program::result` attaches the trace to normal results.
- **Tests.** Real published-SDK (`bobzhang/openseek_tools@0.1.0`, called through `@tools.call`) scripts on the wasm host: foreground edit, structured search results, no job runtime, foreground timeout, unchanged script source; background handoff with persisted traces, stop joining an active call, opt-out, stale edits after foreground edits, exit-zero with an unfinished call, handoff failure; agent-level nested-call storage and cancellation persistence.

## Not in this slice

Prompt guidance (G) follows. Default-on is the one-line policy in `definition`; the per-request prompt cost discussed on #1532 is the reason to keep that decision visible.


bobzhang:

Independent review run by Claude Code on the branch head against `origin/main`. Transcript posted verbatim; the fix commit that follows addresses both findings.

### Review (`review_f`)

## Blocking defects

1. **[P1] Cancellation can discard a completed foreground mutation’s trace.**  
   [agent/turn_loop.mbt:719](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_f/agent/turn_loop.mbt:719) clears the trace slot before the awaited `record_tool_result` at line 853. The recovery `errdefer` only covers execution, ending at line 717. If a PTC edit completes and cancellation arrives while `SessionStore::append` is acquiring its lock, the file remains changed but neither the normal result nor a recovery trace is persisted. Keep recovery active through the durable append, and clear the slot only after it succeeds.

2. **[P2] Metadata persistence failures do not trigger the promised PTC job failure.**  
   [agent_tool/mbtx/mbtx.mbt:982](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_f/agent_tool/mbtx/mbtx.mbt:982) forwards trace publication to `metadata_changed`, but [bgjobs/records.mbt:156](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_f/agent_tool/bgjobs/records.mbt:156) catches storage errors and returns normally. Consequently, `Program::trace_error()` stays empty and the cleanup check at mbtx line 951 passes. For example, if the metadata directory becomes unwritable after adoption, a background edit can complete and exit zero with its trace unsaved, yet emit `Finished` and return successful `job_output`. Propagate publication failure to the program or incorporate it into terminal failure classification.

## Non-blocking suggestions

- [agent_tool/mbtx/README.mbt.md:434](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_f/agent_tool/mbtx/README.mbt.md:434) advertises bundled `tools.edit`/`tools.multi_edit` helpers; align it with the explicit SDK import and `@tools.call` API described by the tool.

Read-only review; no builds or tests run.

VERDICT: REQUEST_CHANGES

bobzhang:
## Codex CLI re-review after the first fix (`review_f2`)

## Blocking defects

**[P2] Handoff’s explicit trace publication can fail without failing the job** — [mbtx.mbt:982–985](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_f2/agent_tool/mbtx/mbtx.mbt:982).

Concrete scenario: a script completes its calls and exits zero while `adopt` is awaiting persistence. The exit hook stops the program, so `Program::handoff` installs a no-op observer ([server.mbt:325](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_f2/agent_tool/ptc/server.mbt:325)). If the explicit `metadata_changed(id)` then fails—for example, the record destination is unwritable—that exception bypasses `Program::deliver_trace` and never sets `trace_error`.

Terminal cleanup consequently sees neither a trace error nor unfinished calls. `finish_job` selects `Finished` ([records.mbt:205–216](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_f2/agent_tool/bgjobs/records.mbt:205)), and `job_output` can report `is_error=false`. The foreground mbtx response reports the failure, but the adopted job incorrectly reports success.

Make this catch-up publication’s failure part of the job’s terminal outcome, including when exit races adoption.

## Fix verification and non-blocking suggestions

- **Previous P1 is fixed:** recovery encloses execution and result recording; the later `errdefer` runs before the clearing `defer`, consistent with [MoonBit’s cleanup ordering](https://docs.moonbitlang.com/en/latest/language/fundamentals.html#defer-and-errdefer-expressions).
- **Previous P2’s observer path is fixed**, but the handoff path above remains.
- Add a regression that cancels during result append after a completed mutation. The current agent cancellation test cancels an active call.
- README wording is corrected. No files changed; no builds or tests run.

VERDICT: REQUEST_CHANGES

bobzhang:
## Codex CLI re-review after the second fix (`review_f3`)

## Blocking defects

None found. The `republish` fix addresses the reported race: publication failures remain in `trace_error`, and terminal cleanup waits for publication before checking that error. Already-ended programs retain their final trace through the job’s data supplier.

## Non-blocking suggestions

- Add a regression test for catch-up publication failure concurrent with process exit at `agent_tool/mbtx/mbtx.mbt:988`. Existing tests cover publication failure and handoff separately.
- Update `agent_tool/ptc/README.mbt.md:3–5`; it still describes this integration as forthcoming.

Reviewed the full diff and surrounding lifecycle code. No files modified; no builds or tests run.

VERDICT: APPROVE

PR #1551, opened 2026-09-15
chore: bring every manifest to a moonc 0.10.13 clean check

bobzhang:
Unblocks CI. The MoonBit stable toolchain released today (moon 0.1.20260915, moonc v0.10.13) adds an `unused_package` warning and no longer needs a package imported to resolve its enum constructors by type; CI runs `moon check --deny-warn`, so main's three check jobs fail on every target (see the run on 73d2b4fc7).

## What changes

55 `moon.pkg` manifests, no `.mbt` code:

- Imports the new compiler reports unused are removed from the block they were in.
- Where a removed main-block import is still used by that package's `*_test.mbt` or `*_wbtest.mbt` files (for example `moonbitlang/core/debug` in `desktop/protocol`, `inspect`, `desktop/frontend/mention_context`, and `moonbitlang/editor/viewer/common/agent_feedback_api` in the agent-feedback browser package), it moves into the matching `for "test"` / `for "wbtest"` block instead of disappearing.
- Five packages imported `bobzhang/openseek_protocol` only for unqualified constructors passed to `@emit.emit`; the new compiler resolves those by type and flags the import unused, while qualifying them trips `unnecessary_annotation` (which those packages opt into). The import is dropped, which matches the compiler CI runs.

## Verification

Done twice, because the two compilers disagree on exactly those five imports:

- **Today's stable (what CI runs), installed into an isolated `HOME`:** `moon check --deny-warn` on native, js, and wasm all clean; `moon fmt --check` clean; `moon info` produces no interface drift; full JS suite 3,269/3,269; full native suite 3,166/3,167, the one failure being the live DeepSeek web-search test that needs network access.
- **Previous stable (0.10.12):** js check and the full JS suite pass; native/wasm checks report the five dropped protocol imports as required, because that compiler cannot resolve the constructors without them. After this lands, a local toolchain older than 0.10.13 will need updating, which is the repo's rolling-latest policy already.

The manifest edits were generated by iterating the new compiler's diagnostics to a fixed point (script kept out of the repo), then reviewed.


bobzhang:
## Codex CLI reviews (`codex exec`, high reasoning, read-only sandbox)

### First pass, on the ten-import commit (`review_chore`)

1. **Blocking defects:** None. Reviewed the full ten-line diff and all affected packages’ `.mbt` files, including whitebox tests. No removed alias is needed in its import scope. `agent_subrun/host` and `protocol/emit` retain separate test imports for their `@json` references. Remaining import entries are independent of the removed entries.

2. **Non-blocking suggestions:** None.

No files changed; no builds or tests run.

VERDICT: APPROVE
### Second pass, on the full 55-manifest change (`review_chore2`)

## Blocking defects

None found. Across all 55 changed manifests:

- Removed main imports have no remaining alias references in non-test source.
- Moved imports match the consuming test files.
- No test references are stranded by the removals.
- Protocol constructors have expected types supplied by `emit` or the matched value, consistent with the stated moonc 0.10.13 behavior.

No builds or tests were run.

## Non-blocking suggestions

Update the PR description’s manifest count from **45 to 55**.

VERDICT: APPROVE

PR #1553, opened 2026-09-15
refactor(desktop): use Rabbita DOM wrappers instead of redundant JS shims

tonyfettes:
## Summary

Seven `extern "js"` shims in the browser frontend duplicated bindings the pinned Rabbita release already exposes, and the transcript image observer kept a private JS type for work `@dom` can do. Every replacement keeps the previous behaviour; each removed JS body was checked against the pinned release with `moon ide doc` before deletion.

## What changed

- **jobs**: `scroll_end` / `away_from_end` read the output pane through `Document::get_element_by_id` and the `IsElement` scroll + client metrics.
- **project_picker**: `scroll_path_row` uses `scroll_into_view_with_options(block: "nearest")`, the same call `dom_helpers.mbt` already uses.
- **skills**: `copy_content` uses `@clipboard.copy` instead of a hand-rolled `navigator.clipboard.writeText` (a rejected write was already silent, and `@clipboard.copy` without callbacks maps both outcomes to `@cmd.none`).
- **transcript images**: `TranscriptImageObserver` is gone. The observer is now a `@dom.IntersectionObserver`; only its construction stays JS, because the callback reads `isIntersecting` / `target.dataset` and `IntersectionObserverEntry` has no `@dom` accessors in 0.15.7. Scanning uses `query_selector_all` and disposal uses `IntersectionObserver::disconnect`.

`extern "js"` sites in `desktop/frontend`: 33 -> 27.

## Validation

- `moon check --target js` (desktop/frontend): clean.
- `moon test --target js` (desktop/frontend): 3261/3261 passed.
- `moon fmt --check`: clean. `moon info`: no public-interface change (only private functions and a private struct field moved).
- `moon check --target js --deny-warn`: still the 10 pre-existing unused-package warnings, none in a touched package.
- Emitted bundle inspected by hand: the replaced helpers now compile to the Rabbita wrappers (for example `set_scroll_top(el, get_scroll_height(el))`, and `NodeList::iter(query_selector_all(...))` + `IntersectionObserver::observe`), and the old observer `scan` body is gone.

## Not verified

Browser behaviour of the three changed surfaces specifically. CI's `desktop e2e` (Playwright against Chromium) passes, so the app still boots, streams and renders, but no spec asserts the jobs output pane follow, the picker row reveal, or the skills copy button — those DOM effects are checked only at the type and emitted-JS level.

## Follow-ups

The remaining 27 sites are either genuinely missing bindings (Selection/Range, `classList`, `input.select` / `setSelectionRange`, `crypto.randomUUID`, Intl date formatting, `fetch` status) or imperative glue the declarative model cannot express (layout geometry, promise ordering, capture-phase listeners). The next step is a single `desktop/frontend/internal/browser` package that owns all of them, plus a guard that fails when `extern "js"` appears elsewhere.

Generated with SeekMoon


tonyfettes:
@codex review

PR #1560, opened 2026-09-15
simplify(ptc): Program back-reference, trace-entry builder, shared test helpers

bobzhang:
Post-landing simplification sweep, slice 1 of the programmatic-tool-call work: the `agent_tool/ptc` package. Behavior-neutral; the existing 25 async tests pass with no assertion changed. Two commits so the production change can be reviewed on its own.

## Commit 1: production (`server.mbt`, −7 lines net)

- `Program` held five things copied from its `Service` (the stateful-call mutex, both deadlines, a `revoke` closure over the registry, and a pre-built environment map). It now holds `service` and `token`, derives the SDK handoff on demand in `env()`, and `Program::execute` no longer takes the tool registry it can reach through the service. The mutex is the same service-wide object, reached through the back-reference.
- Five hand-built trace-entry objects become one `call_entry(name, arguments, status, result?)`.
- The identical cancel-and-join pair in the request handler (errdefer and the deadline catch) is one local closure.
- `Program::result` computes `interrupted()` once.

## Commit 2: tests and README (−62 lines net)

- Whitebox test helpers shared by both files: `leaf(name~, execute~, concurrent_safe?)` replaces 21 six-line tool-definition literals whose description was never asserted on; `capability(env)` replaces the four-line handoff guard; `settle(task)` and `await_admission(program)` replace the join and polling idioms.
- The README no longer restates the model-facing `description()` text (activation rule, `CallResult` shape, never-retry rule, search data shape, printed-output rule); it points at `description()` as the script-facing contract and keeps what the description lacks: hosting, lifetime, wire bounds, and the runnable examples.

## Not done, deliberately

The review of #1532 rejected removing the hand-written bounded HTTP parser, the service-wide mutex, the pull-style job data supplier, and the per-program publication mutex; none of those are touched.

Verified with moonc 0.10.13: `moon check --deny-warn` on native, js, and wasm; `moon fmt --check`; `moon info` with no interface drift; ptc, mbtx, and agent native tests (267); full JS suite (3,271).


bobzhang:

### Review (`review_s1`)

## Blocking defects

None found. No concrete output, ordering, cancellation, or lock-scope regression in this slice.

## Non-blocking observations

- **Extra work on control results:** [server.mbt:415](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_s1/agent_tool/ptc/server.mbt:415) now scans `calls` even for `Control(_)`, which previously returned immediately. Move the cached `interrupted` binding inside `Respond`. Output remains unchanged.
- **Test coverage shifts slightly:** [server_wbtest.mbt:28](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_s1/agent_tool/ptc/server_wbtest.mbt:28) supplies `concurrent_safe=false` explicitly through `leaf`; converted tests no longer exercise the constructor’s default. Assertions and current flag values are preserved. `capability` also replaces distinct failure messages (“first capability”, “second capability”) with one generic diagnostic.
- **Missed straightforward cleanup:** [server_wbtest.mbt:86](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_s1/agent_tool/ptc/server_wbtest.mbt:86) still contains a definition that fits `leaf(name=…, concurrent_safe~, execute=…)`.

## Simplification assessment

- **Service back-reference:** real reduction in duplicated fields and plumbing. The previous `revoke` closure already captured `self`; this does not establish a newly introduced service-retention bug.
- **`env()`:** moves construction from registration to each invocation, adding repeated JSON serialization. It still returns a fresh map with the same handoff values.
- **`call_entry` / `abandon`:** useful consolidation; field order, optional-result omission, cancellation-before-join, and cancellation protection are preserved.
- **Test helpers / README:** genuine duplication reduction. The removed script-facing guidance remains in `description()`; no assertion was removed.

Reviewed the full diff and surrounding code read-only. No builds or tests run.

VERDICT: APPROVE

PR #1561, opened 2026-09-15
simplify(mbtx): derive the program environment, one retry hint, aligned ptc decoding

bobzhang:
Post-landing simplification sweep, slice 2: the mbtx and agent integration of programmatic tool calls. Small, behavior-neutral changes plus one test the earlier slices left uncovered.

## What changes

- `mbtx.execute` loses its `program_env?` parameter: it was always the program's own `env()`, so the injection site derives it from `program` and the `run` closure in `definition` takes one argument.
- `on_program_trace?` has no default and passes straight through to `Service::register`, which already has one.
- The saved-script retry hint is built once as a map with the `ptc` key added whenever PTC ran or was explicitly declined (so a re-run lands in the same mode), instead of three JSON literals.
- The two trailing `program is Some(program) && …` guards in the job's terminal cleanup are one block.
- `ptc` decoding treats an explicit `null` as unset, like every other optional mbtx argument (`None | Some(Null) => None`). Only observable change: `"ptc": null` is now accepted (pinned by the new test) as "unset" instead of rejected, consistent with every other optional mbtx argument.
- `agent/ptc_test.mbt` reuses the package's `wait_call_payload` instead of a 25-line copy of it. The copied `usage` block was asserted on by nothing; the shared payload also carries `"type": "function"` and a different call id, neither of which the two tests observe.
- New test: an explicit `ptc: true` against a registry without a PTC service is refused, and omitted or `false` runs the plain path with no trace. The survey found this the one branch of `definition` no test reached.

Verified with moonc 0.10.13: `moon check --deny-warn` on native, js, and wasm; `moon fmt --check`; `moon info` with no interface drift; mbtx, decode, and agent native tests (256); full JS suite (3,271).


bobzhang:

### Review (`review_s2`)

## Blocking defect

**[P1] Revocation now happens after asynchronous output reading.** At [mbtx.mbt:1073](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_s2/agent_tool/mbtx/mbtx.mbt:1073), removing the explicit `program.stop()` leaves the capability active until `read_capped(log)` finishes. That function suspends on file open/read ([line 1606](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_s2/agent_tool/mbtx/mbtx.mbt:1606)). `Program::stop` explicitly requires revocation at process exit and marks outstanding calls interrupted ([server.mbt:378](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_s2/agent_tool/ptc/server.mbt:378)).

Concrete scenario: a runtime-less script exits successfully with an accepted, unfinished mutation. While mbtx reads its log, that executor resumes, mutates a file, and records `done`. Previously, the immediate stop cancelled it and marked it `interrupted`; now the result can report success without the unfinished-call warning. Client disconnect does not cancel accepted execution, as [service_wbtest.mbt:334](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_s2/agent_tool/ptc/service_wbtest.mbt:334) explicitly verifies.

Restore the immediate stop, or narrow the defer’s scope to finish before reading output.

## Non-blocking suggestions

- **Correct the retry comment and PR description.** [mbtx.mbt:229](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_s2/agent_tool/mbtx/mbtx.mbt:229) says defaulted choices stay implicit, but omitted `ptc` with a compatible service produces `"ptc": true`. This preserves existing behavior; the explanation is inaccurate.
- **Pin null decoding.** Add a direct assertion that `ptc: null` yields `None`. The new integration test does not cover this intentional behavior change.
- **Account for fixture changes.** [agent/ptc_test.mbt:4](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_s2/agent/ptc_test.mbt:4) also changes the ID from `program` to `wait-call` and adds `"type":"function"`. Besides dropping usage, these tests stop exercising tool-call chunks without `type`.

## Simplification assessment

The environment parameter removal, callback forwarding, merged guards, retry map, and shared fixture genuinely reduce duplication. The stop removal breaks ordering. The decoder rewrite adds an arm; a smaller alternative is retaining `fields.get("ptc")` and combining `None | Some(Null) => None`.

VERDICT: REQUEST_CHANGES

bobzhang:
## Codex CLI re-review after fixes (`review_s2b`)

### Blocking defects

None found in the full diff. The previous P1 is fixed: `agent_tool/mbtx/mbtx.mbt:1078–1081` stops the program before reading the log; the defer remains for exceptional exits. The remaining changes preserve behavior apart from the documented acceptance of `ptc: null`.

### Non-blocking suggestions

- Add a regression test for normal foreground exit without a job runtime while an RPC remains active. Existing tests cover foreground timeout and background exit, but not this specific stop-before-log-read ordering.

VERDICT: APPROVE

PR #1562, opened 2026-09-15
simplify(desktop): one trace origin, one result row call, one job status

bobzhang:
Post-landing simplification sweep, slice 3: desktop rendering of nested program calls. Behavior-neutral; the 574 desktop JS tests and the full JS suite pass unchanged.

## What changes

- `program_calls` took two booleans (`terminal?`, `snapshot?`) that encode a three-way choice and had one meaningless combination. It now takes `origin? : TraceOrigin` (`Live`, `Terminal`, `Recorded`) and the snapshot note is computed once instead of per running entry.
- `tool_result_view` had three arms each calling `result_row_view` with nine arguments; it now matches the outcome into the four values that differ and makes one call.
- The jobs panel called `status_label` and `status_class` five times per card, each passing the cleanup error separately. One `job_status(job)` returns the label and class together, with the cleanup override applied once.

Verified with moonc 0.10.13: `moon check --deny-warn` on native, js, and wasm; `moon fmt --check`; `moon info` (two interface files updated for the new enum and parameter); full JS suite (3,272).


bobzhang:

### Review (`review_s3`)

## Blocking defects

None found. All callers preserve the previous live/terminal/recorded mapping. Result-row arguments, ordering, error text, and cleanup overrides remain equivalent. Existing test assertions are unchanged. This slice is self-contained.

## Non-blocking suggestions

- [program_calls.mbt:23](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_s3/desktop/frontend/transcript/program_calls.mbt:23): `snapshot_note` is now evaluated even for live, terminal, empty, or entirely completed traces. With `ptc_job_id`, this constructs an unused interpolated string. Consider computing it lazily once when a recorded running entry needs it. This changes computation cost, not rendered output.
- [jobs/view.mbt:8](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_s3/desktop/frontend/jobs/view.mbt:8): `job_status` still matches `job.state` twice. A single match returning `(label, class)` would complete the simplification and keep each state’s presentation together.

## Simplification assessment

`TraceOrigin` genuinely reduces representable combinations; the shared `result_row_view` call removes duplication without changing arguments. `job_status` reduces repeated calls and cleanup checks, although it retains duplicated state classification.

Read-only review of the full diff and surrounding code; no builds or tests run.

VERDICT: APPROVE

PR #1565, opened 2026-09-15
prompt: keep only PTC policy in the system prompt; point at the tool description

bobzhang:
## Summary

Proposal PR (prompt wording is the owner's call; tweak freely). The system prompt's `### Programmatic tool calls` section restated most of the `mbtx` tool description, which is appended to every request whenever PTC is available: the SDK import block, the enumerated `@tools.call` forms, the "no `await`" note, the search `result.data` shape, the background-handoff/`job_output`/`job_stop` paragraph, and the 64-call/four-in-flight limits. The model saw each of those twice.

This trims the section to what only the system prompt says, in both languages:

- the activation default and `ptc: false` opt-out,
- the policy paragraph (when PTC saves round trips, keep direct tools for simple calls and `multi_edit` for batch validation, separate calls are not one atomic batch, return to the model when the next decision needs judgment),
- the output rule (only printed output reaches the model; print evidence with URLs; keep errors and truncation that affect the answer),
- the re-read-before-retry rule after a transport failure,
- one sentence pointing at the `mbtx` tool description for the import, call API, handoff, and limits.

The section shrinks from 2207 to 1125 characters (English) and from 1096 to 452 characters (Chinese). Generated prompt literals regenerated with `scripts/md_to_mbt_string` for both.

## Why it is safe

- The `### Programmatic tool calls` heading and the following `## Tool Protocol` heading are kept; `eval/ptc_prompt/run.py` splices prompt-A/B variants between exactly those two markers.
- Every dropped fact is still in `agent_tool/ptc/description.mbt`, which reaches the model only when a host can actually serve PTC. The prompt no longer describes an API the model may not have.
- No code change. The `prompt`, `agent`, and `agent_tool/mbtx` native suites and the full JS suite pass; the 15 eval-harness unit tests pass; `moon check --deny-warn` is clean on native/js/wasm; `moon fmt --check` and the generated-prompt drift check are clean.

## Not done here

- No prompt A/B run. The historical `eval/ptc_prompt/baseline.md` and `candidate*.md` fixtures are untouched, so the A/B harness still reproduces the earlier comparison.


bobzhang:

### Review (`review_s5`)

## Blocking defects

**[P2] Restore the instruction to finish calls before exiting.**  
[prompt/default_prompt.mbt.md:474](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_s5/prompt/default_prompt.mbt.md:474) replaces the lifecycle paragraph with a tool-description reference; the Chinese version does likewise at line 284. However, `agent_tool/ptc/description.mbt` never states the deleted requirement: “Complete every call and join any spawned tasks before exiting.”

Concrete failure: a script starts a PTC request with `spawn_bg(..., no_wait=true)` and exits while it remains active. Process exit cancels the request (`agent_tool/ptc/server.mbt:378–400`), potentially after mutations occurred. The existing test at `agent_tool/mbtx/background_ptc_test.mbt:270–332` demonstrates that even exit code 0 produces a failed job with “unfinished tool calls.”

This removes unique operational guidance, contrary to the behavior-neutral claim. Retain the sentence in both prompts, or add it to the shared tool description.

## Non-blocking suggestions

- The remaining deletions are genuine deduplication against the existing tool description. Generated literals reflect the Markdown changes, and eval delimiters remain intact.
- Further optional simplification: replace the inline `@tools.call("multi_edit", …)` example at English line 464 / Chinese line 282 with “batch edits with `multi_edit`,” leaving API syntax in the tool description.

No builds or tests run, as requested.

VERDICT: REQUEST_CHANGES

bobzhang:
## Codex CLI re-review after fixes (`review_s5b`)

## Blocking defects

None found in the full diff.

The previous blocker is fixed: both languages restore completing every call and joining spawned tasks before exit (`prompt/default_prompt.mbt.md:474`, `prompt/default_prompt_zh.md:284`), and both generated literals match. The stated failure consequence agrees with foreground handling (`agent_tool/ptc/server.mbt:413–425`) and background handling (`agent_tool/mbtx/mbtx.mbt:954–959`).

The removed API guidance remains in the conditionally appended tool description. The eval splice markers are unchanged. No concrete regression or dependency on later slices found.

## Non-blocking suggestions

None additional.

VERDICT: APPROVE

PR #1567, opened 2026-09-15
docs(plans): per-edit revert guards and structured feedback for PTC warning fixes

bobzhang:
## Summary

Proposal document only; no code. Revised around the primary use case: a PTC script that fixes most of a project's warnings in one pass and leaves the rest for a manual fix.

- **Unit of work is a single `edit`** with two new post-write guards, `revert_on_errors` and `revert_on_warnings` (both default off). Each fix is compile-checked alone against the tree that already holds the previous kept fixes; a bad fix is restored by the tool and reported exactly. Two `moon check` runs per edit (baseline first, then after the write).
- **"Introduced" is a multiset diff of `(path, code, message)`** before and after, ignoring `loc`, so line shifts do not create false positives and a fix that swaps one warning for another is caught.
- **No seed tool**: the script runs `moon check --output-json` itself (the sandbox's spawn list admits it, as the bundled `check-json.mbtx` shows) and parses the JSON lines; the SDK README example carries the parser.
- **Warning sites are retained** in the tally (capped) and every edit/multi_edit outcome gets a bounded `data` payload; the SDK needs no change.
- **Deferred**: `dry_run` (matching + parse gate), batch-oriented loops on `multi_edit`. **Rejected**: a full compile dry-run that always restores.
- Five-PR split; three decisions listed at the end (flag defaults, `edit` check-line change, raising the 64-call per-script budget).


bobzhang:
## Codex CLI review of the plan (`codex exec`, high reasoning, read-only sandbox)

Independent design review run by Claude Code on the branch head, asked to validate every claim in the plan against the code it names and to find what would make the deprecation-fix script fail on a real project. Transcript posted verbatim.

### Review (`review_plan`)

The plan needs changes before implementation. The main blockers are incorrect occurrence selection, diffing capped diagnostic lists, and treating an unavailable check as successful protection. I treated the call cap as **1024**. No files were modified; no builds or tests were run.

1. **The proposed span-to-`edit` conversion can silently modify the wrong occurrence.**  
   The [plan:15](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/docs/plans/ptc-edit-feedback.md:15) extracts only the diagnostic span and supplies its starting line. But [`edit`:250](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/edit/edit.mbt:250) replaces the **first** matching string from that line onward; it receives no column. For example, a diagnostic on the call in `let s = "length"; xs.length()` produces an edit that changes the string literal first. Neither guard rejects that: it introduces no diagnostic. The existing [preview implementation:330](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/edit/edit.mbt:330) explicitly documents this same-line ambiguity.

   Build an exact whole-line or multiline replacement that changes only the diagnosed columns, with a bounded search range. Also verify that the intended warning disappears; “kept” does not establish that the fix worked.

2. **Capping the tally at 200 sites makes the guard incorrect on its primary workload.**  
   The [plan:54](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/docs/plans/ptc-edit-feedback.md:54) caps stored warning sites before introducing the multiset comparison. With 201 distinct warnings, fixing one of the first 200 brings the previously omitted warning into the retained set. The diff labels that existing warning “introduced” and reverts a valid fix. New warnings outside the retained prefix can conversely escape detection.

   Keep complete identity multiplicities internally and cap only serialized/displayed sites. If the underlying process capture is incomplete, return an explicit inconclusive guard result. Today’s tally already distinguishes capture truncation and retains all captured error sites ([`auto_check`:105](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/internal/auto_check/auto_check.mbt:105), [test:309](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/internal/auto_check/auto_check.mbt:309)).

3. **The multiset rule does not justify “introduced,” or “bad fix identified exactly.”**  
   Ignoring location avoids line-shift false positives, but removing one diagnostic and creating the same `(path, code, message)` elsewhere in that file cancels out. The [existing fixture:275](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/internal/auto_check/auto_check.mbt:275) already contains repeated identical deprecation messages at different locations.

   Conversely, fixing an error can expose existing diagnostics in previously skipped dependents. The plan acknowledges skipped dependents at [153](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/docs/plans/ptc-edit-feedback.md:153), but does not connect that to guard decisions. Existing [`Comparability`:735](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/multi_edit/multi_edit.mbt:735) explicitly treats such attribution as ambiguous; its classifier still uses the old location-based identity at [959](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/multi_edit/multi_edit.mbt:959).

   State the rule’s limitation, define incomplete-build handling, and make guard decisions and comparability reports use consistent evidence. Requiring an error-free baseline is a reasonable smaller scope for the initial warning fixer.

4. **The structured contract lacks essential “not checked / outcome uncertain” states.**  
   [`count_errors`:137](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/internal/auto_check/auto_check.mbt:137) returns `None` for unsupported paths, missing project roots, timeouts, and spawn failures. Existing `multi_edit` deliberately keeps edits unguarded in that case ([587](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/multi_edit/multi_edit.mbt:587)). Moreover, [`count_errors`:155](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/internal/auto_check/auto_check.mbt:155) discards the process exit code: a failed command producing no recognized diagnostic objects becomes a zero-error tally.

   Simply adopting this behavior violates the new promise that guarded edits are checked before being kept. The payload needs check status, completeness, and failure reason; the plan must choose what requested guards do when verification fails.

   Rollback and cancellation need equally explicit treatment. Existing [restore writes:681](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/multi_edit/multi_edit.mbt:681) can fail and propagate cancellation. PTC explicitly returns “changes may have occurred” after a deadline ([server:201](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/ptc/server.mbt:201)). The proposed `edit` payload has no restore-failure field. A script must not classify those cases as ordinary kept/reverted outcomes or retry blindly.

5. **The payload budget does not guarantee delivery of `data`.**  
   The [plan:148](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/docs/plans/ptc-edit-feedback.md:148) budgets data separately while promising complete text. PTC measures the **serialized entire response** and replaces oversized responses with a generic error *after execution* ([server:509](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/ptc/server.mbt:509)). Preview already allows approximately 24K context plus 24K JSON entries ([`edit`:315](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/edit/edit.mbt:315)); duplicating those entries in `data` can exceed 64K. Dropping warnings and all but ten errors also does not bound ten arbitrarily long messages, preview edits, or failure reports.

   Budget the complete serialized envelope, with explicit omission metadata. Request bodies separately have a **65,536-byte** limit ([request:7](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/ptc/request.mbt:7)), relevant to multiline/whole-line replacements.

   Raising the call cap does not make these results small: repeated 24K checks consume roughly 512 KiB in twenty replies, before arguments and other fields. Trace admission includes arguments and reserved headroom ([server:180](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/ptc/server.mbt:180)); results can be omitted from the trace while still reaching the caller ([521](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/ptc/server.mbt:521)). Later calls can be refused before execution. The script needs a budget-exhaustion stop path.

6. **Freshness and check ownership are underspecified.**  
   The [plan:30](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/docs/plans/ptc-edit-feedback.md:30) promises fresh diagnostics for every edit, but the described loop initially seeds once. Either it must rerun its own check after edits—making approximately three checks per attempted fix—or consume the host’s post-edit diagnostics. That second approach must handle truncation and rollback: current revert reports carry the **rejected candidate’s** check, not diagnostics for the restored disk state ([`multi_edit`:615](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/multi_edit/multi_edit.mbt:615)). Name the check’s phase explicitly.

   Serialization within one script is insufficient for a globally stable snapshot. The existing [file gate:2](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/serialization.mbt:2) protects complete host edit/check/rollback operations, but script-spawned checks and filesystem reads do not acquire it. Background handoff leaves the script running ([`mbtx`:682](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/mbtx/mbtx.mbt:682)); other operations can run between its diagnostic collection and edit admission. Preserve the existing transaction gate and validate the expected source snapshot. Holding that gate around the whole script would deadlock nested edits.

The smaller design and factual corrections are:

- **The lazy baseline costs two checks, not three.** Current code checks the candidate, restores, checks the baseline, and optionally reapplies without another check ([`multi_edit`:587](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/multi_edit/multi_edit.mbt:587), [633](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/multi_edit/multi_edit.mbt:633)). An upfront baseline avoids temporary restoration; the claimed check-count saving is false.
- **Use compact edit outcomes for this loop.** Counts, guard status, and bounded introduced diagnostics suffice when the script owns diagnostic collection. Returning the project’s remaining warning list on every edit spends the trace budget repeatedly.
- **Preview already constructs the exact JSON edit objects** ([`edit`:490](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/edit/edit.mbt:490)). Retain those objects for `data`; no new preview planner is needed.
- **The SDK transport and sandbox premises are correct.** `ToolOutput.data`, published-SDK structured-result tests, the `moon check` spawn allowance, and PTC policy injection already exist. However, the current [SDK README:72](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/tools_sdk/README.mbt.md:72) links to an example; it contains no ten-line diagnostic parser.
- **The check-line change loses model-visible warning details.** Current [`append_summary`:23](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/internal/auto_check/auto_check.mbt:23) requests one diagnostic; [`keep_response`:714](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/multi_edit/multi_edit.mbt:714) prints only a warning count when there are no errors. [`ToolOutput.data`:161](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/agent_tool.mbt:161) is not appended to model messages. This is an acknowledged design choice, but unnecessary: render the first warning from the JSON tally without another check.

Implementation also needs these concrete pieces:

- **A diagnostic-to-source contract.** Locations are one-based with an exclusive end ([existing parser:262](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/editor/server/host/moon_check.mbt:262)). The local compiler source computes columns over decoded Unicode code points ([lexer:228](/Users/dii/git/ideas/lib/xml/lex/lex_unicode_lex.ml:228), [960](/Users/dii/git/ideas/lib/xml/lex/lex_unicode_lex.ml:960)), while MoonBit strings use UTF-16 units ([language reference:108](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/share/doc/moonbit/language/fundamentals.md:108)). Direct `column - 1` string slicing therefore fails after astral characters. Cover Unicode, CRLF, multiline and empty spans, and qualified names.
- **Conservative migration recognition.** Deprecation messages may be custom or contain no replacement ([attribute documentation:28](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/share/doc/moonbit/language/attributes/deprecated.md:28)). Use recognized message/source patterns and report the rest. Keep `--deny-warn` as final verification: it promotes warnings to errors ([local Moon implementation:229](/Users/dii/git/moon/crates/moonbuild-rupes-recta/src/build_lower/compiler/build_common.rs:229)), so the proposed `level == "warning"` discovery filter would miss them if discovery used that mode.
- **Shared check scope and result semantics.** Script execution defaults to workspace cwd ([`mbtx`:386](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/mbtx/mbtx.mbt:386)); guards select the nearest project marker ([root lookup:34](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_plan/agent_tool/internal/moon_check/moon_check.mbt:34)). Define root/path normalization and target/options agreement. Also define warning-triggered `multi_edit` reports: the existing threshold, verdict advice, and `reissue_with` logic are expressly error-based, so their text cannot remain unchanged for warning-only rollbacks.

VERDICT: REQUEST_CHANGES

PR #1574, opened 2026-09-16
feat(desktop): filter review files by path and extension

Milky2018:
Review previously listed every changed file with no way to narrow the queue. Add a live path filter and a multi-select file-extension popover above the Changes list. Path matching is case-insensitive and literal, including rename source paths; extension options and counts come from the full changed-file list and include extensionless files.

Both filters apply together to the list and cross-file navigation. Keep the current diff open even if it falls outside the filter, show matching counts and empty results, and clear both conditions with Clear. The extension menu uses the browser's native popover invoker for keyboard activation, Escape, and outside-click dismissal.

Validation:
- `just check`, `just test`, and `just build` passed using isolated MoonBit 0.10.12+ef027fb8a.
- 3187 native tests, 3279 JS tests, and cram/integration checks passed; the file-editor package's 164 tests also passed.
- Browser tests cover path/extension intersection, clearing, empty selections, focus, menu dismissal, and existing Line/Token/Tree navigation.
- Dark/light screenshots and narrow-window menu bounds checked.
- macOS app packaging and deep/strict code-signature verification passed.


Milky2018:
image

Milky2018:
New CSS style:

image

PR #1575, opened 2026-09-16
feat(session): keep each conversation's model in a settings.json sidecar

tonyfettes:
## Why

The model a conversation was last on lived only in the desktop page's localStorage: a 200-entry LRU keyed by channel authority + session id (`openseek.conversation_models`). Reopening a conversation from another device, after clearing site data, or past the cap silently fell back to whatever model an unrelated chat had picked last.

## What

**`agent_session/store`** — the one-value `session-title` sidecar becomes `sessions/<id>/settings.json`, one JSON object whose keys belong to their writers.
- `settings(id) -> Map[String, Json]`, `set_setting(id, key, value)`: assignment of one top-level key under the exclusive session lock, atomic replace; requires an existing record (a session directory is a transcript). No merge semantics anywhere: a key's value is replaced whole, and each key has one owner.
- `append(session, item, settings? = ...)` and `compact(..., settings? = ...)`: the appending process's settings keys are settled in the same locked step as the event. The sidecar is read and merged **before** the event is committed, so an unreadable sidecar fails the append with nothing written; the file is rewritten only when some key differs; keys not named are never touched.
- The store itself understands only `title` (for `sessions list` / TUI picker); `set_title` is a thin wrapper.
- Existing `session-title` files are still read; the first settings write folds the title into the JSON and removes the old file.

**engine (`internal/openseek`)** — every durable append in `serve` (turns, idle items, compaction), `run`, `fleet` and durable subruns passes `@setup.run_settings(model)` (`{"model": <wire name>}`) to the store. A new session's first append — which creates the record — is immediately followed by its settings; a settled session costs no writes; a session another process configured differently is corrected by the process running it. No engine-side state.

**desktop host** — reads only. `session.load` / `session.load_archived` reply with a three-state `ConfiguredModel`: `Configured(wire)`, `Unconfigured`, or `Unreadable(reason)` when the sidecar exists but cannot be read or holds a non-string `model` (logged too; never folded into "unconfigured"). `session.event` broadcasts carry the same value, read from the settings the committing append settled.

**protocol** — `LoadSessionReply` and `SessionCommitPayload` have explicit `FromJson` / `ToJson` (no derived codecs, per `desktop/AGENTS.md`). On the wire: a string `model` is `Configured`, a string `settings_error` is `Unreadable`, neither or `null` is `Unconfigured`; both present, or either of another type, is malformed; unknown fields are ignored; absent fields are omitted.

**frontend**
- `Conversation.model_chosen` says whether the model is the user's own chip pick. A `Started` (live or `agent.runs` resync) and a `session.load` snapshot replace an inherited or reported model whenever nothing is chosen and no run of this page is in flight; a pick is never overwritten by a start or a re-read. A session first met through a `session.event` materializes on the model the commit reports. `Unreadable` keeps the current model and pops a toast when the conversation is opened.
- `conversation_models`, `remember_model` / `remembered_model` / `remembered_with`, `save_conversation_model`, `parse_conversation_models`, the storage key and cap are removed. `default_model` remains for new chats.

## Ownership contract

One writer per top-level key of `settings.json`: the engine owns `model` (and any future run-configuration key it adds to `run_settings`); the desktop owns `title`. Neither reads the other's key before writing its own.

## Behavior change to note

A chip pick on a conversation is durable only once a run uses it. Picking a model and closing the app without sending loses that pick (the old localStorage list kept it).

## Review

- Commit 2 (`ebb87be7e`): settings settle inside the append lock; `Unreadable` as a distinct reply state; explicit `LoadSessionReply` codecs; durable subruns pass settings.
- Commit 3: `model_chosen` replaces the transcript-activity guess (commit-first sessions were pinned to the page default); `session.event` carries `ConfiguredModel`; settings are read before the event commits; compaction settles settings; non-string `model` reads as `Unreadable`.

## Testing

- Root `moon test --target native`: `agent_session/store` (settings settle with the first event, no rewrite when in step, another process corrects `model` and leaves `title`, compaction settles settings, an unreadable sidecar fails the append before anything is committed, shared keys, legacy fold-in, blank/missing rejection, damaged sidecar), `agent_session/compact`, `internal/openseek/setup`, `serve`, `run`, `commands` — 122/122.
- `desktop`: `moon test --target js` 3304/3304 (incl. commit-first materialization on the reported model; a start moves an unchosen conversation but never a chosen one; a re-read keeps a pick; unreadable settings toast); `moon test --target native backend protocol` 81/81 (incl. both payloads' three states, `null`, omission on encode, malformed shapes).
- Not E2E-verified in the running desktop app: run a turn on one device, reopen the conversation from another, and confirm the chip shows that run's model.


agent_session/store/store.mbt
```diff
@@ -16,10 +16,18 @@ fn session_file_name(id : @agent_session.SessionId) -> String {
...
-/// User-authored display titles live beside the append-only transcript. A
-/// separate file lets a running session be renamed without rewriting history
-/// or replacing the inode its engine is still appending to.
-let title_file_name : String = "session-title"
+/// Facts about a session that live beside its append-only transcript rather
+/// than in it — the user's display title, the model its last run used. One
+/// JSON sidecar, replaced atomically under the session lock, lets a running
+/// session be renamed or re-tiered without rewriting history or replacing the
+/// inode its engine is still appending to.
+let metadata_file_name : String = "metadata.json"
```
bzy-debug: test

bzy-debug: reply test

bzy-debug:
test overall comment

PR #1582, opened 2026-09-16
prompt: verified .mbtx examples inlined by the generator (checked errors, PTC edit, PTC search)

bobzhang:
## Summary

First half of the prompt refactoring: language and PTC material in the system prompt becomes **verified `.mbtx` examples** that are written once, checked by CI, and inlined into the generated prompt text. This PR adds the mechanism and the first three examples; the second PR reorganizes the prompt into its two parts (how to work / MoonBit) and converts the remaining MoonBit sections the same way.

- **Inclusion is ordinary markdown.** A link that is a whole line and points at a `.mbtx` file, such as `[share/examples/checked_errors.mbtx](../share/examples/checked_errors.mbtx)`, is expanded by `scripts/md_to_mbt_string` into a fenced `mbtx` block holding the file's text. A link inside a sentence stays a link; targets resolve relative to the markdown file. The existing `--check` therefore also fails when an example changes without regeneration.
- **Three examples under `share/examples/`**, each a runnable `main` whose comments carry the explanation and whose asserts carry the claims: `checked_errors.mbtx` (checked errors as an effect, `suberror`, translation at a boundary, `fn main raise`), `ptc_guarded_edit.mbtx` (one guarded `edit` from a script, reading `result.data`), `ptc_search_filter.mbtx` (a `web_search` filtered before it reaches the model, keeping URLs and truncation).
- **Verified three ways.** `scripts/check-workflows.mbtx` now type-checks `share/examples` as well as `share/workflow` under `--deny-warn` on wasm; `tests/cram/examples.md` runs the language example and pins its output; `agent_tool/mbtx/examples_test.mbt` runs the two PTC examples through the real sandbox, SDK and host (a guarded fix applied, a warning-introducing one reverted with the counts the script prints; a stub search filtered by domain with truncation reported).
- **The prompt's "Checked Error Handling" section** is reduced to its rules plus the example link; **the PTC tool description** is now generated from `agent_tool/ptc/ptc_description.md`, whose two links expand into the PTC examples, and `description()` returns the generated text. `just prompt` / `just check-prompt` and the CI drift step cover both generated files.

## Why this shape

The failure mode of prose is silent staleness. The prompt's own inline snippet used `@string.from_str`, a generic `FromStr` call that only compiles with a result annotation; the verified example failed its first check in a `catch` position and now uses `@string.parse_int`, whose type is explicit. Every example is now compiled and run by CI, so the prompt cannot teach something the toolchain rejects.

## Tests

Generator unit tests (link detection, expansion); the checker passes on 14 bundled scripts; cram page green; the two host-backed example tests pass; `moon check --deny-warn` clean on native, js and wasm; `moon fmt --check`; both generated files match their sources; full JS suite passes.


bobzhang:

Independent review run by Claude Code on the branch head against `origin/main`. Transcript posted verbatim for the human reviewer.

### Review (`review_ex1`)

Blocking defects:

1. **[P2] Failed restoration is reported as a successful revert.** At [ptc_guarded_edit.mbtx:52](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_ex1/share/examples/ptc_guarded_edit.mbtx:52), the `reverted` branch discards `restore_failed` and `result.content`. The host explicitly returns `outcome="reverted"` even when restoration fails ([edit.mbt:1272](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_ex1/agent_tool/edit/edit.mbt:1272)). If an editor saves during the check, or restoration encounters an I/O error, the script still prints only “reverted: …”. Since only printed output reaches the model, it loses the warning that the file was **not restored**. Handle `restore_failed` before printing the normal revert summary and preserve the failure explanation.

2. **[P2] Relative includes break for Windows input paths.** [main.mbt:32](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_ex1/scripts/md_to_mbt_string/main.mbt:32) derives the Markdown directory using only `/`, despite the existing Windows basename support. With input `prompt\default_prompt.mbt.md`, `base_dir` becomes `"."`; its `../share/examples/checked_errors.mbtx` link consequently resolves outside the repository and generation fails. Use platform-aware parent-path handling and cover backslash paths.

Non-blocking suggestions:

- Add generator coverage through `generated_source`, including examples without a trailing newline and literal `<!-- ... -->` inside included code. The new expansion test stops before comment stripping.
- Use `String?` for the optional domain at [ptc_search_filter.mbtx:20](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_ex1/share/examples/ptc_search_filter.mbtx:20); the empty-string absence sentinel contradicts the repository’s stated convention.

VERDICT: REQUEST_CHANGES

bobzhang:

Independent review run by Claude Code on the branch head against `origin/main`. Transcript posted verbatim for the human reviewer.

### Review (`review_ex2`)

The reported fixes are present: restore failures take precedence and print host content; path handling uses the platform-aware basename; domain absence uses `String?`; and the new test exercises comment preservation through `generated_source`.

1. **Blocking — P2: CRLF Markdown silently skips example inclusion.** At [main.mbt:77](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_ex2/scripts/md_to_mbt_string/main.mbt:77), splitting on `"\n"` retains the trailing `"\r"`. The detector at [main.mbt:105](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_ex2/scripts/md_to_mbt_string/main.mbt:105) then rejects `[x](ex.mbtx)\r` because it does not end with `")"`. With CRLF inputs, including a Windows `core.autocrlf=true` checkout, regeneration replaces the embedded examples with links. Subsequent example edits also become invisible to `--check`. Normalize line endings before detection and add a CRLF regression case.

2. **Non-blocking:** Preserve the filesystem root in [main.mbt:57](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_ex2/scripts/md_to_mbt_string/main.mbt:57). `parent_dir("/p.md", "p.md")` currently returns `"."`, so relative includes resolve from the working directory instead of `/`. The revised path tests omit this edge case.

No files were modified, and no builds or tests were run.

VERDICT: REQUEST_CHANGES

bobzhang:

Independent review run by Claude Code on the branch head against `origin/main`. Transcript posted verbatim for the human reviewer.

### Review (`review_ex3`)

1. Blocking defects: none found in the full diff or surrounding code.
2. Non-blocking suggestions: none.

Confirmed that CRLF normalization precedes link detection, with a regression test, and `parent_dir` preserves the root separator for both `/p.md` and `\p.md`, with tests. This slice is self-contained.

Read-only review completed; no files modified, builds or tests run.

VERDICT: APPROVE

PR #1585, opened 2026-09-16
prompt: two-part reorganization (how to work / MoonBit) with the remaining language sections as verified examples

bobzhang:
## Summary

Second half of the prompt refactoring (stacked on #1582). The system prompt is reorganized into two parts, and the remaining MoonBit language sections are carried by verified `.mbtx` examples.

- **Part 1: Working With The Tools** is how the agent works on any task, unchanged in content: bundled references, reading files, running commands (with `moon` subcommands, `moongrep`, programmatic tool calls), the tool protocol and delegation, git and shipping, file references.
- **Part 2: MoonBit** opens with the compiler-driven loop (the paragraphs that used to sit in the preamble, moved verbatim), then project setup, Proton, syntax discipline, checked errors, strings/maps/JSON, CLI parsing, and validation before finish. The preamble now says what the two parts are.
- **Three more verified examples under `share/examples/`**, each type-checked under `--deny-warn` on wasm by the bundled-script checker and run by `tests/cram/examples.md`: `syntax_basics.mbtx` (bindings, negation, `Map([])`, match arms, lambdas), `strings_and_views.mbtx` (interpolation, multi-line literals, code units, clamping views, shortlex order, in-place sort, map lookup, JSON patterns), `cli_count_input.mbtx` (the `argparse` + async IO pattern, run under cram with stdin).
- **Sections reduced to rules plus the example link.** "Syntax And API Discipline" keeps its `moon ide doc` bullets and states the syntax rules in one bullet; "Multi-Line Strings" is folded into "Strings, Maps, JSON, And Tests", which keeps its JSON bullets; "CLI Parsing And Native IO" keeps its bullets and links the example instead of the inline block. Heading names are otherwise stable, so the PTC prompt eval's slicing on `### Programmatic tool calls` / `## Tool Protocol` and the sub-prompts' "keep in sync with Running Commands" notes still apply.

## What verification caught

Writing the examples surfaced two stale claims in the old prompt text, both corrected:

- A bare `{}` is not "an empty map": under `--deny-warn` it is the `ambiguous_braces` error. The rule now says `Map([])` for an empty map and `()` for the no-op expression.
- `\d` in a `re"..."` literal is deprecated syntax; the example uses `[[:digit:]]`.

Also fixed: the "Match arms are separated by newlines" bullet was followed by an unrelated `from_str` block; the moved loop paragraph's "see Running Commands below" now points at Part 1.

## Not in this PR

- `prompt/default_prompt_zh.md` and its generated literal are not wired into any build rule, CI step, or caller (`default_prompt_zh()` has no references), so they were left untouched rather than half-mirrored. Owner's call whether to drop them or re-sync by hand.
- The "Common `moon` subcommands" bullet says "never probe local packages through `moon run -e` (see below)", but nothing below explains it; that dangling pointer predates this PR.
- Prompt wording is the owner's to tune; this PR changes structure and replaces inline code with the examples.

## Tests

Bundled-script checker: 17 scripts clean on wasm under `--deny-warn`; cram page: 4 cases green (three language examples plus the CLI example on stdin); generator tests 16/16; `moon check --deny-warn` clean on native, js and wasm; `moon fmt --check`; generated prompt and PTC description match their sources; native tests for prompt/mbtx/ptc/agent and the full JS suite pass.


bobzhang:

Independent review run by Claude Code on the branch head against `origin/prompt/verified-examples`. Transcript posted verbatim for the human reviewer.

### Review (`review_rg1`)

- **[P2] Restore the mutable-field rule** — [default_prompt.mbt.md:903](/Users/dii/git/openseek/.claude/worktrees/pr_1532_review_rg1/prompt/default_prompt.mbt.md:903). The reduction drops the instruction to declare reassigned struct fields as `mut field : T`. `syntax_basics.mbtx` covers local rebinding and collection mutation, but contains no mutable struct field. This loses the distinction needed to implement methods assigning `self.field`. Preserve it in the bullet or a verified example.

Part 1’s sections and the moved compiler guidance otherwise match the base. The generated prompt matches the expanded source, and API documentation supports the examples’ claims.

Compiler verification was blocked by the read-only sandbox: `moon check` could not create its target directory.

bobzhang:

Independent review run by Claude Code on the branch head against `origin/prompt/verified-examples`. Transcript posted verbatim for the human reviewer.

### Review (`review_rg2`)

No actionable findings against `prompt/verified-examples`.

Confirmed the P2 fix: the syntax rule preserves field mutability, and `Counter::bump` assigns through `self` with an assertion in `main`. The generated prompt includes both changes.

Source/generated consistency, unchanged tool sections, PTC slicing, and diff whitespace checks passed. Cram and Python test suites were blocked by the read-only sandbox.

VERDICT: APPROVE

prompt/default_prompt.mbt.md
```diff
@@ -777,7 +808,7 @@ SDK import, the call API, background handoff, and the call limits.
 - Do not import `moonbitlang/core` as a package. Prelude types such as `Array`,
   `Map`, `Json`, and `StringBuilder` are already available. Import specific
   core subpackages only when needed, for example
-  `moonbitlang/core/string` for typed `@string.from_str` parsing,
```
bobzhang, line 780: `@string.from_str` is preferred, it is more general
let x : Ty = @string.from_str(..)

prompt/default_prompt.mbt.md
```diff
@@ -752,6 +734,55 @@ SDK import, the call API, background handoff, and the call limits.
     instead of papering over it. A red check you cannot explain is a
     finding to report, not a detail to omit.
 
+## File References
+
+When referencing a real local file, prefer a clickable markdown link.
+
+- Clickable file links should look like
+  [app.py](/abs/path/app.py:12): plain label, absolute target, with optional
```
bobzhang: use mbt extension?

bobzhang (commented):
done

prompt/default_prompt.mbt.md
bobzhang: Done in 072816343: the prompt says `let n : Int = @string.from_str(text)` again (project setup pointer and the strings bullet), and the three examples that parsed integers (strings_and_views, checked_errors, ptc_guarded_edit) use `from_str` with an annotated binding.

prompt/default_prompt.mbt.md
bobzhang: Done in 072816343: the example link is now `[main.mbt](/abs/path/main.mbt:12)`. The section itself is folded into one bullet under Tool Protocol, since it is about how answers render, not a topic of its own.

PR #1588, opened 2026-09-16
feat(desktop): refine minimal tool groups and status presentation

tonyfettes:
## Summary

Minimal mode now shows every call in the current tool group and folds earlier groups after new assistant prose arrives. Folded summaries preview the first three descriptions instead of generic tool categories or the latest call alone.

- Render single calls without duplicate disclosures, preserve edit previews, and keep manual expansion state across updates.
- Use aligned 14×14 Codicons, place chevrons after titles, remove vertical guides, and separate 20px turn spacing from 12px internal spacing.
- Retain pending spinners and failure counts on the right while keeping successful calls quiet. Preserve completion timestamps and recorded snapshot behavior from main.

## Validation

- CI follow-up: file-tree tests now await inserted directory rows before clicking the next sibling; Codex tests wait for the Codex textbox instead of filling the outgoing SeekMoon composer. Each of six affected tests passed five local repetitions (30 runs).
- The native symlink edit-guard test failed once in CI but passed 20 local repetitions. Its assertion now prints the tool result on failure; the cause remains unconfirmed pending the next CI result.

- Latest spacing-only follow-up: browser build and four focused E2E tests passed for 12px internal spacing, single/group consistency, wide/narrow layouts, and icon alignment.

- `just check`, `just test`, and `just build` passed on the rebased branch. Full tests include 3,216 native tests, 3,284 JS tests, 65 documentation cases, CLI integrations, and bundled workflows.
- Transcript component tests: 40 passed.
- Minimal transcript, tool-status, and background-jobs browser regression: 38 passed. Includes active-group transitions, retained disclosure state, single-call flattening, failure counts, narrow-width truncation, equal spacing, and 14×14 icon alignment.
- macOS app packaging and signature verification passed before rebase.

The full check/test commands were rerun with sandbox restrictions lifted after local cache locks, TCP listeners, and test Git writes were blocked. Remote CI runs separately on this draft PR.


tonyfettes:
image

PR #1589, opened 2026-09-16
feat(desktop): key editor documents by absolute resource URI

tonyfettes:
## Summary

The dock strip and the fileeditor document table keyed open documents by checkout-relative path. That made document identity depend on the active checkout: a text file outside the conversation checkout could not be opened at all (the host refused it), and switching workspace could retarget an open tab to another checkout's same-named file.

This PR keys documents by their device-bound absolute URI (`@common.Uri`) instead.

- `DockTab::File`, `State.docs`, and the per-tab view-mode / package-graph maps now use `@common.Uri`.
- Workspace membership (`Ctx::relative` / `Ctx::resource`) is derived per operation — tree selection, breadcrumbs, Changes review, diagnostics, package graphs — and is no longer a precondition for opening a file.
- `host.open_path` returns an absolute URI path for text targets. Directory targets stay checkout-relative for the Files tree. `docs/remote-protocol.md` updated.
- Files outside the checkout are not covered by the checkout watcher, so they are re-stated every 2s while any is open (`RefreshExternalFiles`).
- The viewer's scroll-state cache and the review hunk bridge key on the absolute document rather than the relative name.
- The viewer's `ViewerHost` keeps one `document` URI plus a `live` flag instead of separate `absolute` / `document` / `relative` names; the comment chip path and HEAD-side feedback gate derive the workspace name from the navigation context.

Cleanups folded in: `request_diagnostics` derives the relative path itself; the `State`-level `relative`/`resource` helpers are gone in favour of `Ctx`; test fixtures use `test_file` / `work_file` helpers instead of repeating the resource construction.

## Not in this PR

- Dropping `absolute` from the editor's `Content` / `TabLoaded` types: stacked as #1608.
- Feeding external files into the fs watcher (which accepts absolute paths) instead of polling. Follow-up.
- Comments on a file outside every root are dropped silently (there is no workspace-relative chip path for them). Same as before for relative-keyed documents, but now reachable.
- `OpenPathEditorTarget` grew a hand-written JSON codec that requires an absolute path. `open_editor_at` already falls back to relative resolution, so this could revert to `derive`. Left as-is for review.

## Verification

- `moon check --target js`, `moon test --target js` in `desktop/` (3286 passed).
- Debug app builds with `moon run package/macos --target native -- --no-open`.
- Manual E2E still pending: open a checkout-external text path from the transcript, same-basename files inside and outside the checkout, review hunk navigation, MoonBit diagnostics only for in-checkout files, workspace switch keeps tabs pointed at their original files.


tonyfettes:
@codex review

tonyfettes:
@codex review

PR #1596, opened 2026-09-16
feat(env): resolve bundled references from the running executable

tonyfettes:
Builds on #1591, now merged as 829db3de: `current_exe()` and
`current_exe_dir()` exist, and this PR is their consumer. It removes the
`OPENSEEK_REFERENCES` environment variable and derives the bundled resource
root from the running executable instead.

`internal/references` answers "where are the bundled resources" once, for both
consumers. Resolution starts at the executable's own directory and walks up; at
each ancestor it probes `share/` and
`toolchains/moonbit/{macos-arm64,linux-x64,windows-x64}/share`, and the first
candidate holding both `doc/moonbit/` and `workflow/` wins. That covers a plain
install (`share/` beside `bin/`), the packaged desktop tree (`seekmoon/bin/openseek`
to `seekmoon/toolchains/moonbit/<platform>/share`), and a source checkout
(`_build/...` walking up to the repository's `share/`). Nothing is read from the
environment, and nothing from the working directory.

- The system prompt names that directory on its `References:` line. The prompt
  sources, the Chinese source, and both generated prompts are updated, and the
  layout placeholder is renamed to `{{REFERENCES_LAYOUT}}`.
- `mbtx` resolves `@builtin/` under the same root, and its namespace error now
  reports that the installation ships no bundled resources instead of naming the
  variable.
- The desktop advertises no path at all: `ResolvedToolchain` loses `references`,
  and the engine environment neither sets nor scrubs the variable.
- Readmes, the mbtx tool description, the subagent prompts, and the eval harness
  describe the resolved directory instead.

Verification, all local on macOS arm64 with moon 0.1.20260911 / moonc 0.10.12
(the branch is rebased onto `main` at 4073d809, so every gate was rerun on the
rebased tree):

- `moon check --target native --deny-warn`, `moon check --target js --deny-warn`,
  `moon fmt --check`
- `moon run scripts/check-workflows.mbtx` (17 scripts), and
  `scripts/md_to_mbt_string --check` for the default prompt and the PTC
  description, which is also the generated-resource-layout CI step
- `moon info` leaves every `.mbti` byte-identical (the guard at
  `.github/workflows/ci.yml:177-192`)
- `moon test --target native` (whole workspace), `moon test --target js` (3286
  passed), `moon cram test tests/cram --shell bash` (68/68), and
  `moon run tests/integration/workflows`
- End to end: `moon build cmd/openseek --target native`, then
  `moon run tests/integration/turn_finish.mbtx <exe>`. The Read scenario passes
  with nothing in the child's environment naming the resources: "real CLI
  advertises the read workflow, omits read, and returns ranged batch output"
- The packaged layout was checked against the installed app on this machine:
  `SeekMoon.app/Contents/Resources/seekmoon/{bin/openseek, toolchains/moonbit/macos-arm64/share/{doc/moonbit,workflow}}`

Not verified here:

- `moon test --target wasm` cannot run locally: this machine's moonrun does not
  provide the `moonbitlang/core` `env/current_exe(_dir)` imports, so wasm
  artifacts fail at instantiation before any test runs. #1591's
  `check (stable, wasm)` row was green, which is the row that covers this
  requirement.
- An independent review subrun (`@builtin/review.mbtx`) timed out before
  reporting, so it produced no findings.

Sessions saved before this change keep whatever absolute path their recorded
prompt carries; they are not migrated.

Generated with SeekMoon


tonyfettes:
Superseded by #1607: https://github.com/moonbitlang/openseek/pull/1607

We are proceeding with `codex/executable-relative-resources`. It removes `OPENSEEK_REFERENCES` using a fixed `bin/` + `share/` installation layout, resolves the real executable path before resource lookup. This replaces the ancestor/toolchain-directory search approach in this PR.

Keeping #1596 closed in favor of #1607. The branch for #1596 has been retained.

PR #1599, opened 2026-09-16
lint: enforce and fix four more warning families

bobzhang:
Follows the merged #1594 (`test_unqualified_package`); based on `main`.

## What

Four more warning families are now enabled in every module of the
workspace, and their backlog is fixed — **1897 sites**, each rewritten by
the rule the compiler states in its own message:

| warning | sites | the compiler's rule | the fix |
---|---|---|---|
| `unnecessary_annotation` (73) | 1814 | "This `@protocol.` annotation is unnecessary." | delete the named qualifier |
| `unnecessary_view_op` (75) | 25 | "`[:]` is unnecessary here" | delete the `[:]` |
| `unqualified_local_using` (72) | 23 | "…replace it by `@icons.icon`" | see below |
| `declaration_implemented` (69) | 1 | "The declaration 'args' has been implemented" | delete the redundant `declare`; the implementation stays |

167 files, +1852 / −2960 (net −1108 lines): the deletions also let
`moon fmt` rejoin calls the shorter expressions no longer need to wrap.

### The `unqualified_local_using` fix keeps the interfaces as they were

The 23 sites are private `using` declarations: the three
`icon_imports.mbt` shims that existed only to bring the frontend icon
package's names into their package's scope, and three
`using @cmd {trait Scheduler}` lines. Each is gone, every name it supplied
is now written qualified (`@icons.icon(…)`, `@cmd.Scheduler`), and the
three shim files are deleted — the alternative `moon explain` names
("make the alias public") would have *widened* five packages' public API,
which the interface gate caught. `moon info` now reports no `.mbti` drift.

## Second-order effect the guards caught

Deleting the qualifiers made three packages' imports genuinely unused, so
the guarded batches rolled back until the imports went too:
`desktop/frontend/moon.pkg` loses `bobzhang/openseek_protocol` (`@wire`)
and `moonbit-community/rabbita/url`, and
`desktop/frontend/fileeditor/moon.pkg` loses
`moonbitlang/editor/viewer/common/agent_feedback_api`.

## Measured, deliberately not enabled

These need a human, so they are reported here instead of turned on:
`missing_doc` (1162 — prose), `unqualified_record` (682 — the message's
`T::{}` template leaves the type unnamed), `missing_invariant` /
`missing_reasoning` (155 each — per-loop proofs), `unused_default_value`
(96) and `unused_optional_argument` (59 — API-visible),
`prefer_fixed_array` (3), `result_error_return` (1).

## Validation

- `moon check --target js|native|wasm --deny-warn` (the root CI gate) —
  all three clean, with the four families enforced by the manifests.
- `moon info` — no generated-interface drift (the CI gate that first
  failed on the `pub using` variant of this fix).
- `moon fmt --check` — clean.
- `moon test desktop/frontend` — 476 passed; the desk/backend, editor and
  agent_tool suites pass too.

Generated with SeekMoon


bobzhang:
Superseded: the same four families are now four reviewable PRs, one per warning — #1602 (73), #1603 (75), #1604 (72), #1605 (69), stacked in that order. The `sweep/more-warnings` branch is deleted.

PR #1608, opened 2026-09-16
refactor(desktop): drop the read reply's absolute path from the editor's content types

tonyfettes:
Follow-up to #1589, rebased onto `main` after it merged.

## Summary

`FileContent::Content` and `TabState::TabLoaded` carried the host's echo of the requested path as `absolute`. With documents keyed by their absolute URI (#1589), that echo always equals the document key, so nothing needs it:

- The viewer already switched to `document` + `live` in #1589.
- `semantic_review_input` takes the resource from the active document.
- `resolve_navigation_preview` compared the echo against the path it had just requested; the check was tautological and is gone.
- `FileContent::from_reply` no longer needs the channel or an `Option` result. Two tests that fenced an empty or relative echo go with it.

Net: 11 files, +70 / −183, almost all of it test fixtures that no longer construct an absolute resource for each `Content` / `TabLoaded`.

## Wire contract unchanged

`fs.read_file` still returns `{content, absolute, sig}`. Older frontends bind their resource from `absolute`, and `docs/remote-protocol.md` keeps host/client compatibility in both directions, so the host keeps echoing it and this decoder skips it. Dropping the field from the wire is a separate decision.

## Verification

- `moon check --target js`, `moon test --target js` in `desktop/` (3286 passed).
- `pkg.generated.mbti` for fileeditor regenerated; the only change is the two constructor signatures.


desktop/backend/internal/host/open_ops.mbt
```diff
@@ -55,33 +55,13 @@ async fn open_path_op(payload : OpenPathPayload) -> OpenPathReply {
...
-      return refusal("Text file is outside the conversation checkout: \{path}")
-    }
-    let real_root_text = @fs.realpath(root.to_string()) catch {
-      error if @async.is_being_cancelled() => raise error
-      error => return refusal("Could not resolve working directory: \{error}")
-    }
-    let real_path_text = @fs.realpath(path) catch {
-      error if @async.is_being_cancelled() => raise error
-      error => return refusal("Could not resolve file: \{error}")
+      absolute.to_uri_path() is Some(resource) else {
```
bzy-debug, line 61: Test

bzy-debug: test suggestion

tonyfettes:
hello

tonyfettes:
`world`

desktop/backend/internal/host/open_ops.mbt
tonyfettes, line 204: test

desktop/backend/internal/host/open_ops.mbt
tonyfettes: reply

PR #1642, opened 2026-09-18
simplify(edit): state conditions directly and share the preview parse

bobzhang:
- agent_tool/edit/edit.mbt: `errors_hit` and `warnings_hit` are `guards.errors is Some(limit) && delta.errors.length() > limit`, replacing two `match` blocks whose only job was to turn an `Option` into a `Bool` via a `None => false` arm.
- agent_tool/edit/edit.mbt: the replace-all preview scan is `while region.body[from:].find(needle) is Some(found) {`, replacing a `for ;;` whose first statement was a `guard ... else { break }` — the loop's real condition now reads as the condition.
- agent_tool/edit/edit.mbt: `render_site_context` binds `let number = "\{line}".pad_start(width, ' ')` at its one call site, replacing the single-use `preview_pad` wrapper, which is deleted.
- agent_tool/edit/edit_test.mbt: both preview tests call a new `preview_edits(content) -> Array[Json] raise`, replacing two copies of the same thirteen-line find-the-section, find-the-brackets, `@json.parse` sequence; the parse and its failures stay inside the helper.
- agent_tool/edit/edit_guard_test.mbt: `shown` is `data.map_or("none", json => json.stringify())`, replacing a two-arm `match` on the `Option` that shadowed its own parameter.

2× pattern match, 1× dedupe, 1× functional loop, 1× inline single-use helper.

## Verification

| gate | result |
| --- | --- |

## How these were picked


bobzhang:
## Review sign-off

> The commit is a behavior-preserving simplification of the edit tool and its tests. In edit.mbt, an infinite `for ;;` loop with `guard ... else { break }` was rewritten as an equivalent `while region.body[from:].find(needle) is Some(found)` loop (line 491); the `preview_pad` helper was inlined into the one call site (lines 657-667); and the `errors_hit`/`warnings_hit` `match` expressions were rewritten as short-circuiting `is Some(limit) && delta...length() > limit` conditions (lines 1158-1160). In the tests, the duplicated preview-array extraction logic was factored into `preview_edits` (edit_test.mbt:86-98) and `shown` was rewritten with `Option::map_or` (edit_guard_test.mbt:409-411). I […]

### Scope reviewed

`agent_tool/edit/edit.mbt`, `agent_tool/edit/edit_guard_test.mbt`, `agent_tool/edit/edit_test.mbt`

### Gates

### Provenance

PR #1645, opened 2026-09-18
simplify(bgjobs): one status read, one spill catch, map_or footers

bobzhang:
- agent_tool/bgjobs/records.mbt: the `Stopping` pre-check is a guarded `Exited(_) | Killed if self.phase is (Live | Finishing)` arm of the status match it used to wrap, so `self.exec.status()` is read once instead of twice and the match no longer sits inside an if/else.
- agent_tool/shell_exec/sink.mbt: `open_spill` wraps its own body in the "failed to create output log" try/catch and says so in its doc comment; `append_locked` and `materialize` are plain `self.open_spill(path)` calls instead of two identical six-line catch blocks.
- agent_tool/job_output/job_output.mbt: the `paging` and `file_hint` footer fragments are `Option::map_or` with an empty-string default instead of two `Some`/`None` matches.

2× pattern match, 1× dedupe.

## Verification

| gate | result |
| --- | --- |

## How these were picked


bobzhang:
## Review sign-off

> The three changed files are pure refactors with no observable behavior change:

### Scope reviewed

`agent_tool/bgjobs/records.mbt`, `agent_tool/job_output/job_output.mbt`, `agent_tool/shell_exec/sink.mbt`

### Gates

### Provenance

PR #1646, opened 2026-09-18
simplify(agent_tool): one write test driver, guards over no-op matches

bobzhang:
- agent_tool/write/write.mbt: four copies of the `match definition(...).execute { Async(run) => run; Sync(_) => fail(...) }` unwrap collapse into one private `run_write(tool, arguments)` driver, the shape `remove`'s `run_remove` and `mbtx`'s `dispatch` already use
- agent_tool/internal/sandbox/source_write_profile.mbt: the profile-cache `match` whose `Some(_)` and `None` arms were both `()` is now a single `if ... is Some(cached) && cached_source_write_profile_valid(cached)` early return
- agent_tool/agent_tool.mbt: `brief_line`'s per-char if/else-if chain is a three-arm `match`, the `to_int() < 0x20 || == 0x7F` arithmetic written as the `'\u{00}'..='\u{1F}' | '\u{7F}'` range pattern the repo already uses for this set
- agent_tool/remove/remove.mbt: `remove_file`'s single-use `exists` and `kind` bindings fold into the `guard`s that tested them, matching the three guards below
- agent_tool/write_scope/write_scope.mbt: `allowed_display` joins `allowed_paths` directly instead of mapping `String::to_string` over it into a duplicate array first
- agent_tool/write_scope/write_scope_test.mbt: `assert_admits` is an else-less `if scope.deny_reason(path) is Some(reason) { fail(...) }` instead of a match with a `None => ()` arm

4× pattern match, 1× dedupe, 1× functional loop.

## Verification

| gate | result |
| --- | --- |

## How these were picked


bobzhang:
## Review sign-off

> This is a single, behavior-preserving simplification commit ("simplify(agent_tool): one write test driver, guards over no-op matches"). I compared each changed hunk against its base (26132c893) and confirmed semantic equivalence:

### Scope reviewed

`agent_tool/agent_tool.mbt`, `agent_tool/internal/sandbox/source_write_profile.mbt`, `agent_tool/remove/remove.mbt`, `agent_tool/write/write.mbt`, `agent_tool/write_scope/write_scope.mbt`, `agent_tool/write_scope/write_scope_test.mbt`

### Gates

### Provenance

PR #1648, opened 2026-09-18
simplify(viz): share one tool-card classifier and fold Option matches

bobzhang:
- viz/view.mbt: the two byte-identical `(is_error, is_escalated)` matches in `tool_card` and `model_tool_card` become calls to one private `tool_card_kind(is_error~, is_escalated~)`, sited next to `failed_flag` alongside the other shared classifiers
- viz/view.mbt: `scalar_text`'s nested `match repr { Some/None }` inside the `Number` arm becomes `repr.unwrap_or(n.to_string())`
- viz/view.mbt: `model_tool_card`'s `match shown { Some(r) => r.tool_name(); None => "" }` becomes `shown.map_or("", r => r.tool_name())`
- inspect/main.mbt: `session_envelope_body` folds into its only caller `session_envelope_reply`, which already documents the `{found, events, events_bytes}` shape; the helper's "so the exporter can bake the same body" rationale was stale, since the exporter builds its envelope in `@viz_export.Session::envelope_json`
- inspect/main.mbt: `session_list_json` folds into its only caller `session_list_reply` as `catalog.rows().map(row => row.to_json()).to_json()`; its "shared with the standalone exporter" doc was likewise stale
- viz_export/export.mbt: `Session::listing_json`'s four-line `last_active` match becomes `self.last_active.map_or(Json::null(), n => Json::number(n))`

3× pattern match, 2× inline single-use helper, 1× dedupe.

## Verification

| gate | result |
| --- | --- |

## How these were picked


bobzhang:
## Review sign-off

> The change set is a mechanical, behavior-preserving refactor across three files: it inlines the now-unused `session_envelope_body` and `session_list_json` helpers into `session_envelope_reply`/`session_list_reply` in `inspect/main.mbt`, factors the duplicated tool-card class match into a shared `tool_card_kind` in `viz/view.mbt` (used by both the raw and model views), and replaces three explicit `match`/`Some` expressions with `map_or`/`unwrap_or` in `viz/view.mbt` and `viz_export/export.mbt`. I verified equivalence against the pre-change source (`git show 26132c893:...`) and confirmed no stale references to the removed helpers remain (the `session_list_json` found in […]

**Acted on the one nit.** `openseek` flagged that `repr.unwrap_or(n.to_string())` evaluates its default eagerly, so every rendered JSON number formatted itself even when the parser had preserved its source spelling (the common case). Output was unchanged, but the work was not. Fixed by making the fallback lazy — `repr.unwrap_or_else(() => n.to_string())` — and the branch was amended and re-pushed. Re-gated: `moon check --target js --deny-warn viz` clean, `moon test --target js viz` 19/19.

### Scope reviewed

`inspect/main.mbt`, `viz/view.mbt`, `viz_export/export.mbt`

### Gates

### Provenance

PR #1649, opened 2026-09-18
simplify(protocol,jsonrpc): job decode, fold, unwrap in one line each

bobzhang:
- protocol/jobs.mbt: new private `job_required` helper replaces seven copies of the three-line `FromJson::from_json(job_field(fields, <key>, path), path)` dance with one-line calls that read like the neighbouring `job_optional` / `job_uint` lines
- protocol/jobs.mbt: `valid_job_component` tests the characters with `value.all(...)` instead of the `value.iter().all(...)` hop that forced the chain across three lines
- protocol/jobs.mbt: the `JobRecord` decode guard ends in `output_file.map_or(true, valid_job_component)` instead of a four-line parenthesized `match` wedged into the boolean chain
- protocol/subrun_session.mbt: the digit accumulator is `digits.fold(init=0, (ordinal, c) => ...)` instead of a `for ... nobreak` loop that never breaks
- jsonrpc/client.mbt: `Client::request` ends with `inbox.get().unwrap_or_error()` instead of the `Ok(json) => json` / `Err(error) => raise error` match that spells out that function's body
- jsonrpc/client_wbtest.mbt: the reader-cancellation test calls the existing `request_outcome(client, "ping")` helper instead of a hand-inlined copy of its try/catch, including the cancellation guard

2× dedupe, 2× functional loop, 2× pattern match.

## Verification

| gate | result |
| --- | --- |

## How these were picked


bobzhang:
## Review sign-off

> Reviewed the four changed files. All changes are behavior-preserving refactors: `Client::request` now uses `Result::unwrap_or_error` (verified equivalent to the removed `match`, raising the contained `ResponseError`), the cancellation regression test reuses the existing `request_outcome` helper, `JobRecord::from_json` introduces a `job_required` helper plus `String::all`/`Option::map_or` simplifications, and `split_child_session_id` replaces a `for`/`continue`/`nobreak` accumulator with `fold`. `moon check --target all` passes. The affected packages pass on native (jsonrpc 33/33, openseek_protocol 28/28) and the full suite passes on wasm (2427/2427). The default native `moon test` did not […]

**Declined the one nit, on the facts.** `openseek` asked for a doc comment on the new `job_required`, saying the neighbouring private helpers each carry a one-line summary. They do not: `job_field` (jobs.mbt:90), `job_optional` (:102), `job_optional_json` (:114), `job_uint` (:122) and `job_optional_uint` (:137) all open with a bare `///|` and no prose. `job_required` already matches its neighbours; adding a summary would make it the odd one out.

### Scope reviewed

`jsonrpc/client.mbt`, `jsonrpc/client_wbtest.mbt`, `protocol/jobs.mbt`, `protocol/subrun_session.mbt`

### Gates

### Provenance

PR #1650, opened 2026-09-18
simplify(subagents): name the Option/Result shapes, one must_git helper

bobzhang:
- agent_subtask/lifecycle_test.mbt: eight setup `match @agent_subtask.git_require(...) { Ok(_) => () / Err(reason) => fail(reason) }` blocks become `must_git(repo, args)` over one private async helper, with each command's argv still at its call site; the `rev-parse` match that consumes its `Ok` value and the `discard` match stay as they were.
- agent_subtask/capture.mbt: the four-line `match entry.commit_oid` pin lookup becomes `entry.commit_oid.unwrap_or(entry.base_oid)`.
- agent_subtask/integrate.mbt: `reload_entry`'s line-broken `match ... .last()` becomes the chain `entries.filter(..).last().unwrap_or(entry)`.
- agent_subtask/exports.mbt: `resume_running`'s rewrapping `Ok(_) => Ok(entry) / Err(reason) => Err(reason)` match becomes `save_entry(geometry.common_git_dir, entry).map(_ => entry)`.
- agent_skill/skill.mbt: the single-caller `Skill::Skill` constructor is deleted; its one caller pushes the record literal directly as `skills.push({ name, description, path, })`.

3× pattern match, 1× dedupe, 1× inline single-use helper.

## Verification

| gate | result |
| --- | --- |

## How these were picked


bobzhang:
## Review sign-off

> This commit is a purely mechanical simplification across two packages. It replaces verbose `match`/`Some`/`None` shapes with `Option::unwrap_or` and `Result::map`, drops a hand-written struct constructor in favor of a struct literal, and extracts the repeated `git_require`-then-`fail` setup boilerplate in the lifecycle tests into a `must_git` helper annotated with `#callsite(autofill(loc))` so failures still report the call site. I compared each transformation against its original semantics: no evaluation-order or side-effect differences exist (the unwrap_or/map arguments are plain field accesses / a local value), and the `#callsite(autofill(loc))` pattern matches established usage […]

**2. Independent correctness self-review** (adversarial: arm order and exhaustiveness, eager `map_or`/`unwrap_or` defaults, `&&`/`||` short-circuit order, loop-rewrite iteration counts, Array vs Iter vs String receivers, ArrayView aliasing, error handling moved into a callee, `errdefer`/`defer` scoping, lost doc invariants, changed error strings) — **1 finding** (P3) — addressed below.

**Acted on the self-review's one finding (P3).** `must_git` initially swallowed the call-site location: `fail` is `#callsite(autofill(loc))`, so all seven git-setup failures would have reported the `fail` line inside the helper, and the two same-argv call sites inside one test would have been indistinguishable. Fixed the way the repo already does it for test helpers (`agent/auto_compact_test.mbt:140`): `#callsite(autofill(loc))` on `must_git`, `loc~ : SourceLoc` threaded into `fail(reason, loc~)`. Branch amended and re-pushed; `moon test --target native -p agent_subtask` 20/20.

### Scope reviewed

`agent_skill/skill.mbt`, `agent_subtask/capture.mbt`, `agent_subtask/exports.mbt`, `agent_subtask/integrate.mbt`, `agent_subtask/lifecycle_test.mbt`

### Gates

### Provenance

PR #1651, opened 2026-09-18
simplify(mcp): match request shapes and fold fake-server helpers

bobzhang:
- mcp/client.mbt: `list_tools` pagination ends with `guard next is Some(_) else { break }` and assigns the option straight through, replacing a two-arm `match` that unwrapped the cursor only to rebuild `Some(next)`.
- mcp/types.mbt: `decode_tool` writes the `{ "type": "object" }` fallback inline in an annotated `let input_schema : Json = match`, replacing a call to the one-use `default_object_schema`, which is deleted.
- mcp/tools/tools_test.mbt: `server_reply` dispatches on `{ "method": "...", .. }` JSON patterns against the request, replacing `match method_of(request)` over strings; the `method_of` helper is deleted.
- mcp/client_test.mbt: `fake_client` spawns the message pump, the writer and the serve loop itself, replacing the `spawn_client` / `serve_loop` / `spawn_server` chain of one-caller helpers.
- mcp/client_test.mbt: five tests build their fake server with `answers("<method>", request => ...)`, replacing hand-rolled `fn(request : Json) -> Json? { match method_of(request) { ... } }` closures that differed only in the method and the reply.

2× pattern match, 2× inline single-use helper, 1× dedupe.

## Verification

| gate | result |
| --- | --- |

## How these were picked


bobzhang:
## Review sign-off

> This is a focused, behavior-preserving refactor (single commit e96625bb2). client_test.mbt folds the spawn_client/serve_loop/spawn_server helpers into fake_client and adds an `answers` helper to replace repeated per-test method matchers; tools_test.mbt drops the method_of indirection in favor of direct request-shape patterns; types.mbt inlines the one-use default_object_schema; client.mbt rewrites the list_tools pagination match as a guard. I found no correctness issues: `moon check --target all` passes (984 tasks, no warnings) and the full `moon test` suite passes on all targets (103 wasm + 2972 js + 1927 native, 0 failures). One readability nit only.

**Noted the one nit, no change.** `openseek` observed that `guard next is Some(_) else { break }` followed by `cursor = next` leans on narrowing rather than binding the cursor, and called it more indirect than the `match` it replaced — while confirming it is correct (it verified the narrowing with a standalone snippet and `moon check --target all`). Left as is: the guard states the loop's exit condition where the exit happens, which is the point of the change.

### Scope reviewed

`mcp/client.mbt`, `mcp/client_test.mbt`, `mcp/tools/tools_test.mbt`, `mcp/types.mbt`

### Gates

### Provenance

PR #1652, opened 2026-09-18
simplify(eval): dispatch with patterns, inline two one-line helpers

bobzhang:
- scripts/serve-static.mbtx: `content_type` is a single `match` over string suffix patterns instead of a 12-branch `has_suffix` if/else-if chain, same arm order
- eval/report/report.mbt: the sole call to `has_detail_section` is now `self.rows.any(row => row.has_details())`, and the one-caller helper is deleted
- scripts/internal/markdown_comments/markdown_comments.mbt: `inline_raw_html_is_comment` ends in `text.to_string().trim_start().has_prefix("<!--")`, and the `is_html_comment` helper is deleted
- eval/tool_harness/harness.mbt: `expect_finish`'s if/else inside the `Finish` arm is a guarded arm plus its unguarded fallback, one line each
- eval/file_edit/harness/harness.mbt: the tool_result arm matches `"is_error": True` in the JSON pattern instead of binding `is_error` and re-testing it in the body
- eval/session_analyzer/analyzer.mbt: the nested `match terminal` folds into the outer arm as `Terminal(Finished | Aborted | Interrupted | Failed)`

4× pattern match, 2× inline single-use helper.

## Verification

| gate | result |
| --- | --- |

## How these were picked


bobzhang:
## Review sign-off

> The change set is a mechanical refactor of the same logic into newer MoonBit pattern/guard forms: JSON field patterns tightened to literal `True`, a nested `if` converted to a match guard, a single-use helper inlined, and the `content_type` suffix chain converted to an array-pattern match. I diffed all six files against the base, read each changed site with context, and verified behavior: `moon check --target all` succeeds at the root (984 tasks) and in the `scripts` project (30 tasks); the serve-static script itself compiles and runs (fails only on its expected usage guard). Targeted tests for every affected package pass (eval/report 4/4, eval/session_analyzer 11/11, eval/tool_harness 4/4, […]

**Noted the one nit, no change.** `openseek` preferred the original `has_suffix` chain to the `match path { [.., .. ".html"] => ... }` form on readability grounds, after verifying the two agree exactly — including on `.html` itself, `foo/a.html`, `html`, `a.htmlx`, the empty string and non-ASCII prefixes such as `é.html`. Keeping the match: twelve `else if path.has_suffix(...)` lines collapse to twelve one-line arms, and the suffix is now the pattern rather than a call buried in a condition.

### Scope reviewed

`eval/file_edit/harness/harness.mbt`, `eval/report/report.mbt`, `eval/session_analyzer/analyzer.mbt`, `eval/tool_harness/harness.mbt`, `scripts/internal/markdown_comments/markdown_comments.mbt`, `scripts/serve-static.mbtx`

### Gates

### Provenance

eval/file_edit/harness/harness.mbt
```diff
@@ -244,11 +244,9 @@ fn summarize_agent_log(log_text : String) -> AgentLogSummary {
         has_json_events = true
         steps += 1
       }
-      { "event": String("tool_result"), "is_error": is_error, .. } => {
```
bobzhang, line 247: is this semantic perserving?

bobzhang: Yes — and it is worth spelling out why, because it is only safe thanks to the arm that follows it.

```
-      { "event": String("tool_result"), "is_error": is_error, .. } => {
+      { "event": String("tool_result"), "is_error": True, .. } => {
         has_json_events = true
-        if is_error is True {
-          tool_errors += 1
-        }
+        tool_errors += 1
       }
       { "event": String(_), .. } => has_json_events = true
```

Narrowing the arm means `tool_result` lines whose `is_error` is *not* `True` no longer match it. They fall to the next arm, `{ "event": String(_), .. } => has_json_events = true` — which is exactly the part of the old body that used to run for them. The old arm's body was `has_json_events = true` plus a conditional increment; the fall-through arm's body is that same `has_json_events = true`. So nothing is lost in the fall-through, and the counter still only moves on `True`.

Three shapes to worry about, all of which behave the same before and after:

| `is_error` | before | after |
| --- | --- | --- |
| `true` | arm 2, `has=true`, `+1` | arm 2, `has=true`, `+1` |
| `false` / null / string / number / object / array | arm 2, `has=true`, no increment | arm 3, `has=true`, no increment |
| **key absent** | arm 3 (a non-optional `"k": pat` requires the key), `has=true`, no increment | arm 3, `has=true`, no increment |

The absent-key row is the one that would bite if JSON key patterns were optional by default — they are not; `"is_error": is_error` already failed to match a `tool_result` line with no `is_error` field, so both versions route it to arm 3 identically.

I checked this differentially rather than by reading, running both arm-sets over the same inputs:

```
same  is_error true: old=(true, 1) new=(true, 1)
same  is_error false: old=(true, 0) new=(true, 0)
same  is_error ABSENT: old=(true, 0) new=(true, 0)
same  is_error null: old=(true, 0) new=(true, 0)
same  is_error string: old=(true, 0) new=(true, 0)
same  is_error number: old=(true, 0) new=(true, 0)
same  is_error object: old=(true, 0) new=(true, 0)
same  is_error array: old=(true, 0) new=(true, 0)
same  with extra keys: old=(true, 1) new=(true, 1)
same  agent_step: old=(true, 0) new=(true, 0)
same  other event: old=(true, 0) new=(true, 0)
same  event not a string: old=(false, 0) new=(false, 0)
same  no event key: old=(false, 0) new=(false, 0)
same  empty object: old=(false, 0) new=(false, 0)
same  not an object: old=(false, 0) new=(false, 0)
same  a string: old=(false, 0) new=(false, 0)

EQUIVALENT on all 16 shapes
```

(pair of `(has_json_events, tool_errors)` per input; the two arm-sets were pasted verbatim into one `.mbtx` and run side by side.)

One honest caveat on the readability trade: the narrowed pattern is only correct *because* the following catch-all is a superset with a matching body. That coupling is invisible if you read the arm on its own — so if you would rather keep the binding and the explicit `if`, which states the intent locally and survives someone later editing the arm below it, say the word and I will revert this hunk. The other five changes in this PR are independent of it.

PR #1671, opened 2026-09-18
prompt(ptc): use the default @openseek_tools alias and trim the PTC section

bobzhang:
Drops the `@tools` import alias from the PTC example and the system prompt so
both use the package's own default alias, `@openseek_tools`, and trims the host
tool calls section: the call-budget, exit-with-active-calls, and background
handoff paragraphs are gone, and the delegation heading is now `Orchestration`.

## Verification

- `moon test -p bobzhang/openseek/agent_tool/mbtx --target native` — 125/125
  pass, including the host-backed `examples_test.mbt` test that runs the edited
  `share/examples/ptc_guarded_edit.mbtx` through the real sandbox and host, so
  `@openseek_tools` resolves.
- `moon run scripts/md_to_mbt_string -- --check prompt/default_prompt.mbt.md prompt/generated_default_prompt.mbt`
  — generated prompt is in sync.

## Known follow-ups

Not addressed here:

- `agent_tool/mbtx/mbtx.mbt:1681` — the `ptc` schema text still says
  `@tools.call(name, arguments)` while telling the model to import
  `"bobzhang/openseek_tools@0.1.0"`, so it now teaches a symbol a plain import
  does not bind.
- `prompt/default_prompt_zh.md:281` — still `@tools.call(...)`.
- The removed call-budget sentence was the only model-visible statement of
  `MaxActiveCallsPerProgram = 4` (`agent_tool/ptc/server.mbt:13`), which is
  enforced with a hard failure.


bobzhang (commented):
## Review

### 1. The `ptc` schema text still teaches `@tools.call` — `agent_tool/mbtx/mbtx.mbt:1681`

```
"Enable @tools.call(name, arguments) for edit, multi_edit, and web_search
 through an explicit \"bobzhang/openseek_tools@0.1.0\" import."
```

That string (added in 16606c72a) now contradicts both the prompt and the
example. With the plain `"bobzhang/openseek_tools@0.1.0"` import this PR moves
to, the bound symbol is `@openseek_tools` — a model that follows the schema
text verbatim writes an unresolved `@tools.call`. This is the one I would fix
before merging, since the schema text is model-visible on every `mbtx` call.

### 2. Same drift in the Chinese prompt — `prompt/default_prompt_zh.md:281`

Still `@tools.call("multi_edit", { "edits": edits })`, and that `@tools` reaches
the generated `prompt/generated_default_prompt_zh.mbt` unchanged.

### 3. The call budget is now invisible to the model

This PR drops *"Up to 1024 calls per script, four in flight…"*. Those are hard,
enforced limits — `MaxCallsPerProgram = 1024` and
`MaxActiveCallsPerProgram = 4` (`agent_tool/ptc/server.mbt:10,13`) — and
exceeding either fails the call with
`"PTC program exhausted its 1024-call or four-active-call budget"`
(`agent_tool/ptc/server.mbt:175`). Grepping the model-visible descriptions,
nothing else states them.

The four-in-flight cap is the one that bites: the same prompt encourages
`@async.with_task_group` fan-out for independent searches, which now fails at
runtime with no prior warning. The 1024 cap and the size limits seem fine to
drop; I would keep a clause for the concurrency one.

The removed *"a script that exits with calls still active fails"* line and the
background-handoff paragraph read as reasonable trimming — shutdown cancels
active calls (`agent_tool/ptc/server.mbt:387`).

### 4. Cosmetic — these land verbatim in the shipped prompt

- `prompt/default_prompt.mbt.md:154` — trailing space after
  `A mbtx script can call the host's own tools. `
- `prompt/default_prompt.mbt.md:189-191` — three consecutive blank lines before
  `### Orchestration`
- `prompt/default_prompt.mbt.md:155` — wraps at 86 cols while the rest of the
  file sits at ~76
- Heading style diverges from its sibling:
  `### Host tool calls with bobzhang/openseek_tools` vs
  `### Orchestration: the \`moonbitlang/workflow\` package`. Also "A mbtx" →
  "An mbtx".

### Verified

- `moon test -p bobzhang/openseek/agent_tool/mbtx --target native` — 125/125
  pass, including `agent_tool/mbtx/examples_test.mbt:11`, which runs the edited
  `share/examples/ptc_guarded_edit.mbtx` through the real sandbox and host, so
  `@openseek_tools` does resolve.
- `md_to_mbt_string --check` — the generated prompt matches the markdown.

PR #1686, opened 2026-09-20
prompt: use shell star imports in the default prompt's scripts

bobzhang:
## What

`.mbtx` scripts can now write `import { "moonbitlang/async/shell" * }` to use
the package's public names unqualified — `Cmd(...)`, `Pipeline(...)`,
`glob(...)` — while `@shell.Cmd(...)` keeps working. This applies the short
form to the prompt's own scripts and documents when to use it.

- `share/examples/command_output.mbtx`, `share/examples/check_diagnostics.mbtx`,
  and `share/workflow/check.mbtx` (the minimal script the prompt inlines):
  `"moonbitlang/async/shell" *` and bare `Cmd(...)`.
- `prompt/default_prompt.mbt.md`: new guidance bullet — `*` is `.mbtx`-only,
  preferred for shell in short scripts, other packages stay qualified to
  avoid ambiguous names; the `moon.pkg` import bullet now says `*` is not
  allowed there. Inline prose and the CI-monitor snippet switched to `Cmd`.
- `prompt/generated_default_prompt.mbt` regenerated.

## Verification

Run from the workspace root:

- `moon run scripts/md_to_mbt_string -- --check prompt/default_prompt.mbt.md prompt/generated_default_prompt.mbt` — fresh
- `moon run scripts/check-workflows.mbtx` — all 22 bundled scripts type-check under `--deny-warn`
- `moon cram test tests/cram` — 72/72
- `moon run tests/integration/workflows` — all passed
- `moon check --target native --deny-warn`, `moon check --target js --deny-warn`, `moon fmt --check` — clean
- `moon test --target native agent_tool/mbtx --filter "*example*"` — 3/3

Not run: the `eval/prompt_task` model evals (they need an API key).

## Note

Supersedes #1381, which made the same change before the prompt's code blocks
moved into `share/examples`, so its diff no longer matches the tree.

Generated with SeekMoon


bobzhang:
Closing this: measured against the model it costs more than it saves.

Encoded into the system prompt, this change grows the generated prompt
from 64341 to 64810 chars (+469) to replace nine `@shell.` prefixes. The
guidance that has to teach the form costs far more context than the
prefixes it removes, and it adds a rule to follow plus a second spelling
of the same call. The qualified form is also what discovery uses
(`moon ide doc @moonbitlang/async/shell`), and `@alias.name` reads
self-documenting when copying a snippet without its import header.

The stated goal was ergonomics for the model's shell scripts; on that
measure `*` is a net negative, so the prompt should not teach it. Whether
to keep `*` for hand-written scripts is a separate question about human
ergonomics, not about the agent.

The branch is left in place; delete it if the choice is to drop this.

PR #1707, opened 2026-09-21
feat(edit): align diagnostic count guards and rollback feedback

bobzhang:
Single and batched edits exposed different diagnostic guard semantics and retry guidance. Both now use inclusive thresholds on the reported count change (`after - before`), while retaining a lower default tolerance for a single edit.

| Parameter | `edit` default | `multi_edit` default | Rollback condition |
| --- | --- | --- | --- |
| `revert_if_error_delta_greater_or_equal` | `5` | `10` | Error count delta `>= threshold` |
| `revert_if_warning_delta_greater_or_equal` | Off | Off | Warning count delta `>= threshold` |

Thresholds are non-negative integers with no additional host cap. Either condition can trigger rollback. For a warning-fix loop, explicitly pass error threshold `1` and warning threshold `0`: reported errors must not increase and warnings must decrease. Removing two diagnostics while adding one is allowed by a net-decrease guard. Diagnostic identities remain informational and do not decide rollback.

Both tools now share the existing package-graph comparability analysis and structured feedback (`verdict`, evidence, and `reissue_with`). When diagnostics indicate previously unreached errors, the retry suggestion is `delta + 1`, so it actually passes the inclusive guard, including values above `200`. Tool descriptions explain that fixing errors may expose existing diagnostics, while new errors may hide unchecked packages; lower counts do not guarantee no new problems. The agent must review the evidence before raising a threshold.

The shared implementation replaces duplicated policy and feedback. `multi_edit` now measures a baseline before writing, checks all touched projects for both severities, and removes its restore/check/reapply baseline path. It refuses an unavailable or incomplete baseline and rolls back an unverifiable post-write result. Cross-project counts remain usable, but a single package graph is not presented as proof of cross-project attribution. A baseline-time content change is rejected before writing.

This changes ordinary `edit` calls from opt-in error protection to a default threshold of `5`, and changes `multi_edit`'s former `> 10` boundary to `>= 10`. Checking a baseline can add a compiler run. Bundled workflows, examples, SDK docs, and English/Chinese prompts use the new names. There is no special legacy-name migration handling. Syntax-error protection remains enabled by default.

Validation:

- Common end-to-end tests cover both default boundaries, explicit overrides, newly reached errors, rollback, and retry at the suggested threshold, including a source symlink.
- Regression tests cover net warning progress, changed diagnostic identities at equal error count, inclusive thresholds, overrides and retry suggestions above `200`, and multiple projects.
- `moon info`, `moon fmt`, `just check`, `just build`, and `env -u DEEPSEEK -u KIMI -u GLM just test` (real API smoke tests skipped).


agent_tool/edit/internal/decode/decode.mbt
```diff
@@ -42,6 +42,16 @@ pub fn decode(arguments : Json) -> EditInput raise {
 
 ///|
 fn parse_fields(fields : Map[String, Json]) -> EditInput raise {
+  if fields.contains("revert_when_errors_above") {
```
bobzhang: this is not needed.
LLM does not absorb the info yet

agent_tool/edit/internal/decode/decode.mbt
```diff
@@ -42,6 +42,16 @@ pub fn decode(arguments : Json) -> EditInput raise {
...
+    fail(
+      "arguments.revert_when_errors_above was renamed to revert_when_errors_greater_or_equal; increase the old threshold by one",
+    )
+  }
+  if fields.contains("revert_when_warnings_above") {
```
bobzhang: same

agent_tool/edit/edit.mbt
```diff
@@ -114,13 +113,14 @@ pub fn definition(
...
+        "revert_when_errors_greater_or_equal": {
           "type": "integer",
           "description": (
             #|Post-write guard, off unless given. moon check runs from the file's module
-            #|before and after the write; if the edit INTRODUCED more than this many
+            #|before and after the write; if the edit INTRODUCED at least this many
             #|errors the tree did not have — counted by diagnostic identity (path, code,
-            #|message), so line shifts never count; 0 reverts on any — the file is
+            #|message), so line shifts never count; 1 reverts on any new error,
+            #|0 always reverts a checked edit — the file is
```
bobzhang: "0 always reverts a checked edit" what does this mean?
so line shifts never count -- what does this mean

bobzhang (commented):
Architecture review (not line-level). CI is green on all channels; the `comparability` lift into `agent_tool/internal/diagnostic_guard` is a 1:1 move with the tests carried over, and `multi_edit` losing the restore/check/reapply dance in favour of the same pre-write baseline + content re-check that `edit` already does is a clear simplification. The `errdefer` rollback now covering every checked batch is a good property.

One concern I think needs a decision before merge, two smaller shape points.

### 1. `multi_edit` is now fail-closed with no opt-out — and that is the one place the two tools are still *not* aligned

Because the error guard is default-on (`None → 10`) and `measure_baseline` runs unconditionally whenever any written file is in a MoonBit project, every `multi_edit` now requires a successful, complete baseline `moon check` before a single byte is written. `measure_baseline` refuses on:

- `Unavailable` (watchdog timeout, failed to spawn),
- `Failed(code, _)` (moon exits non-zero with no diagnostics: broken manifest, missing dep, unresolvable import),
- `Checked(tally) if tally.truncated` (>400k chars of JSON diagnostics),
- `NotApplicable`.

Previously the error guard's baseline was lazy (measured only on a breach) and an unverifiable check failed *open* — the batch was applied and the check line said so. The `Failed(code, _)` case in particular was deliberately folded into a zero-error tally; now it is a hard refusal.

There is no argument value that turns the guard off (`null` = default, a huge threshold still needs the baseline), so the model cannot use `multi_edit` at all on a tree whose check cannot run — which is exactly the tree it is most likely trying to repair. Its only escape is to fall back to one `edit` per site, because `edit` in this same PR is opt-in and fail-open. So after a PR whose purpose is aligning the guard semantics, the two tools now disagree on the most consequential property: what happens when the baseline is unavailable.

Two consistent options; either is fine, but the current middle ground is the worst of both:

- **Mirror `edit`:** fail closed only when a guard was *explicitly* passed; with defaults, an unavailable/incomplete baseline degrades to the old "applied, unverified" report. Cheap: `baseline` becomes `Option` on the explicit-guard path only, and the after-check keeps its old fail-open branch when `baseline is None`.
- **Keep fail-closed but give an explicit off** (e.g. `revert_if_error_delta_greater_or_equal: null` meaning off rather than default, or a separate boolean). Then the refusal message can point at it.

Related cost note, acknowledged in the description: the happy path now always pays two `moon check`s. On a warm cache the baseline is near-free, so I would not block on it, but it is worth stating in the README that the baseline is unconditional now, not "when a guard is requested".

### 2. The multi-project caveat leaks out of the shared package

Both callers do the same thing after `gather_comparability`:

```moonbit
let evidence = if projects.length() > 1 { { ..measured, graph_available: false } } else { measured }
```

That is the shared package's knowledge, not the callers'. If `gather_comparability` took the project list (`check_paths : Array[String]`) it could pick the graph and mark it unavailable itself, the two callers could not drift, and `Comparability` would no longer need to be `pub(all)` just to permit the struct update.

### 3. Silent guard loss on the old names (accepted trade-off, flagging only)

I saw the earlier legacy-name rejection was removed on purpose ("LLM does not absorb the info yet"), and I agree the model never sees the old names. The residual exposure is user-authored PTC scripts written against the `tools_sdk` README: an old `revert_when_warnings_above: 0` now decodes as "no guard", applies, and reports `applied`. Not asking for the check back — just noting that this is the one path where a safety argument fails silently rather than loudly, so it may deserve a line in the release note.

Everything else (inclusive thresholds, `delta + 1` reissue, the `0 = must decrease` semantics for warning-fix loops, count-over-identity as the rollback decision with identities kept as evidence) reads as a coherent design and the prompts/workflows are updated consistently.

bobzhang:
Narrowing point 1 after discussion, with a reproduction.

Withdrawn: the `edit` vs `multi_edit` asymmetry framing. A single edit is cheap for the agent to reverse itself, a batch is not, so opt-in on `edit` and default-on for `multi_edit` is the right split.

Still standing, and narrower than I first wrote it: **raising the threshold does not opt out of the baseline requirement**, because the threshold is never consulted before `measure_baseline` runs. In `multi_edit_files` the baseline branch is `if !check_paths.is_empty()`, and `Refused(reason) => return not_verified(reason)` fires before `revert_if_error_delta_greater_or_equal` is read at all.

Reproduction (moon 0.10.x, this machine):

```
moon.mod:       name = "tmp/broken"
lib/moon.pkg:   import { "tmp/broken/nonexistent" }
lib/lib.mbt:    pub fn f() -> Int { 0 }
$ moon check --output-json ; echo exit=$?
Error: Failed to calculate build plan
  Cannot find import 'tmp/broken/nonexistent' in tmp/broken/lib@0.0.0
exit=255
```

No JSON diagnostic, non-zero exit, so `@auto_check.check` yields `Failed(255, empty tally)`, `measure_baseline` returns `Refused("moon check exited 255 without diagnostics")`, and every `multi_edit` on that tree answers `not applied: a diagnostic guard is active, but moon check could not verify the tree` — including the batch that would fix the import, and at any threshold. On `main` the same batch applies: the old error guard folded `Failed` into a zero-error tally on purpose (the comment in the removed code says so), and only the warning guard, when explicitly requested, refused.

Minimal fix that keeps "opt-out = raise the threshold" true: in `measure_baseline`, treat `Failed(code, tally)` as `Checked(tally)` (the old fold) unless the warning guard was explicitly requested. Then a broken-plan tree is a 0-error baseline, the post-write `Failed` is also 0, delta is 0, and the batch is kept with the check line saying moon exited without diagnostics. `Unavailable` (timeout / spawn failure) and `truncated` can stay refused; those are rare and a refusal there is defensible.

bobzhang:
Resolving point 1 after discussion: the refusal is the intended contract, and I agree it is a defensible one.

`multi_edit` requires a computable build plan; a tree whose plan moon cannot make gets repaired with `edit` first (guards off by default, so the post-write check only reports), then batched. That is a clean precondition and it removes the need for the fold I suggested above. Withdrawn.

Two follow-ups so the precondition is visible rather than discovered, neither blocking:

- The refusal currently says `moon check exited 255 without diagnostics`. `Failed(code, tally)` discards moon's stderr, so the agent never sees `Cannot find import 'tmp/broken/nonexistent'` and is not told that `edit` is the way out. Carrying the first stderr line into the reason and appending "repair the build plan with `edit` first" would turn a dead end into a one-step detour.
- State the precondition in the `multi_edit` README / tool description: a batch is refused, at any threshold, until `moon check` can produce a build plan.

Nothing else from my side blocks merge.

bobzhang:
Pushed 846c6c955 with the rule we settled on: a tree moon cannot plan is the floor. `multi_edit` writes the batch, keeps it if the plan is computable afterwards (`data.plan_repaired`, no count comparison), and rolls it back with moon's reason plus "repair the build plan first, before any other edit" if it still fails. `Failed` now carries moon's explanation, so refusals name the cause. One field on `Baseline`, no new variants, tool descriptions untouched. 157/157 tests pass on native; `moon check --deny-warn` and `moon fmt --check` are clean.

